PR #1, opened 2018-09-06
[SIP-4] add lerna monorepo and`@superset-ui/core` package with `SupersetClient`

williaster:
This PR 
1) sets up the `superset-ui` **monorepo** (i.e., a repository of multiple packages) with [`lerna`](https://github.com/lerna/lerna/) and 
2) adds the first package `@superset-core`, which implements `SupersetClient` as described in [SIP-4](https://github.com/apache/incubator-superset/issues/5667). `SupersetClient` is pulled from [step 2 this PR](https://github.com/apache/incubator-superset/pull/5772), which was abandoned for smaller steps, but supports all the client-side async functionality encompassed by the current Superset app. 

Below are some more details about the monorepo + `@superset-ui/core` package, see the README (link soon) for API/usage details for the `SupersetClient`.

### Monorepo overview
[`lerna`](https://github.com/lerna/lerna/) is used to manage versions and dependencies between
packages in this monorepo.

```
superset-ui/
  lerna.json
  package.json
  ...
  packages/
    package1/
      package.json
      ...
      src/
      test/
      ...
      lib/
      esm/
      ...
    ...
```

To develop:

1. clone this repo
2. install the root npm modules including lerna and yarn
3. have lerna install package dependencies and manage the symlinking between packages for you

```sh
git clone ...superset-ui && cd superset-ui
npm install
lerna bootstrap
```

After this you can run scripts across the entire monorep (e.g., `yarn run test` runs tests across all repos) or within a given package. The latter means that **each package defines its own build config, linting, and testing**.  I initially set it up to have shared things like `jest`, `prettier`, and `linting` but I think this will be more painful down the line. I think we can achieve consistency with yeoman package generators which @kristw and I will do in a future PR.

### @superset-ui/core

After this PR is merged, we can publish the first version of this package. 

`@data-ui/build-config` is used to manage all `babel`, `jest` testing, `eslint`, `prettier`. This type of "shareable" build config will save a lot of copy/paste as the # of packages grows. 

### Testing
We should have good coverage for client-side code from the start 🔥 In a future PR we'll set up code coverage but here are the stats for what I wrote 🚀 

```
=== Coverage summary ===                                                                                                                
Statements   : 98.82% ( 84/85 )
Branches     : 90% ( 81/90 )
Functions    : 100% ( 34/34 )
Lines        : 98.78% ( 81/82 )
=====================
```

@kristw @conglei @mistercrunch @graceguo-supercat @john-bodley @hughhhh @betodealmeida @michellethomas


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+💔 Breaking Changes
```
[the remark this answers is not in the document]
williaster: maybe! I'll give it a shot

packages/superset-ui-core/src/SupersetClient.js
```diff
@@ -0,0 +1,180 @@
...
+      credentials: this.credentials,
+      headers: {
+        ...this.headers,
+      },
+      method: 'GET',
+      mode: this.mode,
+      timeout: this.timeout,
+      url: this.getUrl({ endpoint: 'superset/csrf_token/', host: this.host }),
+    })
+      .then(response => {
```
kristw: Perhaps use `.then(onFulfilled, onRejected)`
instead of `.then(onFulfilled).catch(onRejected)`

`.catch(onRejected)` is equivalent to`.then(undefined, onRejected)`

The current `catch` clause will handle `Promise.reject` thrown from the `then` clause as well, which has the same result in this case, but is redundant work.

packages/superset-ui-core/src/SupersetClient.js
```diff
@@ -0,0 +1,180 @@
...
+  isAuthenticated() {
+    return this.didAuthSuccessfully;
+  }
+
+  init() {
+    return this.getCSRFToken();
+  }
+
+  getCSRFToken() {
+    this.requestingCsrf = true;
```
kristw: Instead of having boolean flags `this.requestingCsrf`, we can store the promise returned from `callApi()` in `this.requestingCsrf`. The initial value of `this.requestingCsrf` can be `Promise.reject('not initialized')`. 

This may remove the need for `this.waitForCSRF` and polling/setTimeout mechanism as `this.ensureAuth` can just `return this.requestingCsrf;` directly.

packages/superset-ui-core/src/callApi.js
```diff
@@ -0,0 +1,109 @@
...
+
+// Fetch does not support timeout natively, so we add a layer with it.
+// We pass the timeoutId to callApi so that it can clear it
+export default function callApiWithTimeout({ timeout, ...rest }) {
+  let timeoutId;
+
+  return Promise.race([
+    new Promise((resolve, reject) => {
+      if (typeof timeout === 'number') {
+        timeoutId = setTimeout(
```
kristw: Perhaps create another function called `createTimer`.

```js
function createTimer(timeout) {
  return new Promise((resolve, reject) => {
    setTimeout(() => reject(...), timeout);
  });
}
```

To avoid worrying about clearing timeout or passing around `timeoutId`, can break the `callApi` function into two parts

```js
function callApiWithoutParsing() {
  // the first half of original callApi()
  // handle params and return fetch();
}

function parseResponse(apiPromise) {
  return apiPromise.then( .... ).then( .... ); // do the parsing
}

function callApi() {
  return parseResponse(callApiWithoutParsing());
}
```

If `timeout` is not a number, there is no need to race.

```js
function callApiWithTimeout() {
  const apiPromise = callApiWithoutParsing(...);
  // This will race until results come back from server
  // excluding parsing time
  const racedPromise = (typeof timeout === 'number') 
    ? Promise.race([createTimer(timeout), apiPromise])
    : apiPromise;
  return parseResponse(racedPromise);
}
```

packages/superset-ui-core/src/SupersetClient.js
williaster: sure, I think the `.catch` is a little more readable but don't have a super strong opinion.

packages/superset-ui-core/src/SupersetClient.js
williaster: that's a great idea, I'll try to move toward that 👍

packages/superset-ui-core/src/callApi.js
```diff
@@ -0,0 +1,109 @@
...
+    .then(({ response, json, text }) => {
+      if (!response.ok) {
+        return Promise.reject({
+          error: response.error || (json && json.error) || text || 'An error occurred',
+          status: response.status,
+          statusText: response.statusText,
+        });
+      }
+
+      return Promise.resolve(typeof text === 'undefined' ? { json, response } : { response, text });
```
williaster: I think this `resolve` is extraneous, will remove. the `rejection` above is useful to forward the error.

packages/superset-ui-core/src/callApi.js
williaster: done, my only concern here is what happens to the timeouts without clearing them (when say a dashboard makes 15 requests for chart data). I couldn't find much info online about this, assume they're garbage collected.

packages/superset-ui-core/src/SupersetClient.js
williaster: done

packages/superset-ui-core/test/SupersetClient.test.js
```diff
@@ -0,0 +1,497 @@
+/* eslint promise/no-callback-in-promise: 'off' */
+import sinon from 'sinon';
```
williaster: just learned more about jest spys, should be able to move this

packages/superset-ui-core/README.md
```diff
@@ -0,0 +1,90 @@
...
+import { SupersetClient } from '@superset-ui/core';
+import AbortController from 'abortcontroller-polyfill';
+
+const controller = new AbortController();
+const { signal } = controller;
+
+SupersetClient.get({}).then(...).catch(...);
+
+if (IWantToCancelForSomeReason) {
+  signal.abort(); // Promise is rejected, request `catch` is invoked
```
kristw: Does it cancel every outgoing request? Is there a way to cancel just one?

williaster: oops I forgot an important part of this usage (I'll update), you pass the signal to the request, so it's per-request aborting. 

see [here](https://github.com/apache/incubator-superset/pull/5772/files#diff-66f9c9696a5b417d0f13a79c9f1d3a7fR141) for an example in superset.

packages/superset-ui-core/src/callApi/parseResponse.js
```diff
@@ -0,0 +1,29 @@
...
+              error: response.error || (json && json.error) || text || 'An error occurred',
+              status: response.status,
+              statusText: response.statusText,
+            });
+          }
+
+          return typeof text === 'undefined' ? { json, response } : { response, text };
+        }),
+    // forward the error
+    error => Promise.reject(error),
```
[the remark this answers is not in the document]
williaster: linting has complained is some places that we need a `catch` for every `then`, but it doesn't here 🤷‍♀️

packages/superset-ui-core/test/callApi/callApi.test.js
```diff
@@ -0,0 +1,156 @@
+/* eslint promise/no-callback-in-promise: 'off' */
```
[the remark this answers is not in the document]
williaster: learned a lot about jest!

packages/superset-ui-core/src/SupersetClient.js
```diff
@@ -0,0 +1,155 @@
...
+      this.csrfPromise ||
+      Promise.reject({
+        error: `SupersetClient has no CSRF token, ensure it is initialized or
+        try logging into the Superset instance at ${this.getUrl('/login')}`,
+      })
+    );
+  }
+
+  get({ host, url, endpoint, mode, credentials, headers, body, timeout, signal }) {
+    return this.ensureAuth().then(() =>
```
kristw: What do you think about attaching an `abort` function to the returned promise?

```js
const promise = this.ensureAuth().then(...);
promise.abort = signal && () => signal.abort();
return promise;
```

So can use like this. 

```js
const myPromise = SupersetClient.get(...);
if(wannaCancel) {
  myPromise.abort();
}
```

The concern is the function `abort()` will only be available to the original Promise returned from `get` and not on the chained `then()`, which may be not obvious to developer.

williaster: hmm, that's an interesting thought, the abort controller api is a little verbose. I do think it would be non-ideal to have the `abort` just attached to the initial promise. 

maybe we can see how it goes with `signal` and we could develop a nicer / more complex api around aborting in the future? there were only a couple of places in Superset that use this currently.

williaster: in practice you either have to pass the abort controller around, or the promise, and so I'm not sure there's much advantage of one over the other except that the former is full-fledged API spec not just our library API.

kristw: sgtm. We can try that later. I think this first version looks good to me.

packages/superset-ui-core/test/callApi/rejectAfterTimeout.test.js
```diff
@@ -0,0 +1,20 @@
+/* eslint promise/no-callback-in-promise: 'off' */
+import rejectAfterTimeout from '../../src/callApi/rejectAfterTimeout';
+import throwIfCalled from '../utils/throwIfCalled';
+
+describe('rejectAfterTimeout', () => {
```
kristw: nit
Add `()` in `describe('rejectAfterTimeout()'`) so we know it is a function when reading test result

packages/superset-ui-core/test/callApi/parseResponse.test.js
```diff
@@ -0,0 +1,84 @@
+/* eslint promise/no-callback-in-promise: 'off' */
+import fetchMock from 'fetch-mock';
+import callApi from '../../src/callApi/callApi';
+import parseResponse from '../../src/callApi/parseResponse';
+
+import { LOGIN_GLOB } from '../fixtures/constants';
+import throwIfCalled from '../utils/throwIfCalled';
+
+describe('parseResponse', () => {
```
kristw: nit
Add `()` in `describe('parseResponse()'`) so we know it is a function when reading test result

packages/superset-ui-core/test/callApi/callApiAndParseWithTimeout.test.js
```diff
@@ -0,0 +1,102 @@
...
+
+// we import these via * so that we can spy on the 'default' property of the object
+import * as callApi from '../../src/callApi/callApi';
+import * as parseResponse from '../../src/callApi/parseResponse';
+import * as rejectAfterTimeout from '../../src/callApi/rejectAfterTimeout';
+
+import { LOGIN_GLOB } from '../fixtures/constants';
+import throwIfCalled from '../utils/throwIfCalled';
+
+describe('callApiAndParseWithTimeout', () => {
```
kristw: nit
Add `()` in `describe('callApiAndParseWithTimeout()'`) so we know it is a function when reading test result

packages/superset-ui-core/test/callApi/callApi.test.js
```diff
@@ -0,0 +1,156 @@
...
+import fetchMock from 'fetch-mock';
+import callApi from '../../src/callApi/callApi';
+
+import { LOGIN_GLOB } from '../fixtures/constants';
+import throwIfCalled from '../utils/throwIfCalled';
+
+describe('callApi', () => {
```
kristw: nit
Add `()` in `describe('callApi()'`) so we know it is a function when reading test result

packages/superset-ui-core/src/SupersetClient.js
```diff
@@ -0,0 +1,155 @@
...
+          this.csrfToken = response.json.csrf_token;
+          this.headers = { ...this.headers, 'X-CSRFToken': this.csrfToken };
+          this.didAuthSuccessfully = !!this.csrfToken;
+        }
+
+        if (!this.csrfToken) {
+          return Promise.reject({ error: 'Failed to fetch CSRF token' });
+        }
+
+        return response;
```
kristw: Should this return `response` or `this.csrfToken`?
Nobody is really using its result at this point, but seems like `this.csrfToken` will be more expected?

If somebody called this multiple times, it will send new request every time. Is that expected behavior? or if a promise already exists, then return `this.csrfPromise` so it will send request only once.

PR #5, opened 2018-09-13
[core] add `parseMethod`

williaster:
🏆 Enhancements
- This adds a `parseMethod` argument to `parseResponse` in order to support both `json` (default) and `text` responses. This is needed to enable custom response parsing in some cases, for example `BigInt`s which was [added to Sqllab recently](https://github.com/apache/incubator-superset/pull/5829).
- Exposes this `parseMethod` as an optional argument to `SupersetClient.get()` and `.post()` calls.

🏠 Internal
- Bumps `@data-ui/build-config` to `0.0.14` which has a stricter `no-unused-args` eslint rule
- adds `.npmrc` to `.gitignore` for local publishing credentials

@kristw @conglei @mistercrunch

### DIFF
--- a/packages/superset-ui-core/src/SupersetClient.js
+++ b/packages/superset-ui-core/src/SupersetClient.js
@@ -75,7 +75,7 @@ class SupersetClient {
     );
   }
 
-  get({ host, url, endpoint, mode, credentials, headers, body, timeout, signal }) {
+  get({ body, credentials, headers, host, endpoint, mode, parseMethod, signal, timeout, url }) {
     return this.ensureAuth().then(() =>
       callApi({
         body,
@@ -83,6 +83,7 @@ class SupersetClient {
         headers: { ...this.headers, ...headers },
         method: 'GET',
         mode: mode || this.mode,
+        parseMethod,
         signal,
         timeout: timeout || this.timeout,
         url: url || this.getUrl({ endpoint, host: host || this.host }),
@@ -91,16 +92,17 @@ class SupersetClient {
   }
 
   post({
-    host,
-    endpoint,
-    url,
-    mode,
     credentials,
     headers,
+    host,
+    endpoint,
+    mode,
+    parseMethod,
     postPayload,
-    timeout,
     signal,
     stringify,
+    timeout,
+    url,
   }) {
     return this.ensureAuth().then(() =>
       callApi({
@@ -108,6 +110,7 @@ class SupersetClient {
         headers: { ...this.headers, ...headers },
         method: 'POST',
         mode: mode || this.mode,
+        parseMethod,
         postPayload,
         signal,
         stringify,
--- a/packages/superset-ui-core/src/callApi/parseResponse.js
+++ b/packages/superset-ui-core/src/callApi/parseResponse.js
@@ -1,35 +1,29 @@
+const PARSERS = {
+  json: response =>
+    // first try to parse as json, and fall back to text (e.g., in the case of HTML stacktrace)
+    // cannot fall back to .text() without cloning the response because body is single-use
+    response
+      .clone()
+      .json()
+      .then(json => ({ json, response }))
+      .catch(() => /* jsonParseError */ response.text().then(text => ({ response, text }))),
+
+  text: response => response.text().then(text => ({ response, text })),
+};
+
 export default function parseResponse(apiPromise, parseMethod = 'json') {
-  return apiPromise.then(response => {
-    let parsedPromise;
-    if (parseMethod === 'json') {
-      // first try to parse as json, and fall back to text (e.g., in the case of HTML stacktrace)
-      // cannot fall back to .text() without cloning the response because body is single-use
-      parsedPromise = response
-        .clone()
-        .json()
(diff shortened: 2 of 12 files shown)


packages/superset-ui-core/src/SupersetClient.js
```diff
@@ -75,14 +75,15 @@ class SupersetClient {
     );
   }
 
-  get({ host, url, endpoint, mode, credentials, headers, body, timeout, signal }) {
+  get({ body, credentials, headers, host, endpoint, mode, parseMethod, signal, timeout, url }) {
```
kristw: It starts to get long, do you want to separate into one param per line?

kristw: Oh just notice `post` is already doing that, then maybe good to spread this one out too.

packages/superset-ui-core/src/callApi/parseResponse.js
```diff
@@ -1,26 +1,29 @@
...
+      .catch(() => /* jsonParseError */ response.text().then(text => ({ response, text }))),
 
-        return typeof text === 'undefined' ? { json, response } : { response, text };
-      }),
-  );
+  text: response => response.text().then(text => ({ response, text })),
+};
+
+export default function parseResponse(apiPromise, parseMethod = 'json') {
+  const responseParser = PARSERS[parseMethod] || PARSERS.json;
```
kristw: Can developer pass `parseMethod` as a custom function in addition to string of predefined parser name?

packages/superset-ui-core/src/callApi/parseResponse.js
```diff
@@ -1,26 +1,29 @@
...
+
+  return apiPromise.then(responseParser).then(({ json, response, text }) => {
+    // HTTP 404 or 500 are not rejected, ok is just set to false
+    if (!response.ok) {
+      return Promise.reject({
+        error: response.error || (json && json.error) || text || 'An error occurred',
```
kristw: Perhaps should separate error handling for `apiResponse`, `json` parsing and `text` parsing.

packages/superset-ui-core/src/callApi/parseResponse.js
```diff
@@ -1,26 +1,29 @@
...
-      .then(({ response, json, text }) => {
-        if (!response.ok) {
-          return Promise.reject({
-            error: response.error || (json && json.error) || text || 'An error occurred',
-            status: response.status,
-            statusText: response.statusText,
-          });
-        }
+      .then(json => ({ json, response }))
+      .catch(() => /* jsonParseError */ response.text().then(text => ({ response, text }))),
```
kristw: Does this mean the developer always have to check if `output.json` exists or not? If `output.json` does not exist, they should use to `output.text` instead. If they forget to check. This can cause bugs. I wonder if there is a way to make it more explicit. 

Also, generally, does one expect this Promise to reject when fail to parse or resolve text of the supposed-to-be json? IMO, I would expect it to fail and give me the text in the error message for debugging.

packages/superset-ui-core/src/callApi/parseResponse.js
kristw:
```js
return apiPromise
  .then((response) => {
    if (!response.ok) {
      return Promise.reject({
        error: response.error,
        status: response.status,
        statusText: response.statusText,
      });
    }
    return response;
  })
  .then(responseParser)
```

packages/superset-ui-core/src/SupersetClient.js
williaster: it's all based on prettier config, it auto wraps if it's over the 100 char limit (which `post` is but `get` is not). we could decrease the char limit if you feel strongly about it tho.

kristw: nope. this is pretty minor. I'll leave it to you.

packages/superset-ui-core/src/callApi/parseResponse.js
williaster: I thought about that but am not sure a free-form parser makes sense. There are only a few available https://developer.mozilla.org/en-US/docs/Web/API/Response#Methods, of which we are exposing `text` and `json`. 

It doesn't seem worthwhile to copy / mirror the full response API to me (would begin questioning the need for this lib). One option could be to allow `null` or `'none'` as a `parseMethod` to enable returning the response unmodified and un-read. Beyond that, any custom parsing I think should come from a chained `.then(({ json }) => myCustomParser(json))`

packages/superset-ui-core/src/callApi/parseResponse.js
williaster: I dig the latter^

packages/superset-ui-core/src/callApi/parseResponse.js
williaster: Yeah now that we expose the parse method and are explicit about `text`/`json` we should probably _not_ clone and fallback to `text` gracefully, and instead it should probably throw.

williaster: this broke a lot of tests, which were falling back to `.text()` unknowingly because of my mock text response 😜

PR #11, opened 2018-10-23
[build-config] move to monorepo root

williaster:
🏠 Internal

With more packages coming, this PR 

- moves the build-config to the monorepo root `package.json` so that package creators don't have to deal with the boilerplate. This will help keep things consistent across packages, too. Note that currently this build config does _not_ support webpack or TS, but we can easily add support for that with `@beemo`. I tested locally with another test package and everything seems to work okay.

- adds a pre-commit hook to run prettier on any staged files ✨ 

- adds a `test:watch` script to support `jest --watch` mode. this isn't easy out of the box with `beemo` because `jest` runs in a subprocess so output is not piped to the command line, but the script provides a workaround

@kristw @xtinec

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@
 .eslintrc.js
 .idea
 .npm
-.npmrc
 .prettierignore
 .yarnclean
 
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,8 @@ install:
   - npm install
   - npm install -g codecov
 
-env:
-  - PACKAGE=superset-ui-core
-
 script:
-  - cd ./packages/$PACKAGE
   - yarn install
-  - yarn run lint
   - yarn run test
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ matrix:
 install:
   - npm install
   - npm install -g codecov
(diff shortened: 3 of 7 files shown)


williaster:
This looks good except that jest coverage is being reported as unknown in the build. It works fine locally, so unsure what exactly is wrong.

williaster:
Looks like it has been reporting unknown for a couple PRs, will fix in another PR as I think it's (another) build config issue.

williaster:
the jest coverage is fixed 🎉 talk about a worm hole ...

williaster:
notes from this 

**jest**
- jest stopped reporting coverage (only on Travis) because we were defining `jest.config.js` `coveragePathIgnorePatterns` in the form `['node_modules/', 'build/', ...]`. 
- Travis builds are executed in a `/build/` directory on their machines so, and the ignore pattern match uses the full directory location. `@data-ui/build-config@^0.0.25` pre-pends `<roodDir>/` before each of these, e.g., `['<rootDir>/node_modules/', '<rootDir>/build/', ...]` to prevent this.

**workspaces**
- `Beemo` supports workspaces out of the box, i.e., support for monorepos
- this PR defined a `workspaces` entry the root package.json which is used  by both `Beemo` and `yarn`
- for the build, `yarn` `workspaces` actually do a lot of what `lerna` helps us do (technically don't even need `lerna` with `workspaces`), and hoists all package dependencies to the root, and sym-links inter-package dependencies (which `lerna bootstrap would do`)

PR #15, opened 2018-10-26
Add tn / translateWithNumber

kristw:
🏆 Enhancements
- Add `tn` (`translateWithNumber`) to `translation` module
- Remove `sprintf-js` because `jed` already comes with `sprintf`.

### DIFF
--- a/packages/superset-ui-translation/test/Translator.test.js
+++ b/packages/superset-ui-translation/test/Translator.test.js
@@ -30,11 +30,48 @@ describe('Translator', () => {
     it('returns undefined for undefined input', () => {
       expect(translator.translate(undefined)).toBeUndefined();
     });
+    it('returns original text for unknown text', () => {
+      expect(translator.translate('abc')).toEqual('abc');
+    });
     it('translates simple text', () => {
       expect(translator.translate('second')).toEqual('秒');
     });
-    it('translates template text with arguments', () => {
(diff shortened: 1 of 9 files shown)


packages/superset-ui-translation/test/Translator.test.js
```diff
@@ -30,11 +30,48 @@ describe('Translator', () => {
...
+      expect(translator.translateWithNumber('singular', null)).toBeNull();
+      expect(translator.translateWithNumber(null, null)).toBeNull();
+    });
+    it('returns undefined if singular or plural is undefined', () => {
+      expect(translator.translateWithNumber(undefined, 'plural')).toBeUndefined();
+      expect(translator.translateWithNumber('singular', undefined)).toBeUndefined();
+      expect(translator.translateWithNumber(undefined, undefined)).toBeUndefined();
+    });
+    it('returns original text for unknown text', () => {
+      expect(translator.translateWithNumber('fish', 'fishes', 1)).toEqual('fish');
```
williaster: don't need another test, but curious how it behaves with 0 since that is the default `num`. is that singular?

kristw: I think that depends on how the languagePack is setup. I saw there are DSL-ish thing in the language packs that defines `n` for singular and plural.

PR #17, opened 2018-10-30
Update readme and remove unnecessary items from package.json

kristw:
- Add `npm run bootstrap` to call `lerna bootstrap`
- Update instructions on README
- Remove unnecessary items from `package.json` in `translation`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# @superset-ui 
+# @superset-ui
 
 [![Build Status](https://img.shields.io/travis/com/apache-superset/superset-ui.svg?style=flat-square
 )](https://travis-ci.com/apache-superset/superset-ui)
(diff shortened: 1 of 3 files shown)


packages/superset-ui-translation/package.json
```diff
@@ -25,27 +25,11 @@
...
   "homepage": "https://github.com/apache-superset/superset-ui#readme",
   "devDependencies": {
-    "@data-ui/build-config": "^0.0.23",
     "jest-mock-console": "^0.4.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.1.2",
     "jed": "^1.1.1"
   },
-  "beemo": {
```
williaster: why didn't I remove these?! thanks.

README.md
```diff
@@ -44,22 +44,21 @@ superset-ui/
...
 1. clone this repo
-2. install the root npm modules including lerna and yarn
-3. have lerna install package dependencies and manage the symlinking between packages for you
+2. have lerna install package dependencies and manage the symlinking between packages for you
 
 ```sh
 git clone ...superset-ui && cd superset-ui
 npm install
-lerna bootstrap
+npm run bootstrap
```
williaster: I think this is okay but `yarn` is a `root dep`, and most of the scripts are run with `yarn run ...`.

This might not be as important locally / in the docs, but this seems to be important for the `travis` build right now. `learna bootstrap` fails when run in the travis env which I found last week. it seems like `travis` uses it's own flavor of `npm` which gets called if you call `lerna bootstrap`, and it fails the build without a `lock` file, which we don't have or want. I enabled `yarn` workspaces for the build-config, and they actually also can do between-package sym linking like `lerna bootstrap`, so it does what we want.

With the new SIP for moving away from `npm` we might have to re-think this, but wanted to give more context as to why `yarn` is around. I'll leave it to you if you want to use `npm` or `yarn` here, like I said for docs it probably doesn't matter.

PR #20, opened 2018-11-01
@superset-ui/color

kristw:
🏆 Enhancements

Migrate modules related to color from `incubator-superset`


package.json
```diff
@@ -54,7 +54,16 @@
         }
       },
       "prettier"
-    ]
+    ],
```
williaster: just released a new config version, could you try `^0.0.26` before going this route? if that version doesn't work I'm fine merging and I'll fix in another PR.

PR #21, opened 2018-11-05
Add generator-superset

kristw:
🏆 Enhancements

- Add `generator-superset` package for scaffolding.
- This package will be published as `@superset-ui/generator-superset`.
- Update build script to skip `babel` tasks for `generator-superset`, as it is a `node` application.
- Use `jest` project configuration to separate `generator-superset` from others and run in `testEnvironment: "node"`. (Although that was not the real issue that demolished CI in the early attempts.)

### Note about yeoman testing and interference with other packages

The key thing to add to `yeoman` tests, which are not normally included even from official yeoman `generator-generator`, is to **change working directory back to original working directory after the test has completed.** `yeoman` tests switch to `tmp` directory and write files there. Usually this is fine for solo package. However, for a monorepo like this one, it made `jest` confuses with current directory (being in `tmp` directory instead of `superset-ui` root) and interferes with other tests in sibling packages that are run after the `yeoman` tests. 

By adding the directory switch in `afterAll()` block. Everything works fine again! 🍾 

image

> Many yaks were shaved in the making of this pull request.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
           "displayName": "generator-superset",
           "rootDir": "<rootDir>/packages/generator-superset",
           "testMatch": [
-            "<rootDir>/**/?(*.)+(spec|test).{js,jsx}"
+            "<rootDir>/packages/generator-superset/**/?(*.)+(spec|test).{js,jsx}"
           ]
         }
       ]
--- a/package.json
+++ b/package.json
@@ -57,20 +57,18 @@
       "prettier"
     ],
     "jest": {
+      "testPathIgnorePatterns": [
+        "<rootDir>/packages/generator-superset"
+      ],
       "projects": [
+        "<rootDir>",
         {
-          "displayName": "main",
-          "rootDir": "<rootDir>",
-          "testMatch": [
-            "<rootDir>/packages/superset-ui-*/**/?(*.)+(spec|test).{js,jsx}"
-          ]
-        },
-        {
-          "displayName": "generator-superset",
+          "displayName": "node",
           "rootDir": "<rootDir>/packages/generator-superset",
           "testMatch": [
             "<rootDir>/test/**/?(*.)+(spec|test).{js,jsx}"
-          ]
+          ],
+          "testEnvironment": "node"
         }
       ]
     }
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "build": "yarn run build:cjs && yarn run build:esm",
-    "build:cjs": "NODE_ENV=production beemo babel ./src --out-dir lib/ --minify --workspaces=\"@superset-ui/!(demo)\"",
-    "build:esm": "NODE_ENV=production beemo babel ./src --out-dir esm/ --esm --minify --workspaces=\"@superset-ui/!(demo)\"",
+    "build:cjs": "NODE_ENV=production beemo babel ./src --out-dir lib/ --minify --workspaces=\"@superset-ui/!(demo|generator-superset)\"",
+    "build:esm": "NODE_ENV=production beemo babel ./src --out-dir esm/ --esm --minify --workspaces=\"@superset-ui/!(demo|generator-superset)\"",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -55,7 +55,23 @@
         }
       },
       "prettier"
-    ]
+    ],
+    "jest": {
+      "testPathIgnorePatterns": [
```
williaster: can you tell if this gets _added_ to the ones generated `beemo`?

packages/generator-superset/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Usage
+
+Generate a new package in `@superset-ui`
+
+```bash
+cd superset-ui/packages
+mkdir superset-ui-new-package
+cd superset-ui-new-package
+yo @superset-ui/superset
```
williaster: is this line correct? not `@superset-ui/generator-superset`?

packages/generator-superset/README.md
```diff
@@ -0,0 +1,32 @@
...
+[![David (path)](https://img.shields.io/david/apache-superset/superset-ui.svg?path=packages%2Fgenerator-superset&style=flat-square)](https://david-dm.org/apache-superset/superset-ui?path=packages/generator-superset)
+
+> Scaffolder for Superset
+
+## Installation
+
+First, install [Yeoman](http://yeoman.io) and `generator-superset` using [npm](https://www.npmjs.com/) (we assume you have pre-installed [node.js](https://nodejs.org/)).
+
+```bash
+npm install -g yo
```
williaster: 1) is this recommended vs adding `yo` as a `dep` and getting it when you install this package? 
2) also wondering if we should suggest they just run `npm run bootstrap` in the root, and then `@superset-ui/generator-superset` + its deps will be installed?
3) if we do keep the separate installs, do we need them to be global packages?

packages/generator-superset/test/app.test.js
```diff
@@ -0,0 +1,23 @@
+const path = require('path');
+const assert = require('yeoman-assert');
+const helpers = require('yeoman-test');
+
+describe('generator-superset:app', () => {
+  let dir;
+
+  beforeAll(() => {
+    dir = process.cwd();
```
williaster: could we add a comment for why we `cd` since it was so tricky to figure this out?

package.json
kristw: The `jest.config.js` generated by beemo doesn't have this field by default. 
This was added to avoid running the `generator` tests twice.

packages/generator-superset/README.md
kristw: Yes, it is correct. `yo xxx` looks for global npm packages (or `npm link`'ed package) named `generator-xxx`. 

You will notice all `yeoman` generator package names start with `generator-`

This is why I am tempted to name this package `generator-superset` without `@superset-ui` so we can call `yo superset`, but that will not published under the npm org.

packages/generator-superset/test/app.test.js
kristw: Yes!

packages/generator-superset/README.md
kristw: 1/3) This is the recommended way as there should be single `yo` client. and `yo` looks for generator in the global scope. 
http://yeoman.io/learning/index.html

2) This `README` was written in a way that this packaged can be used independent from this repo. I envision this generator can be extended to have multiple subgenerator for scaffolding independent chart plugin repo, etc. It will be more convenient to install the generator package alone without cloning this monorepo only to `npm link` the generator.

package.json
williaster: sorry I wasn't clear, just wanted to see if this ignore pattern is added to or over-writes the `testPathIgnorePatterns` generated by `beemo`

packages/generator-superset/README.md
williaster: ah got it, thanks for the explanation 👍  

I think using the `@superset-ui` namespace would still be good.

package.json
kristw: I tested removing the field from `package.json` and `jest.config.js` remove the field completely. Probably doesn't have that in base. 

I tried adding something to `package.json`'s `testMatch` yesterday and that gets **added** to the `jest.config.js` along with the one coming from `beemo`

PR #27, opened 2018-11-08
Add RegistryWithDefaultKey class

kristw:
🏆 Enhancements

- Add class `RegistryWithDefaultKey`. This was inspired by `ColorSchemeRegistry` that has default color scheme and support getting/setting default color scheme. We will create more registries in the future such as `NumberFormatRegistry` and can take advantage of this base class. 

**Example Usage**

```js
class NewRegistryClass extends RegistryWithDefaultKey {
  constructor() {
    super({
      initialDefaultKey: 'some-key', 
      setFirstItemAsDefault: false, 
    });
  }
}
```

- `ColorSchemeRegistry` is refactored to use `RegistryWithDefaultKey`.

**BREAKING CHANGE:**

`getDefaultSchemeName()` and `setDefaultSchemeName()` are renamed to `getDefaultKey()` and `setDefaultKey()`

🏠 Internal

- Import utility functions from `lodash` instead of `lodash/fp` as the first one is better when chaining is not required. 
- Add more unit tests to increase branch coverage.


kristw:
I just revised and now it is ready for review.

PR #32, opened 2018-11-14
Add ExtensibleFunction class

kristw:
🏆 Enhancements

Add `ExtensibleFunction` class, which was inspired by the need to have classes that are also functions. 

For example, `CategoricalColorScale` or `NumberFormatter` should be functions themselves. Currently `CategoricalColorScale` has an awkward `.toFunction()` that needs to be called to convert to function, and also error-prone if developer forgets to call `.toFunction()`. Current usage in Superset is similar to the example code below.

```js
const colorScale = new CategoricalColorScale(...).toFunction();
colorScale(someValue); // 'green'
```

Would be cleaner if we can call 

```js
const colorScale = new CategoricalColorScale(...);
colorScale(someValue); // 'green'

const format = new NumberFormatter('.2f');
format(100); // 100.00
```

I still want to keep `CategoricalColorScale` and `NumberFormatter` being `class` to add fields and functions in a clean way (Add class-specific functions to `prototype`, rather than creates a function then sticks all the fields and functions to the returned function similar to module pattern that is common pre-es6). The `ExtensibleFunction` base class then provides a nice solution for this scenario. 

The implementation comes from this article. 
https://stackoverflow.com/questions/36871299/how-to-extend-function-with-es6-classes

### DIFF
--- a/packages/superset-ui-core/src/index.js
+++ b/packages/superset-ui-core/src/index.js
@@ -1,3 +1,4 @@
+export { default as ExtensibleFunction } from './models/ExtensibleFunction';
 export { default as Plugin } from './models/Plugin';
 export { default as Preset } from './models/Preset';
 export { default as Registry } from './models/Registry';
--- a/packages/superset-ui-core/src/models/ExtensibleFunction.js
+++ b/packages/superset-ui-core/src/models/ExtensibleFunction.js
@@ -1,0 +1,9 @@
+/**
+ * From https://stackoverflow.com/questions/36871299/how-to-extend-function-with-es6-classes
+ */
+
+export default class ExtensibleFunction extends Function {
+  constructor(fn) {
+    return Object.setPrototypeOf(fn, new.target.prototype);
+  }
+}
--- a/packages/superset-ui-core/test/index.test.js
+++ b/packages/superset-ui-core/test/index.test.js
@@ -1,4 +1,5 @@
 import {
+  ExtensibleFunction,
   Plugin,
   Preset,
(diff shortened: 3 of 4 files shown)


kristw:
Hmm. That's a good read. Let me check a bit more if this is really concerning. Don't want to slow down everything just for this.

From the author of the code in StackOverFlow

> Object.setPrototypeOf is not that much of an optimisation hazard as long it is done right after creating the object. It's just if you mutate the [[prototype]] of an object back and forth during its lifetime that it will be bad.

williaster:
interesting, that makes sense. I guess in the end I'd be disappointed in JS engines if this were to slow things down that much ... 🙂

kristw:
It seems like this is also how babel transpiles inheritance.

> The Babel compiler implements inheritance using a simple function called _inherits :

```js
function _inherits(subClass, superClass) {
  if (typeof superClass !== "function" 
    && superClass !== null) {
      throw new TypeError("Super expression must" + 
        "either be null or a function, not " 
        + typeof superClass);
  }
  subClass.prototype = Object.create(superClass 
    && superClass.prototype, {
      constructor: {

         value: subClass,
         enumerable: false,
         writable: true,
         configurable: true
       }
  });
  if (superClass) Object.setPrototypeOf 
     ? Object.setPrototypeOf(subClass, superClass) 
     : subClass.__proto__ = superClass;
}
```

https://www.todaysoftmag.com/article/2578/front-end-architecture-common-patterns-of-classical-inheritance-in-javascript

PR #33, opened 2018-11-15
make CategoricalColorScale instance a function and remove .toFunction()

kristw:
🏆 Enhancement

**feat:** Use the recently added `ExtensibleFunction` to make an instance of `CategoricalColorScale` be a function itself rather than having to call `.toFunction()` to convert it into a color function.

💔 Breaking Changes

**BREAKING CHANGE:** remove `categoricalColorScale.toFunction()`
**BREAKING CHANGE:** the color scale no longer convert input to lowercase before finding color.

🐛 Bug Fix
**Fix:** Also apply `cleanValue` to the namespace level


packages/superset-ui-color/src/cleanValue.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Ensure value is a string
+ * @param {any} value
+ */
+export default function cleanValue(value) {
```
williaster: could we possibly make this have a more specific name? `clean` is pretty generic, something like `stringyifyAndTrim` would make it more readable/clear what exactly "clean" means.

PR #41, opened 2018-11-21
add Typescript support + convert @superset-ui/connection to Typescript

williaster:
🏠 Internal

This PR
- adds `Typescript` support to the build via an upgraded version of the `@data-ui/build-config` package. 
  - To comment on any of the build internals, please see [that PR](https://github.com/williaster/build-config/pull/9) (which has not been merged but has been used to create the `Typescript`-compatible version and several `rc` versions before that). 
  - This PR added `Typescript` support to `babel`, `eslint`, `jest`, and added the `Typescript` beemo driver.
- converts the `@superset-ui/connection` package to `Typescript` as a proof-of-concept for the build support and to ramp up on `Typescript` myself ✨ 

**DEFINITELY** interested in any input on the Typescript, I'm not sure how best to handle Promise resolve + reject + any errors well without also defaulting to `any`. This is definitely reviewable, but I may need to keep tweaking some things to fix the build/debug things.

@kristw @conglei @xtinec


packages/superset-ui-connection/src/callApi/parseResponse.ts
```diff
@@ -0,0 +1,27 @@
+const PARSERS = {
+  json: (response: Response): Promise<JsonResponse | any> =>
```
williaster: the final line of this file would not compile unless I added `any` here, I thiiink it's due to a possible error but am not sure. It's weird to me that the return type of the `parseResponse` Promise can not include `any` 🤔

packages/superset-ui-connection/types/SupersetClient/index.d.ts
```diff
@@ -0,0 +1,76 @@
+type body = RequestInit['body'];
```
williaster: unsure if it's better for these `type`s to all be `PascalCase` in addition to the `interface`s

packages/superset-ui-connection/types/SupersetClient/index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+  credentials?: credentials;
+  csrfToken?: csrfToken;
+  headers?: headers;
+  host?: host;
+  protocol?: protocol;
+  mode?: mode;
+  timeout?: timeout;
+}
+
+interface CallApi {
```
williaster: aside from the `type`s which are re-used across `interface`s, some of these seem slightly more useful within the files themselves. thoughts?

packages/superset-ui-connection/types/SupersetClient/index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+type postPayload = { [key: string]: any };
+type mode = RequestInit['mode'];
+type redirect = RequestInit['redirect'];
+type timeout = number | void;
+type parseMethod = 'json' | 'text';
+type signal = RequestInit['signal'];
+type stringify = boolean;
+type url = string;
+
+interface ClientConfig {
```
reklawnos: It seems like it's possible to further dedupe these interfaces a bit further, and pull out some types from `RequestInit` directly. Maybe something like: 
```typescript
interface ClientConfig extends Partial<Pick<RequestInit, 'credentials' | 'headers' | 'mode'>> {
  csrfToken?: csrfToken;
  host?: host;
  protocol?: protocol;
  timeout?: timeout;
}
```

packages/superset-ui-connection/src/SupersetClient.ts
```diff
@@ -0,0 +1,200 @@
...
+      },
+      method: 'GET',
+      mode: this.mode,
+      timeout: this.timeout,
+      url: this.getUrl({ endpoint: 'superset/csrf_token/', host: this.host }),
+    }).then((response: JsonResponse | any) => {
+      if (response.json) {
+        this.csrfToken = response.json.csrf_token;
+        if (typeof this.csrfToken === 'string') {
+          this.headers = { ...this.headers, 'X-CSRFToken': this.csrfToken };
```
reklawnos: Adding some types to guarantee that `headers` includes the right properties would probably be pretty helpful. Are there certain headers that are always passed on every request? Maybe something like:
```typescript
type RequiredHeaders = {
  'X-CSRFToken': string;
  // ...
};
type Headers = { [k: string]: string } & RequiredHeaders;
```

One nice side-effect of this is that there's less of a need for the `csrfToken` type alias. As long as the type for the final `headers` object that gets passed into `fetch` is correct, TS can guarantee that all the types that might populate that field are correct. Then it doesn't require as much care to make sure that `crsfToken` is always the same type for all of the request types.

williaster: only `'X-CSRFToken'`is required for now, but this a great idea, thanks for the suggestion!

packages/superset-ui-connection/types/SupersetClient/index.d.ts
williaster: whoaaa had not heard of `Partial` or `Pick` 🤯

packages/superset-ui-connection/src/SupersetClient.ts
```diff
@@ -0,0 +1,200 @@
+import { url } from 'inspector';
+import callApi from './callApi';
+
+class SupersetClient {
+  private credentials: credentials;
+
```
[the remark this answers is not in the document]
williaster: 👍 this came from some `eslint` or `prettier` config

packages/superset-ui-connection/types/SupersetClient/index.d.ts
williaster: 👍by "regular module" you mean within `src/` and using `import`/`export`? 

If I left it in the global namespace, and someone imported this library, would the behavior be that it would conflict with any global/local type declarations they had similar to a JS global variable?

packages/superset-ui-connection/src/callApi/parseResponse.ts
williaster: ah, that makes more sense 👌

williaster:
Linting + tests should be good. One remaining thing I'm not sure about is whether we should export types explicitly in the package `index.js` e.g., `export * from './types'`. Users can find them in `lib/types.d.ts` if they need them 🤷‍♀️ could start with that.

packages/superset-ui-connection/src/SupersetClient.ts
williaster: For now I think we'll punt on required headers vs not, this is a pretty generic client for the time being.

PR #42, opened 2018-11-29
Use factory to create D3 number formatter in @superset-ui/number-format 

kristw:
🏠 Internal

- Instead of creating `D3NumberFormatter` as subclass of `NumberFormatter` with different constructor, use a factory pattern to create `NumberFormat` with the given `d3` format string. This seems like a better API since `D3NumberFormatter` is not taking advantage of the inheritance and also defining a subclass that has quite different constructor from parent can be confusing. The factory function fits better with this context. `time-format` is also written in the factory style.


packages/superset-ui-number-format/test/factories/createD3NumberFormatter.test.js
```diff
@@ -0,0 +1,49 @@
...
+      });
+      describe('if it is valid d3 formatString', () => {
+        it('uses d3.format(config.formatString) as format function', () => {
+          const formatter = createD3NumberFormatter({ formatString: '.2f' });
+          expect(formatter.format(100)).toEqual('100.00');
+        });
+      });
+      describe('if it is invalid d3 formatString', () => {
+        it('The format function displays error message', () => {
+          const formatter = createD3NumberFormatter({ formatString: 'i-am-groot' });
```
xtinec: 😆

PR #44, opened 2018-11-30
Implement ChartClient

kristw:
🏆 Enhancements

Create `ChartClient` to facilitate calls to various APIs necessary for embeddable charts. 

Refer to this [design doc](https://docs.google.com/document/d/1nKnBxKTOrHW1bqidOE-9cMPH7NZFAYDqLh8DAcVcGVA/edit?usp=sharing)

Working on unit tests.


williaster:
yeah, I guess we should align on 
1) should all **new** code in `@superset-ui` be TS?
2)  is this a requirement for all code that will be moved from Superset to Superset-UI? 

I would generally support `1)`, but `2)` could be a lot of work (similar to trying to get 100% test coverage on all the vis code).

PR #47, opened 2018-12-02
Type promises

xtinec:
🏆 Enhancements
- type rejectAfterTimeout with generics.
- fixing an import typo in tests.


xtinec:
It lets the caller of `rejectAfterTimeout` specify a more specific type of the value that is being promised. For example, think using it to limit the return value to `string` instead of `Json`.

I'm not sure what happened with our lint settings. Only noticed it as VSCode complained. Will take a closer look at our lint settings.

PR #49, opened 2018-12-04
Add `createLoadableRenderer`

kristw:
🏆 Enhancements

Add `createLoadableRenderer` from `incubator-superset`


packages/superset-ui-chart/package.json
```diff
@@ -27,6 +27,16 @@
   },
   "dependencies": {
     "@superset-ui/core": "^0.3.0",
+    "prop-types": "^15.6.2",
+    "react": "^16.4.1",
+    "react-loadable": "^5.5.0",
     "reselect": "^4.0.0"
+  },
+  "devDependencies": {
+    "enzyme": "^3.3.0",
```
williaster: you shouldn't need these dev deps/they should come from `build-config`?

packages/superset-ui-chart/package.json
```diff
@@ -27,6 +27,16 @@
   },
   "dependencies": {
     "@superset-ui/core": "^0.3.0",
+    "prop-types": "^15.6.2",
+    "react": "^16.4.1",
```
williaster: could think about extending this if there are no known compatibility issues with e.g., all of 16 and / or 15

packages/superset-ui-chart/test/components/createLoadableRenderer.test.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import { configure, shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import createLoadableRenderer from '../../src/components/createLoadableRenderer';
+
+configure({ adapter: new Adapter() });
```
williaster: this should already done in `build-config`

packages/superset-ui-chart/package.json
kristw: Revised.

packages/superset-ui-chart/test/components/createLoadableRenderer.test.jsx
williaster: I think we need to pass the `--react` in the `test` script for it to trigger.

williaster: https://github.com/williaster/build-config/blob/master/configs/jest.js#L15

PR #51, opened 2018-12-06
Add  reactify

kristw:
🏆 Enhancements

Migrate `reactify` function from `incubator-superset`


packages/superset-ui-chart/test/components/reactify.test.jsx
```diff
@@ -0,0 +1,67 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { mount } from 'enzyme';
+import reactify from '../../src/components/reactify';
```
williaster: super nit -- if you import from `../../src` we'd potentially also get linting to ensure that it's always exported.

kristw: cool

PR #55, opened 2018-12-10
Convert chart models to TS and export additional modules

kristw:
🏆 Enhancements
- Export additional modules from `@superset-ui/chart/query`

🏠 Internal
- Convert several models in `@superset-ui/chart` to TypeScript


packages/superset-ui-chart/src/@types/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@superset-ui/core';
```
williaster: what's the reasoning with adding the `@` prefix to the index file? is that best practice/standard? (tried to google for it but didn't see anything, a bit hard to search for). in `connection we have a `types.ts` file without an `@`.

packages/superset-ui-chart/src/models/ChartMetadata.ts
```diff
@@ -0,0 +1,55 @@
+interface LookupTable {
+  [key: string]: boolean;
+}
+
+interface ChartMetadataConfig {
```
williaster: should we export the interface for consumers?

packages/superset-ui-chart/src/@types/index.d.ts
kristw: This is to manually declare types and fix the error for external module that do not have types. 

```
error TS7016: Could not find a declaration file for module '@superset-ui/core'
```

https://medium.com/@chris_72272/migrating-to-typescript-write-a-declaration-file-for-a-third-party-npm-module-b1f75808ed2

packages/superset-ui-chart/src/models/ChartMetadata.ts
kristw: ah right.

packages/superset-ui-chart/src/@types/index.d.ts
kristw: Seems like it works if I just move this file to `src` with any name but extension `.d.ts` as well.

packages/superset-ui-chart/src/models/ChartPlugin.ts
```diff
@@ -0,0 +1,79 @@
...
+      metadata,
+      buildQuery,
+      loadBuildQuery,
+      transformProps = IDENTITY,
+      loadTransformProps,
+      Chart,
+      loadChart,
+    } = config;
+    this.metadata = metadata;
+    this.loadBuildQuery = loadBuildQuery || (buildQuery ? () => buildQuery : null);
```
williaster: not sure if we should be consistent with `null`s vs `undefined`. previously I was a fan of `null`s but our TS style guide says

> Use undefined in general
>   *Why?* Compatible with optionals (the ? operator). Triggers default argument initializers. 
> Use null where its a part of the API or conventional
>  e.g., optional wouldn't be appropriate if you know a server will respond with a string or null / won't be omitted

kristw: Good point. @conglei added this part but I think either `undefined` or `null` would work for him. Will change to `undefined`

packages/superset-ui-chart/src/query/Metric.ts
```diff
@@ -1,5 +1,5 @@
 import { Column } from './Column';
-import FormData from './FormData';
+import { FormData } from './FormData';
 
 export const LABEL_MAX_LENGTH = 43;
```
xtinec: Why `.filter(metric => metric)` at line 60? (sorry github won't let me give a comment on that line).

Also, I found a bug in my original implication when putting the demo yesterday (metrics in form_data can appear as arrays of metrics). I can put up a fix against this branch.

kristw: Filter the undefined field

PR #56, opened 2018-12-11
[build] fix typescript builds

williaster:
🐛 Bug Fix

This fixes a couple issues with the build for `TS`-containing modules.  

- adds appropriate `.ts,.tsx` to the `build:esm` and `build:cjs` scripts so that `babel` will actually transpile `typescript` files 
  - previously it was ignoring them which meant 
      1) we **did not** generate any `esm/` files from `.ts` files 
      2) we **did** generate `lib/` files from `.ts`, but this was coming from the `typescript` cli (which is not what we want)
  - we use [`@babel/preset-typescript`](https://babeljs.io/docs/en/babel-preset-typescript) in our build config, but adding these extensions is required. note that the future version of `build-config` (not `@data-ui`) will do this for us

- removes the `babel:ts` script in favor of two different ones:
  - `type` will validate types without generating any files
  - `type:dts` will generating `lib/*.d.ts` files. note that it specifically does **not** create `.js` files, this is handled by `babel`

- adds the `type` script / type checking to the `test` script (so it will be forced to pass in the build)

Note: I think `yarn run type` currently fails so I may need to fix some things before the build passes. Ideally we would merge this before #55 so that we ensure the new TS passes the build.

@conglei @kristw @xtinec


packages/superset-ui-connection/test/types.ts
```diff
@@ -0,0 +1 @@
+declare module 'fetch-mock';
```
kristw: Move this file to `packages/superset-ui-connection/types/external.d.ts` should resolve the lint error in CI.

williaster:
@xtinec @kristw I'm still getting the following errors when trying to convert the `@superset-ui/charts` components to `ts`, and haven't been able to debug successfully. anything glaringly wrong that you can see? or suggestions? (have pushed all changes)

`reactify.test.ts`
this makes it seem as though the way I am defining props in the generic is incorrect, as any prop I define throws (aside from `id`)

image

All instances of `shallow(...)` or `mount(...)` are throwing like this. I'm not sure it's specific to our components because I get the same error for `shallow(<div />)`

image

packages/superset-ui-connection/types/external.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'fetch-mock';
```
xtinec: I think I need to add `@types/fetch-mock` as dev dependency to get tests to build.

packages/superset-ui-chart/src/components/createLoadableRenderer.ts
```diff
@@ -0,0 +1,52 @@
...
+    componentDidUpdate() {
+      this.afterRender();
+    }
+
+    afterRender() {
+      const { loaded, loading, error } = this.state;
+      const { onRenderFailure, onRenderSuccess } = this.props;
+      if (!loading) {
+        if (error) {
+          (onRenderFailure as Function)(error);
```
williaster: casting here because it doesn't catch that they will be defined from `defaultProps`. 

I initially had a check that they were defined, but then branch coverage will never be 100% because they'll always be there (unless we force an invalid prop value of `null`, which we could just `@ts-ignore`)

packages/superset-ui-chart/test/clients/ChartClient.test.ts
```diff
@@ -32,56 +32,52 @@ describe('ChartClient', () => {
   });
 
   describe('.loadFormData({ sliceId, formData }, options)', () => {
+    const sliceId = 123;
     it('fetches formData if given only sliceId', () => {
-      fetchMock.get('glob:*/api/v1/formData/?slice_id=123', {
+      fetchMock.get(`glob:*/api/v1/formData/?slice_id=${sliceId}`, {
         form_data: {
           granularity: 'minute',
-          field1: 'abc',
```
williaster: removed invalid `formData` fields

PR #62, opened 2018-12-13
Convert core to TS

kristw:
🏠 Internal
- Convert all code in `@superset-ui/core` to TypeScript
- Add generic definition for `Registry`


packages/superset-ui-core/src/models/Registry.ts
```diff
@@ -59,14 +80,14 @@ export default class Registry<V> {
 
   registerLoader(key: string, loader: () => V | Promise<V>) {
```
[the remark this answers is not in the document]
kristw: `ItemWithLoader` is the stored item in the lookup which is `{ loader }` while the parameter here is `loader`.

kristw:
Run into issue with `eslint-plugin-typescript` `no-unused-vars` rule with generic.
Might be related to this issue: https://github.com/bradzacher/eslint-plugin-typescript/issues/179

packages/superset-ui-core/src/models/Preset.ts
```diff
@@ -0,0 +1,34 @@
+import Plugin from './Plugin';
+
+export interface PresetConfig {
+  name?: string;
+  description?: string;
+  presets?: Preset[];
+  plugins?: Plugin[];
+}
+
+export class Preset {
```
williaster: I thought we decided on default exports for classes/components, etc. and named exports for types/interfaces?

packages/superset-ui-core/src/models/Registry.ts
```diff
@@ -108,26 +144,26 @@ export default class Registry {
     );
   }
 
-  keys() {
+  keys(): string[] {
     return Object.keys(this.items);
   }
 
-  values() {
+  values(): Array<V | Promise<V> | undefined> {
```
williaster: looks like you can also do `(type | type | type)[]`, the `Array<..>` generic seems a little more readable for this case tho so 🤷‍♂️

kristw: I will keep the `Array<>` as it is more readable. Will use `[]` for primitive and types.

PR #63, opened 2018-12-13
[Chart]Unify Metric format

conglei:
💔 Breaking Changes

🏆 Enhancements
Added expressionType to built-in metric to unify the format of all metrics. 

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart/src/query/Metric.ts
+++ b/packages/superset-ui-chart/src/query/Metric.ts
@@ -28,32 +28,35 @@ export enum Aggregate {
 export enum ExpressionType {
   SIMPLE = 'SIMPLE',
   SQL = 'SQL',
+  BUILTIN = 'BUILTIN',
 }
 
-interface AdhocMetricSimple {
+interface MetricSimple {
   expressionType: ExpressionType.SIMPLE;
   column: Column;
   aggregate: Aggregate;
-}
-
-interface AdhocMetricSQL {
-  expressionType: ExpressionType.SQL;
-  sqlExpression: string;
-}
-
-export type AdhocMetric = {
   label?: string;
   optionName?: string;
-} & (AdhocMetricSimple | AdhocMetricSQL);
+}
+
+interface MetricSQL {
+  expressionType: ExpressionType.SQL;
+  sqlExpression: string;
+  label?: string;
+  optionName?: string;
+}
+
+interface MetriBuiltin {
+  expressionType: ExpressionType.BUILTIN;
+  label: string;
+}
 
 // Type of metrics in form data
-export type FormDataMetric = string | AdhocMetric;
+export type FormDataMetric = string | MetricSQL | MetricSimple;
 
 // Type of Metric the client provides to server after unifying various forms
 // of metrics in form data
-export type Metric = {
-  label: string;
-} & Partial<AdhocMetric>;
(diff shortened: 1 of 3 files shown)


packages/superset-ui-chart/src/query/Metric.ts
```diff
@@ -26,34 +26,37 @@ export enum Aggregate {
 }
 
 export enum ExpressionType {
+  BUILTIN = 'BUILTIN',
   SIMPLE = 'SIMPLE',
   SQL = 'SQL',
 }
 
-interface AdhocMetricSimple {
+interface MetricSimple {
```
kristw: How about `SimpleMetric` and `SQLMetric`? But isn't everything through SQL? What does SQL Metric mean?

packages/superset-ui-chart/src/query/Metric.ts
```diff
@@ -26,34 +26,37 @@ export enum Aggregate {
...
   sqlExpression: string;
-}
-
-export type AdhocMetric = {
   label?: string;
   optionName?: string;
-} & (AdhocMetricSimple | AdhocMetricSQL);
+}
+
+interface MetriBuiltin {
```
kristw: Typo. Also `BuiltInMetric` may be better (capital `I`)

packages/superset-ui-chart/src/query/Metric.ts
conglei: So `SimpleMetric` is for the metrics that are generated only by an aggregation operator and a column. For `SqlMetric`, it is for a metric that with a SQL query specified. 

I don't have a strong opinion on `SimpleMetric` or `MetricSimple`.  either works for me 👍

xtinec: I like `SimpleMetric`, `SQLMetric` and `BuiltInMetric`. 👍

packages/superset-ui-chart/src/query/Metric.ts
```diff
@@ -26,34 +26,37 @@ export enum Aggregate {
...
   SIMPLE = 'SIMPLE',
   SQL = 'SQL',
 }
 
-interface AdhocMetricSimple {
+interface MetricSimple {
   expressionType: ExpressionType.SIMPLE;
   column: Column;
   aggregate: Aggregate;
+  label?: string;
```
xtinec: Totally cool with this more descriptive way of listing out all the optional fields of the metrics. Alternatively, we can have something like
```
interface AdditionalMetricFields {
  label?: string;
  optionName?: string;
}
``` 
And then
```
type SimpleMetric = {
  expressionType: ExpressionType.SIMPLE;
  column: Column;
  aggregate: Aggregate;
} & Partial<AdditionalMetricFields>
```

conglei: Actually if we use `Partial<>`, do we need to do 
```
interface AdditionalMetricFields {
  label?: string;
  optionName?: string;
}
```
instead of 
```
interface AdditionalMetricFields {
  label: string;
  optionName: string;
}
```

xtinec: Whoops, typo on my part, I meant
```
interface AdditionalMetricFields {
  label: string;
  optionName: string;
}
```
(Correcting the original comment as well)
Partial makes a subset of the keys (including all keys) optional.

PR #66, opened 2018-12-14
[chart plugin]TableVis Plugin

conglei:
💔 Breaking Changes

🏆 Enhancements
This is to bring tablevis as a plugin. 

📜 Documentation

🐛 Bug Fix

🏠 Internal


packages/superset-ui-plugin-table/src/Table.js
```diff
@@ -0,0 +1,236 @@
+import d3 from 'd3';
+import $ from 'jquery';
+import PropTypes from 'prop-types';
+import dt from 'datatables.net-bs';
+import 'datatables.net-bs/css/dataTables.bootstrap.css';
```
williaster: we may not be able to import this type of file in babel. 

if we refactor `datatables` and still need to import `css`, we could instead copy it in the lib output, and have incubator superset import (since it has proper css loader support)

PR #68, opened 2018-12-17
Add SuperChart

kristw:
🏆 Enhancements

- Migrate `SuperChart` from `incubator-superset`, convert to TypeScript and add 100% unit tests.
- Add default values for `ChartProps`

🏠 Internal

- Change to `export default` for some modules.

### DIFF
--- a/packages/superset-ui-chart/src/components/SuperChart.tsx
+++ b/packages/superset-ui-chart/src/components/SuperChart.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { createSelector } from 'reselect';
-import getChartComponentRegistry from '../registries/ChartBuildQueryRegistrySingleton';
+import getChartComponentRegistry from '../registries/ChartComponentRegistrySingleton';
 import getChartTransformPropsRegistry from '../registries/ChartTransformPropsRegistrySingleton';
 import ChartProps from '../models/ChartProps';
 import createLoadableRenderer, { LoadableRenderer } from './createLoadableRenderer';
@@ -161,7 +161,7 @@ class SuperChart extends React.PureComponent<SuperChartProps, {}> {
         <div className="alert alert-warning" role="alert">
           <strong>ERROR</strong>&nbsp;
           <code>chartType=&quot;{chartType}&quot;</code> &mdash;
-          {JSON.stringify(error)}
+          {error.toString()}
         </div>
       );
     }
@@ -191,8 +191,19 @@ class SuperChart extends React.PureComponent<SuperChartProps, {}> {
       return null;
     }
 
+    const containerProps: {
+      id?: string;
+      className?: string;
+    } = {};
+    if (id) {
+      containerProps.id = id;
+    }
+    if (className) {
+      containerProps.className = className;
+    }
+
     return (
-      <div id={id} className={className}>
+      <div {...containerProps}>
         <Renderer
           preTransformProps={preTransformProps}
           postTransformProps={postTransformProps}
--- a/packages/superset-ui-chart/src/components/SuperChart.tsx
+++ b/packages/superset-ui-chart/src/components/SuperChart.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { createSelector } from 'reselect';
 import getChartComponentRegistry from '../registries/ChartBuildQueryRegistrySingleton';
 import getChartTransformPropsRegistry from '../registries/ChartTransformPropsRegistrySingleton';
-import { ChartProps } from '../models/ChartProps';
+import ChartProps from '../models/ChartProps';
 import createLoadableRenderer, { LoadableRenderer } from './createLoadableRenderer';
 
 const IDENTITY = (x: any) => x;
@@ -13,6 +13,7 @@ const EMPTY = () => null;
 const defaultProps = {
   id: '',
   className: '',
+  chartProps: {},
   preTransformProps: IDENTITY,
   overrideTransformProps: undefined,
   postTransformProps: IDENTITY,
@@ -42,7 +43,7 @@ interface RenderProps {
 export interface SuperChartProps {
   id?: string;
   className?: string;
-  chartProps: ChartProps;
+  chartProps?: ChartProps;
   chartType: string;
   preTransformProps?: TransformFunction;
   overrideTransformProps?: TransformFunction;
--- a/packages/superset-ui-chart/src/components/SuperChart.tsx
+++ b/packages/superset-ui-chart/src/components/SuperChart.tsx
@@ -1,0 +1,198 @@
+import * as React from 'react';
+import { createSelector } from 'reselect';
+import getChartComponentRegistry from '../registries/ChartBuildQueryRegistrySingleton';
+import getChartTransformPropsRegistry from '../registries/ChartTransformPropsRegistrySingleton';
+import { ChartProps } from '../models/ChartProps';
+import createLoadableRenderer from './createLoadableRenderer';
+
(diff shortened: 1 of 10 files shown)


packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+export interface SuperChartProps {
+  id?: string;
+  className?: string;
+  chartProps?: ChartProps | null;
+  chartType: string;
+  preTransformProps?: TransformFunction;
+  overrideTransformProps?: TransformFunction;
+  postTransformProps?: TransformFunction;
+  onRenderSuccess?: HandlerFunction;
+  onRenderFailure?: HandlerFunction;
```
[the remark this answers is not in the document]
kristw: This one is currently hooked up for logging.

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  createLoadableRenderer: (
+    input: {
+      chartType: string;
+      overrideTransformProps?: TransformFunction;
+    },
+  ) => LoadableRenderer<RenderProps, LoadedModules> | (() => null);
+
+  renderChart(loaded: LoadedModules, props: RenderProps) {
+    const Chart = getModule<typeof React.Component>(loaded.Chart);
```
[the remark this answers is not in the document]
kristw: It does. I had `<React.Component>` at first and it fails. 

There was a StackOverflow thread explaining it. `<React.Component>` will mean `Chart` is an instanceof `React.Component` but it is not what we want. `Chart` is a constructor that creates `React.Component`, not the `React.Component` instance, which is what `<typeof React.Component>` is.

packages/superset-ui-chart/src/models/ChartPlugin.ts
```diff
@@ -67,7 +67,7 @@ export class ChartPlugin extends Plugin {
     }
   }
 
-  register(): ChartPlugin {
+  register() {
```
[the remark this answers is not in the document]
kristw: Yes.

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+    // to prepare for rendering once chartProps becomes available.
+    if (chartProps === null) {
+      return null;
+    }
+
+    const containerProps: {
+      id?: string;
+      className?: string;
+    } = {};
+    if (id) {
```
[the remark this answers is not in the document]
kristw: This might still create `<div id="" className="">` eventhough `id` and `className` are `undefined`. The `if` condition ensure it will be just `<div>`, which is slightly cleaner.

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,218 @@
...
+  className: '',
+  preTransformProps: IDENTITY,
+  overrideTransformProps: undefined,
+  postTransformProps: IDENTITY,
+  onRenderSuccess() {},
+  onRenderFailure() {},
+};
+/* eslint-enable sort-keys */
+
+type TransformFunction = (x: any) => any;
```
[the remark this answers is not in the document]
kristw: Will update.

packages/superset-ui-chart/test/components/SuperChart.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { SuperChart } from '../../src';
+
+describe('SuperChart', () => {
+  it('does not render if chartType is not set', done => {
+    const wrapper = shallow(<SuperChart />);
+    setTimeout(() => {
```
xtinec: hmm, it seems we only need `setTimeout` when testing the slow chart, right?

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,222 @@
...
+import getChartComponentRegistry from '../registries/ChartComponentRegistrySingleton';
+import getChartTransformPropsRegistry from '../registries/ChartTransformPropsRegistrySingleton';
+import ChartProps from '../models/ChartProps';
+import createLoadableRenderer, { LoadableRenderer } from './createLoadableRenderer';
+
+const IDENTITY = (x: any) => x;
+
+const EMPTY = () => null;
+
+/* eslint-disable sort-keys */
```
xtinec: Might be worth-while disabling `sort-keys` in our `build-config`.

packages/superset-ui-chart/test/components/SuperChart.test.tsx
```diff
@@ -62,28 +62,28 @@ describe('SuperChart', () => {
       setTimeout(() => {
         expect(wrapper.render().find('div.test-component')).toHaveLength(1);
         done();
-      }, 5);
+      }, 0);
```
xtinec: We could remove `setTimeout` altogether on all except for slow charts, perhaps? 
```
expect(...);
done();
```

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+const BLANK_CHART_PROPS = new ChartProps();
+
+export interface SuperChartProps {
+  id?: string;
+  className?: string;
+  chartProps?: ChartProps | null;
+  chartType: string;
+  preTransformProps?: TransformFunction<ChartProps>;
+  overrideTransformProps?: TransformFunction;
```
xtinec: Should both `overrideTransformProps` and `postTransformProps ` take `ChartProps` as type arguments?

kristw: If `preTransformProps` transforms `ChartProps` into something else, that will be input to `overrideTransformProps` and/or `postTransformProps`

packages/superset-ui-chart/test/components/SuperChart.test.tsx
kristw: tried and it fails tests if I remove `setTimeout`. This might be something to do with `react-loadable`

williaster: I've found that sometimes `setTimeout(fn, 0)` is needed to get to the next tick for some lifecycle or mount-based functionality.

packages/superset-ui-chart/src/components/SuperChart.tsx
williaster: yeah I have mixed feelings on that rule, it's a pain sometimes esp when you want to logically group things like this. I'll make the change in build-config.

xtinec: 👍

packages/superset-ui-chart/test/components/SuperChart.test.tsx
xtinec: Interesting that the "requeueing" of `fn` is necessary sometimes. We can leave this as-is then.

packages/superset-ui-chart/src/components/SuperChart.tsx
xtinec: That's true. I suppose we could create a generic over the output props and use that as the input for `overrideTransformProps` and `postTransformProps`.

I'm cool leaving this as-is for now (perhaps adding a TODO to revisit?). Think we could potentially improve the types here once we have more chart plugins added.

PR #69, opened 2018-12-18
Migrate @superset-ui/color to TypeScript

kristw:
🏆 Enhancements
🏠 Internal

- Migrate `@superset-ui/color` to TypeScript


packages/superset-ui-color/src/CategoricalColorNamespace.ts
```diff
@@ -1,24 +1,31 @@
...
 import getCategoricalSchemeRegistry from './CategoricalSchemeRegistrySingleton';
 import stringifyAndTrim from './stringifyAndTrim';
 
 export default class CategoricalColorNamespace {
-  constructor(name) {
+  name: string;
+  scales: {
+    [key: string]: CategoricalColorScale;
+  };
+
```
williaster: nit -- could remove space to keep consistent with other component's types

packages/superset-ui-color/src/SequentialScheme.ts
```diff
@@ -1,7 +1,7 @@
 import { scaleLinear } from 'd3-scale';
-import ColorScheme from './ColorScheme';
+import ColorScheme, { ColorSchemeConfig } from './ColorScheme';
 
-function range(count) {
+function range(count: number): number[] {
```
williaster: can `number[]` be inferred here? (I know our style guide said rely on inference when possible, this also seems fine to me)

packages/superset-ui-core/src/utils/isRequired.ts
```diff
@@ -1,3 +1,3 @@
-export default function isRequired(field: string) {
+export default function isRequired(field: string): never {
```
williaster: hadn't seen `never` before 😮

packages/superset-ui-color/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "@superset-ui/core": "^0.7.0",
+    "@types/d3-scale": "^2.0.2",
```
williaster: wow sweet!

packages/superset-ui-core/src/utils/isRequired.ts
kristw: https://stackoverflow.com/questions/37910669/what-is-the-difference-between-never-and-void-in-typescript

PR #71, opened 2019-01-03
Handle BigNumber conversions in JSON properly (without loss of precision)

soboko:
🐛 Bug Fix

Part of fix for https://github.com/apache/incubator-superset/issues/6590

There's been an issue in SQLLab where numeric results appear inconsistently - sometimes they have quotes around them and sometimes they don't. This causes havoc with sorting, among other things.

This depends on whether the results come from a sync query or an async query. The logic in those two query handlers was handling numeric values inconsistently. In the case of sync queries, there's a potential loss of precision. In the case of sync queries, there are unwanted quote marks.

This PR mainly addresses the potential loss of precision by making sure that when a GET or POST is issued to SupersetClient (with a desired JSON response) the returned JSON contains BigNumbers. I'll submit a second PR to address the quoting issue (which is in the front end code in incubator-superset).

**Notes:**

1) I had to add a dependency on `json-bigint` (which handles the parsing).
2) Slight error-handling differences caused one of the existing unit tests to fail. I don't know if this makes the PR a non-starter but it's probably worth looking at that (`parseResponse.test.ts`).
3) I had to create a typescript declaration for `json-bigint`. I just added this to `external.d.ts` but I'm not sure if this is the proper place to do it. Maybe it's better to put somewhere where it can be used from other repos, or we might end up with code duplication... 
4) I don't know if I should be commenting the code - I don't see any comments, so I didn't add any.
5) Please LMK if you see any other convention violations. :)

to: @conglei @williaster @kristw 
cc: @john-bodley


packages/superset-ui-connection/test/callApi/parseResponse.test.ts
```diff
@@ -60,8 +60,7 @@ describe('parseResponse()', () => {
       .then(throwIfCalled)
       .catch(error => {
         expect(fetchMock.calls(mockTextUrl)).toHaveLength(1);
-        expect(error.stack).toBeDefined();
-        expect(error.message.includes('Unexpected token')).toBe(true);
+        expect(error.message.includes('Unexpected')).toBe(true);
```
williaster: can you elaborate on what the new error is? I'm not necessarily against changing the tests just want to make sure the change isn't indicative of something else.

packages/superset-ui-connection/types/external.d.ts
```diff
@@ -1 +1,8 @@
 declare module 'fetch-mock';
+
+declare module 'json-bigint' {
```
williaster: @xtinec do you have any thoughts on this type of declaration? it looks okay to me, not sure if we can be more specific than `any`.

packages/superset-ui-connection/src/callApi/parseResponse.ts
```diff
@@ -17,7 +18,13 @@ export default function parseResponse(
   } else if (parseMethod === 'text') {
     return checkedPromise.then(response => response.text().then(text => ({ response, text })));
   } else if (parseMethod === 'json') {
-    return checkedPromise.then(response => response.json().then(json => ({ json, response })));
+    return checkedPromise.then(response =>
+      response.text().then(text => {
```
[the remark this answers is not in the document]
soboko: Good point, I've changed it.

packages/superset-ui-connection/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/connection",
-  "version": "0.7.2",
+  "version": "0.7.3",
```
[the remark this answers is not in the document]
soboko: Will revert.

packages/superset-ui-connection/test/SupersetClientClass.test.ts
```diff
@@ -315,6 +317,23 @@ describe('SupersetClientClass', () => {
         );
       });
 
+      it('supports parsing a response as JSON using BigNumber', () => {
```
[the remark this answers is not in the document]
soboko: Yeah, it's probably not necessary to check the class name. I am asserting on the value in the next line. Will remove the class name check.

packages/superset-ui-connection/test/callApi/parseResponse.test.ts
soboko: The new error text is "Unexpected <" - slightly different, but probably not too important. The bigger possible concern is that `error.stack` is undefined. There is an assertion for it, do you think it's a dangerous change?

williaster: hmm, interesting. I don't think the missing `error.stack` is a huge issue, it's a little weird though because if you `const e = new Error('test')`, `e.stack` is defined so maybe it's using some other `Error` type 🤔

williaster: yeah looks like they have [their own `error` function that throws an object](https://github.com/sidorares/json-bigint/blob/master/lib/parse.js#L108) not an `Error`

soboko: Okay, I've changed the code so that if JSONbig.parse fails, I catch the "error" and rewrap the message in a proper Error, so any code in `incubator-superset` should work as expected... thoughts?

packages/superset-ui-connection/types/external.d.ts
xtinec: I think we could do `BigNumber | string` here instead of `any`. 

Looks like `json-bigint` doesn't have a type definition on DefinitelyTyped but we can add one for it pretty easily (it only has two exported methods and we already wrote the type for one of them here 🙂). Then, we could install our new type definition as a dependency. @soboko would you be interested in doing it?


P.S. Unrelated but we could also add the type def for `fetch-mock` which is already available on `DefinitelyTyped` @williaster

packages/superset-ui-connection/test/callApi/parseResponse.test.ts
williaster: sounds good!

packages/superset-ui-connection/types/external.d.ts
williaster: let us know what you think about this @soboko, this could be done in a separate PR if you want to get this merged.

soboko: I used `any` as the return type for consistency with the return type of `JSON.parse`. I did just add the second method and stole some comments from `lib.es5.d.ts`... 

I agree with adding a type def for `json-bigint` to DefinitelyTyped is the right thing to do and I can certainly do that. Looks like it could take up to a week for that PR to be accepted. Perhaps we could still merge this PR for the time being? There's a PR in `incubator-superset` that's dependent on this one.

I also added the type def for `fetch-mock` in my latest commit.

packages/superset-ui-connection/test/SupersetClientClass.test.ts
```diff
@@ -125,6 +126,7 @@ describe('SupersetClientClass', () => {
     it('does not set csrfToken if response is not json', () => {
       fetchMock.get(LOGIN_GLOB, '123', {
         overwriteRoutes: true,
+        // @ts-ignore
```
williaster: why does this one require a `@ts-ignore`?

packages/superset-ui-connection/test/SupersetClientClass.test.ts
```diff
@@ -1,5 +1,6 @@
 import fetchMock from 'fetch-mock';
 
+import { BigNumber } from 'bignumber.js';
```
williaster: do you need the `.js` here? it shoudl auto-resolve without it.

soboko: Good point, I'll remove it - it was probably added by WebStorm.

packages/superset-ui-connection/test/SupersetClientClass.test.ts
williaster: I see, I'm cool with `#3` if we also add a `@TODO` comment for what is needed to get rid of it.

otherwise `#1` seems next best / hopefully not many breaking changes, but understand that becomes a lot of yak-shaving 🤦‍♂️

packages/superset-ui-connection/test/SupersetClientClass.test.ts
soboko: It looks like `bignumber.js` is the actual name of the module, so I can't remove the .js.

williaster: ohhhh, okay. that's ... interesting :)

packages/superset-ui-connection/types/external.d.ts
soboko: So I looked into creating a proper type def file for DefinitelyTyped, and there's a bit of a complication.

Specifically, the [API for json-bigint](https://github.com/sidorares/json-bigint) looks like:

```js
module.exports = function(options) {
    return  {
        parse: json_parse(options),
        stringify: json_stringify
    }
};
//create the default method members with no options applied for backwards compatibility
module.exports.parse = json_parse();
module.exports.stringify = json_stringify;
```

The idea is that you can construct a parser with non-default options using something like:
```js
var JSONstrict = require('json-bigint')({"strict": true});
```

AFAICT there's no clean way to reconcile the json-bigint API with ES6/TypeScript-style imports and type definition files.

I've [submitted a PR](https://github.com/sidorares/json-bigint/pull/27) that would allow using something like the following to construct a "special" JSONbig parser:

```js
const JSONbigspecial = JSONbig.newInstance({ storeAsString: true, strict: true });
```

If that PR is merged, then I can submit a decent type declaration file to DefinitelyTyped and problem solved.

packages/superset-ui-connection/test/SupersetClientClass.test.ts
soboko: Ok, I stuck with #3 for now and added comments + an `@TODO`

williaster:
hmm not sure why the travis branch is stuck 🤔

PR #77, opened 2019-01-12
Jest/TS config adjustments (allow for debugging tests in WebStorm)

soboko:
🏆 Enhancements 🏠 Internal

These changes are necessary (AFAIK) for setting up a Jest configuration in WebStorm that allows for setting breakpoints while running tests. Tests still work the same from the command line.

Such a config looks like this for one test file:

image

...and for all tests:

image

to: @conglei @kristw @williaster

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "husky": "^1.1.2",
     "lerna": "^3.2.1",
     "lint-staged": "^8.0.4",
+    "ts-jest": "^23.10.5",
     "yarn": "^1.9.4"
   },
   "engines": {
@@ -60,6 +61,13 @@
       "prettier",
       "typescript"
     ],
+    "typescript": {
+      "include": [
+        "./packages/**/src/**/*",
+        "./packages/**/types/**/*",
+        "./packages/**/test/**/*"
+      ]
+    },
     "jest": {
       "testPathIgnorePatterns": [
         "<rootDir>/packages/generator-superset"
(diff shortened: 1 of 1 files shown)


williaster:
Hey @soboko thanks for looking into this. All of our lib configuration lives in [`@data-ui/build-config`](https://github.com/williaster/build-config) so updates should go there.

That said -- the intention with `build-config` is to be replaced entirely with an airbnb open-source build config that has exact parity (sometime in Q1), so I’d like to not deviate far from that one (which does not use `ts-jest`). So I'm not sure about the best way around it, internally they would recommend using `vs code`, but you can ask in our internal #typescript channel to see if it's possible to set this up in Webstorm without `ts-jest`. Is there a reason you need to step through in an IDE vs the browser source view?

soboko:
> Hey @soboko thanks for looking into this. All of our lib configuration lives in [`@data-ui/build-config`](https://github.com/williaster/build-config) so updates should go there.
> 
> That said -- the intention with `build-config` is to be replaced entirely with an airbnb open-source build config that has exact parity (sometime in Q1), so I’d like to not deviate far from that one (which does not use `ts-jest`). So I'm not sure about the best way around it, internally they would recommend using `vs code`, but you can ask in our internal #typescript channel to see if it's possible to set this up in Webstorm without `ts-jest`. 

Hmm, I can always keep this stashed locally and apply it as needed, not a huge deal.

> Is there a reason you need to step through in an IDE vs the browser source view?

Mainly for running tests.

PR #78, opened 2019-01-22
Convert time-format to TypeScript

kristw:
🏆 Enhancements

- Convert `@superset-ui/time-format` to TypeScript

💔 Breaking Changes

- BREAKING CHANGE: Make formatter always returns `string`. This is different from previous behavior.

|Value|formatted value (before)|formatted value (after)|
|------------|------------|-----------|
| `null` | `null` | `'null'` |
| `undefined` | `undefined` | `'undefined'` |
| `NaN` | `NaN` | `'NaN'` |


mistercrunch:
Are we sure there are no truthy evaluation of what comes out of this function in the Superset codebase? Wouldn't it be safer to either return either `null` or a string, or an falsy empty string?

packages/superset-ui-time-format/src/utils.ts
```diff
@@ -29,23 +29,24 @@ import {
   utcSaturday,
   utcMonth,
   utcYear,
+  CountableTimeInterval,
```
williaster: nice

packages/superset-ui-time-format/test/factories/createD3TimeFormatter.test.js
```diff
@@ -1,32 +1,8 @@
 import createD3TimeFormatter from '../../src/factories/createD3TimeFormatter';
-import { PREVIEW_TIME } from '../../src/TimeFormatter';
-import { TimeFormats } from '../../src';
 
 describe('createD3TimeFormatter(config)', () => {
   it('requires config.formatString', () => {
```
williaster: any reason to not just use `@ts-ignore` and have a single (`.ts`) file?

packages/superset-ui-time-format/test/TimeFormatter.test.js
```diff
@@ -1,9 +1,14 @@
-import TimeFormatter, { PREVIEW_TIME } from '../src/TimeFormatter';
+import TimeFormatter from '../src/TimeFormatter';
 
 describe('TimeFormatter', () => {
```
williaster: any reason to not just use `@ts-ignore` and have a single (`.ts`) file?

kristw:
My understanding of what @mistercrunch means is when people do sth like 

```js
formatTime(time, 'formatString') || 'custom label when null or undefined'
```

This will work if the `formatTime()` function returns truthy/falsy. My rationale for making it string was to make the function complete in itself without relying on `|| 'fallback'`. If dev want to manually handle labels for `null` or `undefined` then they can add stricter check before calling `formatTime()`.

packages/superset-ui-time-format/test/factories/createD3TimeFormatter.test.js
kristw: This was written before I know about `@ts-ignore`. Will update.

PR #81, opened 2019-01-24
[WIP] Bundle jed TS declaration with translation module

kristw:
🏆 Enhancements

- Bundle TypeScript declaration for package `jed` with `@superset-ui/translation` distribution. 

**Considered alternatives**
- Have every package that uses `@superset-ui/translation` and TypeScript write a Jed declaration. -- impractical.
- Create a complete declaration file for `jed` and publish to `@types/jed`. -- too much work for the use case at the moment.


kristw:
Replaced by #82

PR #85, opened 2019-01-29
[SupersetClient] implement put and delete HTTP methods

mistercrunch:
🏆 Enhancements
Trying to use METHOD = DELETE

### DIFF
--- a/packages/superset-ui-connection/src/SupersetClient.ts
+++ b/packages/superset-ui-connection/src/SupersetClient.ts
@@ -19,6 +19,7 @@ export interface SupersetClientInterface {
   isAuthenticated: () => boolean;
   post: (request: RequestConfig) => Promise<SupersetClientResponse>;
   reAuthenticate: () => Promise<string | undefined>;
+  request: (request: RequestConfig) => Promise<SupersetClientResponse>;
   reset: () => void;
 }
 
@@ -34,6 +35,7 @@ const SupersetClient: SupersetClientInterface = {
   isAuthenticated: () => getInstance(singletonClient).isAuthenticated(),
   post: (request: RequestConfig) => getInstance(singletonClient).post(request),
   reAuthenticate: () => getInstance(singletonClient).init(/* force = */ true),
+  request: (request: RequestConfig) => getInstance(singletonClient).get(request),
   reset: () => {
     singletonClient = undefined;
   },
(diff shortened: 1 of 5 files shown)


packages/superset-ui-connection/src/SupersetClient.ts
```diff
@@ -34,6 +35,7 @@ const SupersetClient: SupersetClientInterface = {
   isAuthenticated: () => getInstance(singletonClient).isAuthenticated(),
   post: (request: RequestConfig) => getInstance(singletonClient).post(request),
   reAuthenticate: () => getInstance(singletonClient).init(/* force = */ true),
+  request: (request: RequestConfig) => getInstance(singletonClient).get(request),
```
williaster: I don't think we should expose a general `request` method on the public interface, there are only 4 possible methods so if you want `delete` or `put` I would just add those.

packages/superset-ui-connection/src/SupersetClientClass.ts
```diff
@@ -70,12 +70,17 @@ export class SupersetClientClass {
     return this.csrfToken !== null && this.csrfToken !== undefined;
   }
 
-  async get({
+  async get(requestConfig: RequestConfig): Promise<SupersetClientResponse> {
```
williaster: if you're refactoring `get` like this, I think you should also refactor `post`, and then could add `delete` / `put` on top of those.

packages/superset-ui-connection/src/SupersetClient.ts
kristw: +1

packages/superset-ui-connection/src/SupersetClientClass.ts
kristw: +1

packages/superset-ui-connection/src/SupersetClient.ts
mistercrunch: Looks like there are many more methods: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

Happy to add more explicit `delete` and `put` as they are most common. Can I leave `request` just in case? If someone has an edge case where they need to use one of the more obscure methods, it sucks to have to come and add it here...

mistercrunch:
@williaster I just created an account here: https://www.npmjs.com/~mistercrunch

williaster:
sent an invite 👌 Sounds like this should be included in the release @kristw just made.

PR #86, opened 2019-01-29
Add more subgenerators to create package and demo in superset-ui-legacy

kristw:
🏆 Enhancements
- Add more subgenerators to create package and demo in superset-ui-legacy
- Make the default generator show choices to select subgenerator

### DIFF
--- a/packages/generator-superset/generators/legacy-plugin-chart/templates/_package.json
+++ b/packages/generator-superset/generators/legacy-plugin-chart/templates/_package.json
@@ -1,0 +1,40 @@
+{
+  "name": "@superset-ui/legacy-plugin-chart-<%= packageName %>",
+  "version": "0.0.0",
+  "description": "Superset Legacy Chart - <%= description %>",
+  "sideEffects": false,
+  "main": "lib/index.js",
+  "module": "esm/index.js",
+  "files": [
+    "esm",
+    "lib"
+  ],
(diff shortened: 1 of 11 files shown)


packages/generator-superset/generators/legacy-plugin-chart/templates/_package.json
```diff
@@ -0,0 +1,40 @@
...
+  "homepage": "https://github.com/apache-superset/superset-ui-legacy#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@superset-ui/core": "^0.9.0",
+    "prop-types": "^15.6.2"
+  },
+  "devDependencies": {
+    "@superset-ui/chart": "^0.9.0",
```
williaster: one q about generator deps (don't need to address in this PR): do these update automatically when we release new versions? if not, wonder if some type of `@superset-ui/chart@latest` could be useful. worse case scenario they have to manually update the versions 🤔

kristw: Umm, good question. These are not updated automatically. For the current use case as I am going through mass conversion, having fixed version is more helpful, but can switch to latest after things stabilize.

kristw:
Just published as `0.9.2`

PR #88, opened 2019-02-04
Add @superset-ui/dimension

kristw:
🏆 Enhancements
- New package with two functions `getTextDimension` and `computeMaxFontSize` for computing text size in BigNumber and others. 

🏠 Internal
- `jsdom` does not implement `getBBox()` so I had to implement a dummy polyfill for unit tests.


package.json
```diff
@@ -20,7 +20,7 @@
     "prettier": "beemo prettier \"./packages/*/{src,test,storybook}/**/*.{js,jsx,ts,tsx,json,md}\"",
     "release": "yarn run prepare-release && lerna publish && yarn run postrelease",
     "test": "yarn run type && yarn run jest",
-    "test:watch": "beemo create-config jest --react && jest --watch"
+    "test:watch": "yarn run lint:fix && beemo create-config jest --react && jest --watch"
```
xtinec: Good idea to fix up lint errors while tests are running. Shall we add this to the `test` task also?

packages/superset-ui-dimension/src/getTextDimension.ts
```diff
@@ -0,0 +1,54 @@
...
+export default function getTextDimension(
+  input: GetTextDimensionInput,
+  defaultDimension = DEFAULT_DIMENSION,
+) {
+  const { text, className, style = {}, container = document.body } = input;
+
+  const textNode = document.createElementNS(SVG_NS, 'text');
+  textNode.textContent = text;
+
+  if (className !== undefined) {
```
xtinec: I think `className?: string` allows `className` to be both `null` and `undefined`. Do we want to handle both cases with `className != null` here?

package.json
kristw: I thought about that but decided not to because it will allow code that does not pass linting to pass CI when the CI runs `yarn test`. 

`yarn test:watch` on the other hand is only run by developer locally so that should be ok.

packages/superset-ui-dimension/test/computeMaxFontSize.test.ts
```diff
@@ -0,0 +1,54 @@
...
+      ).toEqual(20);
+    });
+    it('computes maximum font size for given maxWidth and idealFontSize', () => {
+      expect(
+        computeMaxFontSize({
+          maxWidth: 250,
+          idealFontSize: 40,
+          text: 'sample text',
+        }),
+      ).toEqual(25);
```
xtinec: Hmm, how did we get `25`? A comment here would be nice.

packages/superset-ui-dimension/test/getTextDimension.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      expect(
+        getTextDimension({
+          text: 'sample text',
+          style: {
+            font: 'italic 700 30px Lobster',
+          },
+        }),
+      ).toEqual({
+        height: 30,
+        width: 1125,
```
xtinec: Would also be nice to have a comment here for how we arrived at `1125` (`250 * (30/20) * 1.5 * 2`).

kristw: done

packages/superset-ui-dimension/test/computeMaxFontSize.test.ts
kristw: done

packages/superset-ui-dimension/src/getTextDimension.ts
kristw: done

packages/superset-ui-dimension/src/computeMaxFontSize.ts
```diff
@@ -0,0 +1,47 @@
...
+    return getTextDimension({
+      ...rest,
+      style: { ...style, fontSize },
+    });
+  }
+
+  let textDimension = computeDimension(size);
+
+  // Decrease size until textWidth is less than maxWidth
+  if (maxWidth !== undefined) {
```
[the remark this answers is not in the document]
kristw: done

package.json
williaster: agree 👍

Issue #104, opened 2019-02-17
superset-ui-connection: Allow callAPI to handle PUT and PATCH requests to support superset SQLlab


fabiomendescom:
**Is your feature request related to a problem? Please describe.**
The call API in super ui connection component currently only supports the POST command. SQLlab in superset, when saving queries, executes a PUT request to save queries. Because that function uses this component, this component needs to be changed to allow for PUT and PATCH to be executed as well. By support I mean parsing the payload in json into form variables for the request. The call API only does this for POST only. See this for the problem in superset -> https://github.com/apache/incubator-superset/issues/5804 

**Describe the solution you'd like**
The call API should work exactly the same (parsing JSON payload into form parameters) for PUT and PATCH the same way

**Describe alternatives you've considered**
SQLlab is already developed with SQLAlchemy to use these other verbs automatically in the framework. Changing this outside of the callAPI would be too cumbersome and unnecessary, considering that this support should exist in this component.

**Additional context**
The change is fairly simple. The call API should simply allow these additional verses to be used. See the

fabiomendescom:
https://github.com/apache-superset/superset-ui/pull/105 pull request created

kristw:
related to #85

kristw:
Released with `@superset-ui/connection` `0.10.2`

PR #105, opened 2019-02-17
feat(superset-ui-connection): Allow PUT and PATCH in call API

fabiomendescom:
🏆 Enhancements

Feat: Support `PUT` and `PATCH`. Address #104


package.json
```diff
@@ -94,5 +94,8 @@
       "yarn run prettier --write",
       "git add"
     ]
+  },
```
kristw: Is this intentional?

williaster: yeah shouldn't be there in a `package.json`. if it's auto-formatting you could call `npm run lint` from the monorepo root (which should generate eslint + prettier config) and it hopefully will stop that.

williaster: yeah this should be installed via the `travis.yml`?

kristw:
@fabiomendescom Your changes look good. Please address the `codecov` issue above and we can merge.

fabiomendescom:
> @fabiomendescom Your changes look good. Please address the `codecov` issue above and we can merge.

I'm not sure what I need to address. It says 100% coverage. What exactly do I need to do?

package.json
fabiomendescom: Im a bit confused about this. This file was not modified in this PR. What do I need to do?

williaster: just remove this added dependency, it seems to have found it's way into the commit/PR even if you didn't intend it to?

PR #108, opened 2019-02-20
Feat: Set default color scheme for the color registries

kristw:
🏆 Enhancements

- At the moment, using `@superset-ui/color` has some friction because developers always need to add color schemes to it before using. Otherwise, it will crash. This PR registers one color scheme to each of the color (`Sequential`, `Categorical`) registries so the registries will always have at least one scheme to use and work out-of-the-box.


xtinec:
`codecov` is giving us grief though 😢

kristw:
np. I'll take care of that. Should have added tests anyway.

PR #109, opened 2019-02-20
Add SMART_NUMBER formatter and make it default

kristw:
### Motivation
From https://github.com/apache/incubator-superset/issues/6916 and past encounters with number formatting, Superset's default `D3`'s SI formatter works ok most of the time but have some too scientific outputs which confuses general users such as 

* Format billions as `xxxG` (`G = Giga`) instead of `B` which people associate with billions and are more familiar with. 
`1,000,000,000 => '1G'`
P.S. Luckily, millions and Mega have the same `M` abbreviation.
* Format floating point using `micro (m)` which can be confused with `m` for minutes when the users are unaware of the context. 
`0.01 => '10m'`
* The SI formatter requires number of precision point as parameter. The default is `3`. This cause unnecessary rounding for small numbers. 
`123.45 => '123'`

Superset users therefore often have to define custom formatting or in the billion case, have to tolerate it. 

### Changes

This PR addresses the issues above and provide more humanized outputs for these edge cases, so the default number formatter should work most of the time out-of-the-box. It will reduce the amount of custom formatting need and improve user experience. 
* `1,000,000,000 => '1B'`
* `0.01 => '0.01'`
* `123.45 => '123.45'`

💔 Breaking Changes

* Rename constants in `NumberFormats` changing `_CHANGE` to `_SIGNED`.
* Default number format is now `SMART_NUMBER` instead of `D3` `.3~s`.

🏆 Enhancements

* Add new number format factory `createSmartNumberFormatter`.


mistercrunch:
We should add a note to `UPDATING.md` when we bump this on the Superset side.

packages/superset-ui-number-format/src/factories/createSmartNumberFormatter.ts
```diff
@@ -0,0 +1,48 @@
...
+export default function createSmartNumberFormatter(
+  config: {
+    description?: string;
+    signed?: boolean;
+    id?: string;
+    label?: string;
+  } = {},
+) {
+  const { description, signed = false, id, label } = config;
+  const siFormatter = d3Format(`.3~s`);
```
williaster: we could move these outside the function so they aren't created every call? looks like they don't ever change so no reason not to?

PR #114, opened 2019-03-06
feat: add mergeMargin() function

kristw:
🏆 Enhancements

* Add `mergeMargin()` function for combining two margins.

### DIFF
--- a/packages/superset-ui-dimension/src/mergeMargin.ts
+++ b/packages/superset-ui-dimension/src/mergeMargin.ts
@@ -8,17 +8,12 @@ export default function mergeMargin(
   const { top = 0, left = 0, bottom = 0, right = 0 } = margin1;
   const { top: top2 = 0, left: left2 = 0, bottom: bottom2 = 0, right: right2 = 0 } = margin2;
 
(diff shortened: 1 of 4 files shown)


packages/superset-ui-dimension/src/mergeMargin.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default function mergeMargin(
+  margin1: Partial<Margin> = {},
+  margin2: Partial<Margin> = {},
+  mode: 'expand' | 'shrink' = 'expand',
+) {
+  const { top = 0, left = 0, bottom = 0, right = 0 } = margin1;
+  const { top: top2 = 0, left: left2 = 0, bottom: bottom2 = 0, right: right2 = 0 } = margin2;
+
+  return mode === 'expand'
```
williaster: you could simplify this a bit

```
const func = mode === 'expand' ? Math.max : Math.min;

return {
  bottom: func(bottom, bottom2),
  ...repeat
};
```

kristw: 👍

PR #119, opened 2019-03-14
feat: add support for conditional requests

betodealmeida:
🏆 Enhancements

This PR adds support for conditional `GET` requests, since they're not supported by the Fetch API. This is how it works:

1. If a `GET` request is successful (status `200 OK`) and it has an `ETag` header, a copy of the response is stored using the [Cache API](https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/cache-api).
2. The next time a request for the same URL is made, the cache is checked.
3. If the cache has a cached request, its `ETag` is retrieved.
4. The client does a request with the header `If-None-Matches` with the stored `ETag`.
5. If the server returns a status `304 Not Modified`, the cached response is used instead.
6. Otherwise, if the response is a `200 OK` the previous cache is deleted and the new response is stored.

This will be used with https://github.com/apache/incubator-superset/pull/7032 to allow for conditional requests in Superset.

I started writing unit tests, but the Cache API is not supported neither in Jest (https://github.com/facebook/jest/issues/7365) nor in jsdom (https://github.com/jsdom/jsdom/issues/2422). I did full end-to-end tests with Superset, verifying that resources are properly cached and retrieved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,12 @@
       "typescript"
     ],
     "jest": {
+      "globals": {
+        "caches": true
+      },
+      "setupFiles": [
+        "<rootDir>/test/setupTests.ts"
+      ],
       "testPathIgnorePatterns": [
         "<rootDir>/packages/generator-superset"
       ],
--- a/packages/superset-ui-connection/README.md
+++ b/packages/superset-ui-connection/README.md
@@ -23,6 +23,7 @@ a high-level it supports:
 - supports `GET` and `POST` requests (no `PUT` or `DELETE`)
 - timeouts
 - query aborts through the `AbortController` API
+- conditional `GET` requests using `If-None-Match` and `ETag` headers
 
 #### Example usage
 
--- a/packages/superset-ui-connection/src/callApi/callApi.ts
+++ b/packages/superset-ui-connection/src/callApi/callApi.ts
@@ -1,6 +1,10 @@
 import 'whatwg-fetch';
 import { CallApi } from '../types';
 
+const cacheAvailable = 'caches' in self;
+const NOT_MODIFIED = 304;
+const OK = 200;
+
 // This function fetches an API response and returns the corresponding json
 export default function callApi({
   body,
@@ -25,6 +29,41 @@ export default function callApi({
     redirect,
     signal,
   };
+
+  if (method === 'GET' && cacheAvailable) {
+    return caches.open('superset').then(supersetCache =>
+      supersetCache
+        .match(url)
+        .then(cachedResponse => {
+          if (cachedResponse) {
+            // if we have a cached response, send its ETag in the
+            // `If-None-Match` header in a conditional request
+            const etag = cachedResponse.headers.get('Etag');
+            if (etag) {
+              request.headers = request.headers || {};
+              request.headers['If-None-Match'] = etag;
+            }
+          }
+
+          return fetch(url, request); // eslint-disable-line compat/compat
+        })
+        .then(response => {
+          if (response.status === NOT_MODIFIED) {
+            return supersetCache.match(url).then(cachedResponse => {
+              if (cachedResponse) {
+                return cachedResponse.clone();
+              }
+              throw new Error('Received 304 but no content is cached!');
+            });
+          } else if (response.status === OK && response.headers.get('Etag')) {
+            supersetCache.delete(url);
+            supersetCache.put(url, response.clone());
+          }
+
+          return response;
+        }),
+    );
+  }
 
   if (
     (method === 'POST' || method === 'PATCH' || method === 'PUT') &&
--- a/packages/superset-ui-connection/src/callApi/callApi.ts
+++ b/packages/superset-ui-connection/src/callApi/callApi.ts
@@ -2,6 +2,7 @@
 import 'whatwg-fetch';
 import { CallApi } from '../types';
 import { CACHE_KEY, NOT_MODIFIED, OK } from '../constants';
+import invariant from 'invariant';
 
 const CACHE_AVAILABLE = 'caches' in self;
 
@@ -39,9 +40,8 @@ export default function callApi({
             // if we have a cached response, send its ETag in the
             // `If-None-Match` header in a conditional request
             const etag = cachedResponse.headers.get('Etag');
-            if (etag) {
-              request.headers = { ...request.headers, 'If-None-Match': etag };
-            }
+            invariant(etag, 'Cannot be null');
+            request.headers = { ...request.headers, 'If-None-Match': etag };
           }
 
           return fetch(url, request);
--- a/packages/superset-ui-connection/src/callApi/callApi.ts
+++ b/packages/superset-ui-connection/src/callApi/callApi.ts
@@ -2,7 +2,6 @@
 import 'whatwg-fetch';
 import { CallApi } from '../types';
 import { CACHE_KEY, NOT_MODIFIED, OK } from '../constants';
-import invariant from 'invariant';
 
 const CACHE_AVAILABLE = 'caches' in self;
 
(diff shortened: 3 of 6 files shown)


betodealmeida:
@williaster, @kristw any thoughts on the unit tests? I know we're aiming for 100% coverage.

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -1,6 +1,10 @@
 import 'whatwg-fetch';
 import { CallApi } from '../types';
 
+const cacheAvailable = 'caches' in self;
+const NOT_MODIFIED = 304;
```
williaster: response numbers should be in a separate `constants` file

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -1,6 +1,10 @@
 import 'whatwg-fetch';
 import { CallApi } from '../types';
 
+const cacheAvailable = 'caches' in self;
```
williaster: I'd `UPPER_CASE` this since it's used as a constant in this file

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -26,6 +30,41 @@ export default function callApi({
     signal,
   };
 
+  if (method === 'GET' && cacheAvailable) {
+    return caches.open('superset').then(supersetCache =>
```
williaster: the `superset` key should also be a constant in a separate file, and maybe more unique, matching the package name seems safer (even though it sounds like `Cache`s are scoped to by origin) e.g., `@SUPERSET-UI/CONNECTION`?

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -26,6 +30,41 @@ export default function callApi({
...
+            // if we have a cached response, send its ETag in the
+            // `If-None-Match` header in a conditional request
+            const etag = cachedResponse.headers.get('Etag');
+            if (etag) {
+              request.headers = request.headers || {};
+              request.headers['If-None-Match'] = etag;
+            }
+          }
+
+          return fetch(url, request); // eslint-disable-line compat/compat
```
williaster: since we have multiple occurrences of this now we should just move this to the top of the file `/* eslint compat/compat: 'off' */`

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -26,6 +30,41 @@ export default function callApi({
...
+    return caches.open('superset').then(supersetCache =>
+      supersetCache
+        .match(url)
+        .then(cachedResponse => {
+          if (cachedResponse) {
+            // if we have a cached response, send its ETag in the
+            // `If-None-Match` header in a conditional request
+            const etag = cachedResponse.headers.get('Etag');
+            if (etag) {
+              request.headers = request.headers || {};
```
williaster: could simplify to `request.headers = { ...request.headers, ['If-None-Match']: etag }`

packages/superset-ui-connection/README.md
```diff
@@ -23,6 +23,7 @@ a high-level it supports:
 - supports `GET` and `POST` requests (no `PUT` or `DELETE`)
 - timeouts
 - query aborts through the `AbortController` API
+- conditional `GET` requests using `If-None-Match` and `ETag` headers
```
williaster: I'd maybe mention `Cache` here? conditional sounds like something else to me.

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: Thanks, will do!

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: +1

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: Good point.

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: +1

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: Cool, I didn't know the spread operator worked with `null`s. Will do.

packages/superset-ui-connection/README.md
betodealmeida: "Conditional request" is the actual name of a `GET` request with a `If-None-Match` header containing the `ETag` of the resource: https://developer.mozilla.org/en-US/docs/Web/HTTP/Conditional_requests#Conditional_headers. The fact that we're using the Cache API here is an implementation detail, we could use local storage, session storage or cookies.

betodealmeida:
@williaster I'll try to use one of those new packages to write the tests.

packages/superset-ui-connection/README.md
betodealmeida: To expand on my comment: we're not simply caching, we're storing the resources and reusing them if and only if the hash hasn't changed.

betodealmeida:
@williaster I added unit tests and got the coverage up to 100%.

kristw:
Could you add another test case when `CACHE_AVAILABLE` is `false`?

package.json
```diff
@@ -66,6 +66,12 @@
       "typescript"
     ],
     "jest": {
+      "globals": {
+        "caches": true
+      },
+      "setupFiles": [
+        "<rootDir>/test/setupTests.ts"
```
williaster: I think ideally we'd fix [`build-config` to resolve `.ts` setup files](https://github.com/apache-superset/build-config/blob/master/configs/jest.js#L12) but this is okay for now.

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -1,5 +1,9 @@
+/* eslint compat/compat: 'off' */
 import 'whatwg-fetch';
 import { CallApi } from '../types';
+import { CACHE_KEY, NOT_MODIFIED, OK } from '../constants';
```
williaster: sorry only nit with the status constants is that there's no context if you read them in isolation e.g., `OK` = okay what? could we prefix with `STATUS_*` or `HTTP_*` ?

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -26,6 +30,38 @@ export default function callApi({
     signal,
   };
 
+  if (method === 'GET' && CACHE_AVAILABLE) {
```
williaster: @kristw and I were discussing whether this should be opt-in or not. currently there's no way to opt-out, though I guess if if depends on the server/backend implementing `Etag`s then maybe that's a non-issue.

any thoughts on this?

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: Fixed this.

packages/superset-ui-connection/src/callApi/callApi.ts
betodealmeida: I could add a new option to `callApi`, to turn this on/off. In the initial tests I did with dashboards and ETag, some of the dashboards load in half the time, so I think we might want to default this to be enabled.

betodealmeida:
Ping @kristw @williaster

packages/superset-ui-connection/src/callApi/callApi.ts
kristw: I have merged and added this new issue for turning conditional request on/off.
https://github.com/apache-superset/superset-ui/issues/128

Smilebags:
Hey, I'm coming here to debug an issue I'm seeing where changes to the superset code aren't reflected in changed behaviour/UI when accessing superset, and wondered whether that might be because some sort of cache is enabled during development. Deleting entire folders of JS assets (the dist folder) has no impact on the functionality of superset (even when the HTTP cache is disabled) which leads me to believe there's a backend cache being used. Any guidance/direction appreciated.

PR #121, opened 2019-03-20
chore(chart): remove faux @superset-ui/core TS declaration

williaster:
🏠 Internal

This PR removes the faux `@superset-ui/core` TS declaration in `@superset-ui/chart` (introduced before `core` was re-written in TS). All of the changes are to clean up resulting TS errors from that removal. Some notes on specifics that were done:

- Refined and consolidated some `type`s that were duplicated across files like `ChartType`, `(Pre/Post)TransformProps`, etc.. Turns out the duplication wasn't always entirely consistent though at surface level it might not be obvious.
- Added the needed generics to all of the chart plugin `registry`s
- Updated the logic for supporting `Value | { default: Value }` generally across loaders. This is now confined to `ChartPlugin` in order to simplify types elsewhere, while still supporting dynamic loading of ES modules: `() => import('foo').default`

(EDIT: The following was refactored so as not to be a breaking change)

~💔 Breaking Changes~
-  ~Removed support for `{ default: Type }` in `Registry` `loader`s, which was previously allowed due to a Chart type using `module.exports` instead of `export default`ing. This simplifies typing when using these components which I think is a win 🏆. I think worst case if this is problematic in incubator `Superset` we could update `() => import('./module.xxx')` calls to `() => import('./module.xxx').default`~

**TODO**
- [x] 💯 % coverage
- [x] fix `.default` issue .. may not be a breaking change

@kristw @conglei @xtinec

### DIFF
--- a/packages/superset-ui-chart/src/models/ChartMetadata.ts
+++ b/packages/superset-ui-chart/src/models/ChartMetadata.ts
@@ -2,6 +2,17 @@ interface LookupTable {
   [key: string]: boolean;
 }
 
+export interface ChartMetaDataConfig {
+  name: string;
+  canBeAnnotationTypes?: string[];
+  credits?: string[];
+  description?: string;
+  show?: boolean;
+  supportedAnnotationTypes?: string[];
+  thumbnail: string;
+  useLegacyApi?: boolean;
+}
+
 export default class ChartMetadata {
   name: string;
   canBeAnnotationTypes?: string[];
@@ -13,16 +24,7 @@ export default class ChartMetadata {
   thumbnail: string;
   useLegacyApi: boolean;
 
-  constructor(config: {
-    name: string;
-    canBeAnnotationTypes?: string[];
-    credits?: string[];
-    description?: string;
-    show?: boolean;
-    supportedAnnotationTypes?: string[];
-    thumbnail: string;
-    useLegacyApi?: boolean;
-  }) {
+  constructor(config: ChartMetaDataConfig) {
     const {
       name,
       canBeAnnotationTypes = [],
--- a/packages/superset-ui-chart/src/models/ChartMetadata.ts
+++ b/packages/superset-ui-chart/src/models/ChartMetadata.ts
@@ -2,7 +2,7 @@ interface LookupTable {
   [key: string]: boolean;
 }
 
-export interface ChartMetaDataConfig {
+export interface ChartMetadataConfig {
   name: string;
   canBeAnnotationTypes?: string[];
   credits?: string[];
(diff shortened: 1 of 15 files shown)


packages/superset-ui-chart/src/registries/ChartMetadataRegistrySingleton.ts
```diff
@@ -1,6 +1,7 @@
 import { Registry, makeSingleton, OverwritePolicy } from '@superset-ui/core';
+import { ChartMetaDataConfig } from '../models/ChartMetadata';
 
-class ChartMetadataRegistry extends Registry {
+class ChartMetadataRegistry extends Registry<ChartMetaDataConfig, ChartMetaDataConfig> {
```
kristw: Should this be `Registry<ChartMetadata>`?

packages/superset-ui-chart/src/models/ChartMetadata.ts
```diff
@@ -2,6 +2,17 @@ interface LookupTable {
   [key: string]: boolean;
 }
 
+export interface ChartMetaDataConfig {
```
kristw: `Meta"d"ata` to be consistent with the class name

packages/superset-ui-chart/src/registries/ChartMetadataRegistrySingleton.ts
williaster: `ChartMetadata` gives errors because most calls do not pass a full `ChartMetaData` object, this is a `Partial` for what's required and is used in / matches the constructor.

williaster:
all right, made this compatible with the `default` syntax, should be good to go 🚀

packages/superset-ui-chart/src/registries/ChartMetadataRegistrySingleton.ts
kristw: I still think it should be `Registry<ChartMetadata, ChartMetadata>` because 

* There are some additional functions that the `ChartMetadata` provides so it would be nice to ensure that the values in this registry are always the instance of this class. Otherwise devs have to check again after getting an item from registry.
* Most call I believe pass the full `ChartMetadata` instance. The only place things are registered to this registry currently is in the `ChartPlugin` file which always pass the `ChartMetadata` instance

```ts
// in `ChartPlugin.ts`
getChartMetadataRegistry().registerValue(key, this.metadata);
```

* Another place where I see it being registered is the calls in `packages/superset-ui-chart/test/clients/ChartClient.test.ts` which pass plain objects but can easily be updated to make them `ChartMetadata` instances.

williaster: really good point, updated 👍

PR #122, opened 2019-03-22
 fix(chart): import types from package, not monorepo

williaster:
🐛 Bug Fix 🏠 Internal

This PR
- fixes an import in `@superset-ui/chart` that should be from a package, not the monorepo. this caused the build to include additional directories in `0.10.5` and `0.10.6`. I confirmed that these directories are no longer generated after the fix.

- adds a flag to the `build` `npm` scripts to delete the out directory, to ensure that deleted files and directories are not included in new releases.

### DIFF
--- a/packages/superset-ui-chart/src/components/ChartDataProvider.tsx
+++ b/packages/superset-ui-chart/src/components/ChartDataProvider.tsx
@@ -1,6 +1,6 @@
 /* eslint react/sort-comp: 'off' */
 import React, { ReactNode } from 'react';
-import { SupersetClientInterface, RequestConfig } from '@superset-ui/connection/lib/types';
+import { SupersetClientInterface, RequestConfig } from '@superset-ui/connection';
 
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart/src/components/ChartDataProvider.tsx
```diff
@@ -1,6 +1,6 @@
 /* eslint react/sort-comp: 'off' */
 import React, { ReactNode } from 'react';
-import { SupersetClientInterface, RequestConfig } from '../../../superset-ui-connection/src/types';
+import { SupersetClientInterface, RequestConfig } from '@superset-ui/connection/lib/types';
```
kristw: Would this work?
```js
import { SupersetClientInterface, RequestConfig } from '@superset-ui/connection';
```

williaster: I'll try this, it's giving a TS error locally / can't find the declaration file 🤔

PR #123, opened 2019-03-25
fix: 🐛 Fixed type definition for chart plugin

conglei:
This PR is to fix the type definiton for `transformProps` and `LoadData`.

For `TransformProps`, the input type should be `ChartProps` instead of `PlainObject`. 

For `LoadData`, before, the `ChartType` was `ComponentType | FunctionComponent`, which means the no props allowed for the component, which is not correct. 

💔 Breaking Changes

🏆 Enhancements

Exported Metric definition. 

📜 Documentation

🐛 Bug Fix

In most of the cases, There will be no `PreTransformProps` defined. Thus, the input for `TransformProp

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart/src/types/Query.ts
+++ b/packages/superset-ui-chart/src/types/Query.ts
@@ -38,8 +38,8 @@ export interface PlainProps {
 
 type TransformFunction<Input = PlainProps, Output = PlainProps> = (x: Input) => Output;
 
-export type PreTransformProps = TransformFunction<ChartProps>;
-export type TransformProps = TransformFunction;
+export type PreTransformProps = TransformFunction<ChartProps, ChartProps | PlainProps>;
+export type TransformProps = TransformFunction<ChartProps | PlainProps>;
 export type PostTransformProps = TransformFunction;
 
 export type BuildQueryFunction<T extends ChartFormData> = (formData: T) => QueryContext;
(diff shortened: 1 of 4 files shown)


packages/superset-ui-chart/src/types/Query.ts
```diff
@@ -38,8 +38,8 @@ export interface PlainProps {
 
 type TransformFunction<Input = PlainProps, Output = PlainProps> = (x: Input) => Output;
 
-export type PreTransformProps = TransformFunction<ChartProps>;
-export type TransformProps = TransformFunction;
+export type PreTransformProps = TransformFunction<ChartProps, ChartProps>;
+export type TransformProps = TransformFunction<ChartProps>;
```
conglei: I tried to do like 
```
export type PreTransformProps = TransformFunction<ChartProps, ChartProps | PlainProps>;
export type TransformProps = TransformFunction<ChartProps | PlainProps>;`
```

However, in Chart plugin if I define `TransfromProps` as 

```
export default function transformProps(chartProps: ChartProps)
```

it will show the error: `Type 'PlainProps | ChartProps' is not assignable to type 'ChartProps'.`

xtinec:
🎉 Thanks for fixing this. 💪  Noticed the same thing yesterday. Beat me to the fix. 🙂

PR #124, opened 2019-03-25
feat: getter function to return color mapping

khtruong:
🏆 Enhancements

Exposed a getter function to return the color mapping between a data value and a color hex code. This is needed so we can save it as dashboard metadata for color consistency in dashboards (https://github.com/apache/incubator-superset/pull/7086).


kristw:
lgtm. waiting for ci

PR #125, opened 2019-03-28
feat: add labels to color scheme

khtruong:
🏆 Enhancements

There was an ask to add tooltips for color schemes so users can remember which is which. I added labels to the color schemes to make tooltips more reader friendly.


williaster:
nice!

PR #126, opened 2019-03-29
Revert "Handle BigNumber conversions in JSON properly (without loss of precision) (#71)"

xtinec:
This reverts commit e3866129b2b3cbd6e099fe4ab66319c4fe6ae485.

🐛 Bug Fix


packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -116,17 +116,21 @@ export default class SuperChart extends React.PureComponent<SuperChartProps, {}>
...
-    preTransformProps?: PreTransformProps;
-    transformProps?: TransformProps;
-    postTransformProps?: PostTransformProps;
-  }) => any;
-
-  createLoadableRenderer: (input: {
-    chartType: string;
-    overrideTransformProps?: TransformProps;
-  }) => LoadableRenderer<RenderProps, LoadedModules> | (() => null);
+  processChartProps: (
```
xtinec: @kristw this is odd -- my linter locally would convert it to this formatting when running `git commit` but CI doesn't like this new format. 🤔 Do you guys have this issue this locally?

kristw: Umm, I don't have this issue.

williaster: I did encounter this issue a couple times, was inconsistent tho 🤔

types/external.d.ts
```diff
@@ -1,30 +1 @@
...
-    ): string;
-  }
-
-  /**
-   * An intrinsic object that provides functions to convert JavaScript values to and from the JavaScript Object Notation (JSON) format.
-   */
-  const JSONbig: JSONbig;
-  export = JSONbig;
-}
+declare module 'fetch-mock';
```
williaster: I think you don't want to add this back since we now have the types as a dep. not sure if that's affecting the TS build.

packages/superset-ui-chart/test/components/ChartDataProvider.test.tsx
```diff
@@ -9,8 +9,11 @@ const defaultMockLoadFormData = jest.fn(allProps => Promise.resolve(allProps.for
 
 // coerce here else get: Type 'Mock<Promise<any>, []>' is not assignable to type 'Mock<Promise<any>, any[]>'
 let mockLoadFormData = defaultMockLoadFormData as jest.Mock<Promise<any>, any>;
-const mockLoadDatasource = jest.fn(datasource => Promise.resolve(datasource));
-const mockLoadQueryData = jest.fn(input => Promise.resolve(input));
+const mockLoadDatasource = jest.fn(datasource => Promise.resolve(datasource)) as jest.Mock<
```
xtinec: These fix the type errors seen in previous builds and on `master`.

xtinec:
@kristw @williaster No worries. Happy to fix them. Thanks for the quick turn-around on these! ❤️

PR #131, opened 2019-04-07
feat: callback hooks for Reactified components

xtinec:
🏆 Enhancements

Add a second argument to the exported `reactify` method to allow wrapped component to pass in callback hooks for React lifecycle methods. This PR only exposes the willUnmount hook. Other hooks can be added in the future.

### DIFF
--- a/packages/superset-ui-chart/src/components/reactify.tsx
+++ b/packages/superset-ui-chart/src/components/reactify.tsx
@@ -17,6 +17,11 @@ export type ReactifyProps = {
   className?: string;
 };
 
+// TODO: add more React lifecycle callbacks as needed
+export type LifeCycleCallbacks = {
+  willUnmount?: () => void;
+};
+
 export interface RenderFuncType<Props> {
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart/src/components/reactify.tsx
```diff
@@ -17,6 +17,11 @@ export type ReactifyProps = {
   className?: string;
 };
 
+// TODO: add more React lifecycle callbacks as needed
+export type LifeCycleCallbacks = {
+  willUnmount?: () => void;
```
williaster: small point -- but I think that these should mirror the actual lifecycle names because
1) devs are already familiar with them
2) some of them start with `componentXXX` and some don't (e.g., `shouldComponentUpdate`). The one's that don't are a little more cryptic when truncated (e.g. `componentUpdate`?), so if/when we add more we'll then have a mix of full names and short names.

xtinec: Good point! Let me rename this to `componentWillUnmount` and I'll adjust the usage in `nvd3` charts in `superset-ui-plugins`.

PR #132, opened 2019-04-08
fix: smartDateVerbose at seconds

datability-io:
🐛 Bug Fix

### DIFF
--- a/packages/superset-ui-time-format/package.json
+++ b/packages/superset-ui-time-format/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/time-format",
-  "version": "0.11.1",
+  "version": "0.11.0",
   "description": "Superset UI time-format",
   "sideEffects": false,
   "main": "lib/index.js",


packages/superset-ui-time-format/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/time-format",
-  "version": "0.11.0",
+  "version": "0.11.1",
```
kristw: The version number is managed automatically by `lerna`. Please do not modify the version number manually.

datability-io: @kristw I'm reverting this line.  Do I need to do anything to increment the version number at all?  Thanks.

xtinec: I can `yarn release` after the change is merged to up the version number, Thomas.

datability-io: @xtinec sounds good

PR #133, opened 2019-04-09
build: migrate to storybook 5

kristw:
🏠 Internal

* Migrate `superset-ui-demo` to Storybook 5


package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "beemo create-config prettier && beemo eslint \"./packages/*/{src,test,storybook}/**/*.{js,jsx,ts,tsx}\"",
     "lint:fix": "beemo create-config prettier && beemo eslint --fix \"./packages/*/{src,test,storybook}/**/*.{js,jsx,ts,tsx}\"",
     "jest": "beemo jest --color --coverage --react",
-    "postrelease": "lerna run gh-pages",
+    "postrelease": "lerna run deploy-demo",
```
williaster: 🎉

packages/superset-ui-demo/package.json
```diff
@@ -29,12 +29,14 @@
   },
   "homepage": "https://github.com/apache-superset/superset-ui#readme",
   "dependencies": {
-    "@babel/core": "^7.1.2",
-    "@babel/polyfill": "^7.2.5",
-    "@storybook/addon-knobs": "^4.0.2",
-    "@storybook/addon-options": "^4.0.3",
-    "@storybook/react": "^4.0.2",
-    "@superset-ui/chart": "^0.11.1",
+    "bootstrap": "^4.3.1",
```
williaster: `bootstrap` is wrong order?

package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "beemo create-config prettier && beemo eslint \"./packages/*/{src,test,storybook}/**/*.{js,jsx,ts,tsx}\"",
```
williaster: I think you've just moved storybook config, but since you've moved directories around just want to verify that this line still works for components/stories?

kristw: This should work. I only move `storybook-config` dir to `.storybook`. 
The `stories` are in the same path as before.
Only the file `storybook/storybookInfo.js` is merged into `config.js`.

PR #137, opened 2019-04-17
Feat: improve caching

betodealmeida:
🏆 Enhancements

This PR changes `SupersetClient` so that it can reuse cached responses using the [Cache API](https://developer.mozilla.org/en-US/docs/Web/API/Cache), in addition to delegating it to the browser. This allows us to pass `cache: 'no-cache'`, bypassing the browser caching and handling it ourselves.

The reason for this is that using the native browser cache we're unable to invalidate the cache from Javascript, eg, when a user saves a new chart of force refreshes a dashboard. Using the Cache API we can easily do that, clearing all cached responses that reference a chart (see https://github.com/apache/incubator-superset/pull/7319).

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,5 +109,8 @@
       "yarn run prettier --write",
       "git add"
     ]
+  },
+  "dependencies": {
+    "@types/sinon": "^7.0.11"
   }
 }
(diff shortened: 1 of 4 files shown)


betodealmeida:
@kristw @xtinec I'm getting unrelated [build errors](https://travis-ci.com/apache-superset/superset-ui/builds/108992986) in this PR, any idea what might be going on?

ktmud:
@betodealmeida I've made some change to `callApi` and there are more changes coming in #666 . Do we still want to keep changes in this PR? How much does the caching actually helps the end user? I suspect it only saves some (mostly intranet) network request as the DB query on the server side should already be cached?

PR #139, opened 2019-04-25
feat: add chart companion components

kristw:
🏆 Enhancements

* Create new package `@superset-ui/chart-companion` which includes the small things that are used to composite a chart. 
* Code are migrated from `@superset-ui/plugins` 's `preset-chart-xy` 
* Add unit tests for 100% coverage


packages/superset-ui-chart-companion/src/legend/WithLegend.tsx
```diff
@@ -0,0 +1,122 @@
...
+    const { className, width, height, position, renderChart, renderLegend } = this.props;
+
+    const isHorizontal = position === 'left' || position === 'right';
+
+    const style: CSSProperties = {
+      display: 'flex',
+      flexDirection: this.getContainerDirection(),
+    };
+
+    style.width = width;
```
williaster: why not just set these when you define `style`? `const style = { ..., width, height };`

packages/superset-ui-chart-companion/src/legend/WithLegend.tsx
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable sort-keys */
+import React, { CSSProperties, ReactNode, PureComponent } from 'react';
+import { ParentSize } from '@vx/responsive';
+// eslint-disable-next-line import/no-unresolved
+import * as CSS from 'csstype';
```
williaster: can you just import `FlexDirectionProperty`?

packages/superset-ui-chart-companion/test/legend/WithLegend.test.tsx
```diff
@@ -0,0 +1,161 @@
...
+      done();
+    }, 500);
+  });
+
+  it('renders without width or height', done => {
+    const wrapper = mount(<WithLegend renderChart={renderChart} renderLegend={renderLegend} />);
+
+    triggerResizeObserver();
+    // Have to delay > 300ms
+    // because the default debounce time for ParentSize is 300.
```
williaster: you could expose this prop so this is more declarative.

kristw:
`chart-composition` sounds good

packages/superset-ui-chart-companion/test/legend/WithLegend.test.tsx
kristw: great comment!

packages/superset-ui-chart-companion/src/legend/WithLegend.tsx
kristw: ah true. delete the `if` condition that is no longer needed and forgot to look around.
will fix.

PR #164, opened 2019-05-22
feat: update ChartFormData and QueryObject to support filters.

kristw:
💔 Breaking Changes

Rename types to make it clearer. The metric types in `ChartFormData` and `QueryObject` are different. 

* `FormDataMetric` => `ChartFormDataMetric`
* `Metrics` => `ChartFormDataMetrics`
* `Metric` => `QueryObjectMetric`

This is pretty mild breaking change as these are internally used types so I would incline to not bump major version.

🏆 Enhancements

* Update type of `ChartFormData` and `QueryObject` to support filters.
* Update `buildQueryObject` to convert filters from `ChartFormData` and put into corresponding fields in `QueryObject`. See `processFilters.ts` and `convertFilter.ts`.
* Improve typings of the `QueryObject` fields.
* Add type guards

📜 Documentation

* Add comments for fields on `QueryObject`

🏠 Internal

* Remove `Metrics` class and replace with two functions `processMetrics` and `convertMetric`.
* Migrate old unit tests and add new unit tests to ensure coverage.


packages/superset-ui-chart/src/query/convertFilter.ts
```diff
@@ -0,0 +1,30 @@
+import { SimpleAdhocFilter, isBinaryAdhocFilter, isUnaryAdhocFilter } from '../types/Filter';
+import { QueryObjectFilterClause } from '../types/Query';
+
+export default function convertFilter(filter: SimpleAdhocFilter): QueryObjectFilterClause {
+  const { subject } = filter;
+  if (isUnaryAdhocFilter(filter)) {
+    const { operator } = filter;
```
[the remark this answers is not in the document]
kristw: Has to be inside the block to infer the correct `Operator` type

packages/superset-ui-chart/src/types/Operator.ts
```diff
@@ -0,0 +1,43 @@
+/** List of operators that do not require another operand */
+export const UNARY_OPERATORS = ['IS NOT NULL', 'IS NULL'] as const;
```
conglei: do we need to export `UNARY_OPERATORS` etc? seems there is no other place to use them.

packages/superset-ui-chart/src/query/processFilters.ts
```diff
@@ -0,0 +1,51 @@
...
+import { QueryObjectFilterClause } from '../types/Query';
+import { isSimpleAdhocFilter } from '../types/Filter';
+import convertFilter from './convertFilter';
+
+/** Logic formerly in viz.py's process_query_filters */
+export default function processFilters(formData: ChartFormData) {
+  // TODO: Implement
+  // utils.convert_legacy_filters_into_adhoc(self.form_data)
+
+  // TODO: Implement
```
conglei: Will these two in the scope of this PR?

packages/superset-ui-chart/src/types/Operator.ts
kristw: Not being used at the moment. I will remove `export`

packages/superset-ui-chart/src/query/processFilters.ts
kristw: These will be beyond the scope. Right now this PR will handle `adhoc_filters` correctly plus provide typings.

packages/superset-ui-chart/src/query/buildQueryObject.ts
```diff
@@ -14,38 +18,40 @@ function getGranularity(formData: ChartFormData): string {
...
+    extras: processExtras(formData),
+    granularity: processGranularity(formData),
     groupby: Array.from(groupbySet),
     is_prequery: false,
-    is_timeseries: isTimeseries,
-    metrics: new Metrics(formData).getMetrics(),
-    order_desc: orderDesc,
+    is_timeseries: groupbySet.has(DTTM_ALIAS),
+    metrics: processMetrics(formData),
+    order_desc: order_desc === undefined ? true : order_desc,
```
williaster: `typeof === 'undefined'`?

packages/superset-ui-chart/src/types/Operator.ts
```diff
@@ -0,0 +1,43 @@
+/** List of operators that do not require another operand */
```
williaster: wow these are great!

packages/superset-ui-chart/src/query/processFilters.ts
williaster: do you think we should `console.warn` if we see legacy filters? (fine if not)

packages/superset-ui-chart/src/query/processFilters.ts
```diff
@@ -0,0 +1,51 @@
...
+  // merge_extra_filters(self.form_data)
+
+  // utils.split_adhoc_filters_into_base_filters(self.form_data)
```
williaster: is this just indicating what the logic below mapped to? (i.e., it's not part of the TODOs above) should we remove it or keep for context?

packages/superset-ui-chart/src/query/processFilters.ts
```diff
@@ -0,0 +1,51 @@
...
+  // eslint-disable-next-line camelcase
+  const { adhoc_filters } = formData;
+  if (Array.isArray(adhoc_filters)) {
+    const simpleWhere: QueryObjectFilterClause[] = [];
```
williaster: (not blocking) the type helps but I find these a little confusing name-wise, not sure what to expect `simple` to mean, maybe `whereFilter` <> `whereFreeform` and `havingFilter` <> `havingFreeform`?

packages/superset-ui-chart/src/query/processFilters.ts
kristw: The legacy filters are using fields `filters` and `having_filters` which do not exist on type `ChartFormData`, so typescript would complain.

I am wondering if we want to support this `utils.convert_legacy_filters_into_adhoc(self.form_data)` going forward. Not sure how long since we have moved from the legacy filters.

packages/superset-ui-chart/src/types/Metric.ts
```diff
@@ -15,41 +15,20 @@ export enum MetricKey {
...
-  MAX = 'MAX',
-  MIN = 'MIN',
-  SUM = 'SUM',
-}
-
-export enum ExpressionType {
-  SIMPLE = 'SIMPLE',
-  SQL = 'SQL',
-}
+export type Aggregate = 'AVG' | 'COUNT' | 'COUNT_DISTINCT' | 'MAX' | 'MIN' | 'SUM';
```
williaster: nice 👌

packages/superset-ui-chart/src/query/processFilters.ts
kristw: Will remove or change to simple sentence. Seem not necessary any more

PR #168, opened 2019-06-11
feat: allow chart plugin to be unregistered

kristw:
🏆 Enhancements

* Add `.unregister()` function to `ChartPlugin`

🏠 Internal

* Improve unit tests


packages/superset-ui-chart/src/models/ChartPlugin.ts
```diff
@@ -17,17 +17,17 @@ type ValueOrModuleWithValue<T> = T | { default: T };
 
 interface ChartPluginConfig<T extends ChartFormData> {
   metadata: ChartMetadata;
-  // use buildQuery for immediate value
+  /** Use buildQuery for immediate value. For lazy-loading, use loadBuildQuery. */
```
kristw: Change comments to `/** */` so they will show up in autocomplete

PR #169, opened 2019-06-11
refactor: separate transform function types into a new file

kristw:
🏠 Internal

* Move the types of transform functions from `types/Query.ts` to `types/TransformFunction.ts`
* Extract basic types `HandlerFunction` and `PlainObject` to `types/Base.ts`

### DIFF
--- a/packages/superset-ui-chart/src/types/Base.ts
+++ b/packages/superset-ui-chart/src/types/Base.ts
@@ -1,0 +1,5 @@
+export type HandlerFunction = (...args: any[]) => void;
+
+export interface PlainObject {
+  [key: string]: any;
+}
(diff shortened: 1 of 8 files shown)


packages/superset-ui-chart/src/types/Base.ts
```diff
@@ -0,0 +1,5 @@
+export type HandlerFunction = (...args: any[]) => void;
```
williaster: these could also just live in the `index` if they are generic/general? regardless, I don't think they're exported from the `index` now/not sure if you want them to be?

kristw: They are internal to this package now. I am trying not to export too many types from index if not necessary.

PR #171, opened 2019-06-12
feat: add parseLength function to @superset-ui/dimension

kristw:
🏆 Enhancements

Add function `parseLength` to parse width or height 
* `'100%'` => `{ isDynamic: true, multiplier: 1 }`
* `'50%'` => `{ isDynamic: true, multiplier: 0.5 }`
* `100` => `{ isDynamic: false, value: 100 }`
* `'100'` => `{ isDynamic: false, value: 100 }`
* `100px` => `{ isDynamic: false, value: 100 }`

### DIFF
--- a/packages/superset-ui-dimension/src/parseLength.ts
+++ b/packages/superset-ui-dimension/src/parseLength.ts
@@ -1,7 +1,5 @@
 const HUNDRED_PERCENT = { isDynamic: true, multiplier: 1 } as const;
 
-const pxPattern = /px$/;
-
 export default function parseLength(
   input: string | number,
 ): { isDynamic: true; multiplier: number } | { isDynamic: false; value: number } {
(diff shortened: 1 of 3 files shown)


packages/superset-ui-dimension/src/parseLength.ts
```diff
@@ -0,0 +1,17 @@
...
+const pxPattern = /px$/;
+
+export default function parseLength(
+  input: string | number,
+): { isDynamic: true; multiplier: number } | { isDynamic: false; value: number } {
+  if (input === 'auto' || input === '100%') {
+    return HUNDRED_PERCENT;
+  } else if (typeof input === 'string' && input.length > 0 && input[input.length - 1] === '%') {
+    // eslint-disable-next-line no-magic-numbers
+    return { isDynamic: true, multiplier: parseFloat(input.substring(0, input.length - 1)) / 100 };
```
khtruong: you could do parseFloat(input). If the string starts with a number, parseFloat will just ignore the random characters. Comment applies below too.

kristw: Oh cool. Didn't know that!

PR #172, opened 2019-06-12
test: extract mock plugins into separate file

kristw:
🏠 Internal

* extract mock plugins into separate file


packages/superset-ui-chart/test/components/MockChartPlugins.tsx
```diff
@@ -0,0 +1,69 @@
...
+  DILIGENT: 'diligent-chart',
+  LAZY: 'lazy-chart',
+  SLOW: 'slow-chart',
+};
+
+export class DiligentChartPlugin extends ChartPlugin<ChartFormData> {
+  constructor() {
+    super({
+      metadata: new ChartMetadata({
+        name: 'diligent-chart',
```
khtruong: nit: Should we use the consts you defined above? Comment applies to the rest of the plugins.

PR #173, opened 2019-06-12
feat: reuse svg element in getTextDimension

betodealmeida:
🏆 Enhancements

This PR optimizes the rendering of a `FilterableTable` component in Superset (see https://github.com/apache/incubator-superset/pull/7690) by allowing an SVG element to be reused when calling `getTextDimension`.

### DIFF
--- a/packages/superset-ui-dimension/src/createSVGNode.ts
+++ b/packages/superset-ui-dimension/src/createSVGNode.ts
@@ -1,5 +1,34 @@
-export default function createSVGNode(container: HTMLElement): SVGSVGElement {
+import { TextStyle } from './types';
+
+export interface CreateSVGNodeInput {
+  className?: string;
+  container?: HTMLElement;
+  style?: TextStyle;
+}
+
+const SVG_NS = 'http://www.w3.org/2000/svg';
+const STYLE_FIELDS: (keyof TextStyle)[] = [
+  'font',
+  'fontWeight',
+  'fontStyle',
+  'fontSize',
+  'fontFamily',
+  'letterSpacing',
+];
+
+export default function createSVGNode(input: CreateSVGNodeInput): SVGSVGElement {
+  const { className, style = {}, container = document.body } = input;
+
   const textNode = document.createElementNS(SVG_NS, 'text');
+  if (className !== undefined && className !== null) {
+    textNode.setAttribute('class', className);
+  }
+  STYLE_FIELDS.filter(
+    (field: keyof TextStyle) => style[field] !== undefined && style[field] !== null,
+  ).forEach((field: keyof TextStyle) => {
+    textNode.style[field] = `${style[field]}`;
+  });
+
   const svg = document.createElementNS(SVG_NS, 'svg');
   svg.style.position = 'absolute'; // so it won't disrupt page layout
   svg.style.opacity = '0'; // and not visible
--- a/packages/superset-ui-dimension/src/createSVGNode.ts
+++ b/packages/superset-ui-dimension/src/createSVGNode.ts
@@ -1,0 +1,11 @@
+export default function createSVGNode(container: HTMLElement): SVGSVGElement {
+  const textNode = document.createElementNS(SVG_NS, 'text');
+  const svg = document.createElementNS(SVG_NS, 'svg');
+  svg.style.position = 'absolute'; // so it won't disrupt page layout
+  svg.style.opacity = '0'; // and not visible
+  svg.style.pointerEvents = 'none'; // and not capturing mouse events
+  svg.appendChild(textNode);
+  container.appendChild(svg);
+
+  return svg;
+}
--- a/packages/superset-ui-dimension/src/getTextDimension.ts
+++ b/packages/superset-ui-dimension/src/getTextDimension.ts
@@ -19,13 +19,27 @@ export interface GetTextDimensionInput {
 
 const DEFAULT_DIMENSION = { height: 20, width: 100 };
 
+export function createSVGNode(container: HTMLElement): SVGSVGElement {
+  const textNode = document.createElementNS(SVG_NS, 'text');
+  const svg = document.createElementNS(SVG_NS, 'svg');
+  svg.style.position = 'absolute'; // so it won't disrupt page layout
+  svg.style.opacity = '0'; // and not visible
+  svg.style.pointerEvents = 'none'; // and not capturing mouse events
+  svg.appendChild(textNode);
+  container.appendChild(svg);
+
+  return svg;
+}
+
 export default function getTextDimension(
   input: GetTextDimensionInput,
   defaultDimension: Dimension = DEFAULT_DIMENSION,
+  existingSVGNode?: SVGSVGElement,
 ): Dimension {
   const { text, className, style = {}, container = document.body } = input;
 
-  const textNode = document.createElementNS(SVG_NS, 'text');
+  const svg = existingSVGNode || createSVGNode(container);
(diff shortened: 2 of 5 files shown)


packages/superset-ui-dimension/src/getTextDimension.ts
```diff
@@ -19,13 +19,27 @@ export interface GetTextDimensionInput {
 
 const DEFAULT_DIMENSION = { height: 20, width: 100 };
 
+export function createSVGNode(container: HTMLElement): SVGSVGElement {
```
kristw: Please separate this into another file.

packages/superset-ui-dimension/src/getTextDimension.ts
```diff
@@ -19,13 +19,27 @@ export interface GetTextDimensionInput {
...
 export default function getTextDimension(
   input: GetTextDimensionInput,
   defaultDimension: Dimension = DEFAULT_DIMENSION,
+  existingSVGNode?: SVGSVGElement,
 ): Dimension {
   const { text, className, style = {}, container = document.body } = input;
 
-  const textNode = document.createElementNS(SVG_NS, 'text');
+  const svg = existingSVGNode || createSVGNode(container);
+  const textNode = svg.lastElementChild as SVGSVGElement;
```
kristw: You will have to remove `className` and remove all styles from this text node before continuing. I am not sure how much performance gain do you get vs. creating a new text node. The `svg` node seems easily reusable.

packages/superset-ui-dimension/src/getTextDimension.ts
betodealmeida: +1

packages/superset-ui-dimension/src/getTextDimension.ts
betodealmeida: I'll give it a try with creating a new one.

By reusing the `svg` node, the time spent in `getTextDimension` went down from 20 seconds to less than a second for a very large table. I'll benchmark to get the exact numbers.

kristw: You could also try to hash `className` and `textStyle` and store hash in the text element. If the hash is the same, then no need to recreate `<text>`.

etr2460:
@betodealmeida are we planning to move forward with this fix? It would be good to either get this or https://github.com/apache/incubator-superset/pull/7690 in as it significantly improves the user experience in sql lab

betodealmeida:
@etr2460 yeah, I want to run some numbers for @kristw about performance improvements. I'll do that today, sorry!

betodealmeida:
I did some benchmarks today, loading a wide table:

1. Without reusing the `svg` node: 4315ms
2. Reusing the `svg` node: 2167ms
3. Reusing the `svg` node and the `textNode` (this PR): 481ms

packages/superset-ui-dimension/src/getTextDimension.ts
```diff
@@ -23,31 +15,17 @@ export default function getTextDimension(
   input: GetTextDimensionInput,
```
etr2460: It would be nice to add some documentation here about how to use getTextDimension. From what I can see, you have two options:

1. Don't provide an svg node, but provide the className and style details required to construct it.
2. Provide an svg node that you've already applied the className and style to; these parameters now have no effect

I don't think it's clear to the developer that providing an svg node and a className/style doesn't apply the className/style to the node. Either a nice docblock or splitting this into 2 different functions with 2 different inputs would make it more clear

packages/superset-ui-dimension/src/createSVGNode.ts
```diff
@@ -0,0 +1,40 @@
...
+];
+
+export default function createSVGNode(input: CreateSVGNodeInput): SVGSVGElement {
+  const { className, style = {}, container = document.body } = input;
+
+  const textNode = document.createElementNS(SVG_NS, 'text');
+  if (className !== undefined && className !== null) {
+    textNode.setAttribute('class', className);
+  }
+  STYLE_FIELDS.filter(
```
kristw: You will have to remove className and remove all styles from this text node before continuing. otherwise style will keep accumulating

betodealmeida: We're only doing this once now, though. Not on every call like the first version of the PR.

etr2460:
if you rebase, then travis should pass and this will be good to merge

packages/superset-ui-dimension/src/createSVGNode.ts
kristw: My point is there is no guarantee that all text in consecutive calls will have the exact same style. and by not clearing them, the later ones will have accumulated style.

betodealmeida: Sorry, @kristw, I still don't understand.

There's 2 ways of calling `getTextDimension` with this PR:

In the old one, you call it basically with a string and possible a style or class name. In that case:

1. An SVG node is created and added to the passed container.
2. A text node is created and appended to the SVG node.
3. The class is applied to the text node.
4. Any whitelisted (ie, in `STYLE_FIELDS`) style field is applied to the text node.
5. The text node content is set to the string.
6. Its bounding box is computed.
7. The SVG node is removed from the container.

I think we both agree that in this case there's no way to accumulate styles, since the node is created and removed for every call.

In the new one, you call `createSVGNode` yourself, and pass it to `getTextDimension` with a string. When calling `createSVGNode`:

1. An SVG node is created and added to the passed container.
2. A text node is created and appended to the SVG node.
3. The class is applied to the text node.
4. Any whitelisted (ie, in `STYLE_FIELDS`) style field is applied to the text node.
5. The SVG node is returned.

Now you can pass this node together with the string to `getTextDimension`. Note that when the existing node is passed, `className`, `style` and `container` are ignored and have no effect. So now, each call to `getTextDimension` will simply:

1. Set the content of the text node.
2. Compute its bounding box.

When the existing SVG node is reused *we're not* setting the styles nor the class name again, we simply replace the text content and compute the bounding box.

betodealmeida:
@kristw did you see my comments here: https://github.com/apache-superset/superset-ui/pull/173#discussion_r298807433 ?

etr2460:
@betodealmeida I was able to talk briefly with Krist and clarify his concerns. The problem he had was that this new API, while technically correct, is very specific to your use case. The API works great when measuring the size of 1 piece of text, or when measuring the size of many pieces of text with the exact same style, but there's no good way to efficiently measure many sets of text with different styling.

I think since we're creating an optimization for a specific use case, we should separate this into two functions: a general purpose `getTextDimension` that works as before, and a `getManyTextDimensions` (or a name like that) for the use case of measuring many strings with the same styling. That second function could take in an array of strings and return an array of widths. I think that would be the cleanest API here, but I'd be open to a better solution too.

ccing @williaster in case he has other ideas too. Krist is still out on leave, but I'd definitely like to get this merged in so we can take advantage of the perf improvement (while still maintaining a clean API!)

williaster:
yeah I think not over-engineering this is a good idea, when there are lots of ways to use something it becomes more confusing. I agree that @etr2460 's proposal of 2 functions would be a decent way to separate the 2 uses 👍 

> - general purpose `getTextDimension` that works as before, and 
> - `getManyTextDimensions` (or similar) for the use case of measuring many strings with the same styling

betodealmeida:
@etr2460 @williaster thanks, I'll work on the PR to split into 2 functions.

kristw:
Replaced by #199

PR #178, opened 2019-06-18
feat: split superset-ui/query from superset-ui/chart

### LINKED ISSUE #176: Split @superset-ui/query from @superset-ui/chart
**Is your feature request related to a problem? Please describe.**

Logic for creating `QueryContext` and `QueryObject` are in the `@superset-ui/chart` package. 

**Describe the solution you'd like**

Separate them into another package for simplicity.

**Describe alternatives you've considered**

Don't do anything.

kristw:
### 💔 Breaking Changes

**A. Some functions and types are removed from `@superset-ui/chart` and moved to new package `@superset-ui/query`**
* `buildQueryContext`
* `buildQueryObject`
* types related to `QueryObject`

*There is no change to the functionalities.*

Fix #176 

**B. `ChartFormData` is renamed to `QueryFormData`**

This is to better reflect what the object actually is. Historically, `formData` in Superset stores all information necessary to reproduce a chart, which include
1. database query information 
2. visual appearance

We have always considered `formData` to be a single type that contains both information in the context of Superset app (`incubator-superset`). However, after the development of embeddable charts, we see the following use cases: 
1. Query data from Superset without using its chart. => only need the `query` part of `formData`.
2. Display Superset chart without query (User supplies data in their own way.) => only need the visual appearance part of `formData`.

It seems more logical to separate the two parts of `formData` into its own type.
1. `QueryFormData`, which is what `@superset-ui/query` handles. `buildQueryObject` converts `QueryFormData` to `QueryObject`.
2. `ChartFormData`, which varies chart by chart and can be defined in the `@superset-ui/plugins` packages. 

```ts
type SupersetFormData = QueryFormData & ChartFormData
```

*Currently, `ChartFormData` is referenced in `plugin-chart-table` which is still under development, and `dataportal`*

### Migration guide

* Change broken reference from `@superset-ui/chart` to `@superset-ui/query`
* For apps that has `@superset-ui/chart` as dependency, also add `@superset-ui/query`
* Rename `ChartFormData` to `QueryFormData`.


packages/superset-ui-chart/src/clients/ChartClient.ts
```diff
@@ -6,19 +6,17 @@ import {
   Json,
   SupersetClientClass,
 } from '@superset-ui/connection';
+import { AnnotationLayerMetadata, QueryFormData, Datasource } from '@superset-ui/query';
 import getChartBuildQueryRegistry from '../registries/ChartBuildQueryRegistrySingleton';
 import getChartMetadataRegistry from '../registries/ChartMetadataRegistrySingleton';
-import { AnnotationLayerMetadata } from '../types/Annotation';
```
etr2460: Judging from the name, `AnnotationLayerMetadata` seems more related to charts than queries. why is it being pulled out here?

kristw:
@etr2460 I have put the `AnnotationLayerMetadata` back

PR #201, opened 2019-08-13
feat: Add types and type guards for encodeable

kristw:
🏆 Enhancements

* Add new package `@superset-ui/encodeable` 
* Add basic types and type guards with new unit tests for type guards. (Migrated from `@superset-ui-plugins/preset-chart-xy`. This is a subset original work from https://github.com/apache-superset/superset-ui-plugins/pull/15 + revision + unit tests

> ### Define encoding config for a visualization
> * Each visualization has one or more *channels* that can be encoded, such as `x`, `y`, `color`, `stroke`, fill. These channels configurations should be consistent/sharable across all visualizations. For example, `x` and `y` usually have scales and axes. Scales have `domain`, `range`. Axes have `label`, `format`. Color can support categorical, sequential or fixed color. All channels can support fixed value, field name in the data, or sometimes aggregates (haven't implemented the aggregate yet).
> * Each encoding channel config is based on `vega-lite` encoding grammar. ([See line chart example](https://github.com/apache-superset/superset-ui-plugins/compare/kristw--line?expand=1#diff-89f0892b8e6df5ff19a42243e2254d6a), [vega-lite encoding documentation](https://vega.github.io/vega-lite/docs/encoding.html))
> * Only adopt a small subset of `vega-lite` encoding grammar for now as we don’t have enough time (and may not intend) to provide 100% feature of `vega-lite`. 

🏠 Internal

* The package is currently marked as `private` until completion.


packages/superset-ui-encodeable/src/index.ts
```diff
@@ -0,0 +1,2 @@
+const x = 1;
```
williaster: is this intentional?

packages/superset-ui-encodeable/src/types/ChannelDef.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export type ChannelDef<Output extends Value = Value> =
+  | ValueDef<Output>
+  | XFieldDef<Output>
+  | YFieldDef<Output>
+  | MarkPropFieldDef<Output>
+  | TextFieldDef;
+
+/** Channel definitions that are not constant value */
+export type NonValueDef<Output extends Value = Value> = Exclude<
```
williaster: `NonConstantValueDef` for more clear naming?

packages/superset-ui-encodeable/src/types/Inherit.ts
```diff
@@ -0,0 +1,6 @@
+// Types directly imported from vega-lite
+
+export { ValueDef, Value } from 'vega-lite/build/src/channeldef';
```
williaster: `types/Vegalite.ts` is more clear to me when seeing these imports elsewhere

packages/superset-ui-encodeable/test/index.test.ts
```diff
@@ -0,0 +1,5 @@
+describe('My Test', () => {
```
williaster: also not sure if you still need this one?

kristw: removed.

packages/superset-ui-encodeable/src/index.ts
kristw: It was templated. Now removed.

packages/superset-ui-encodeable/src/types/ChannelDef.ts
kristw: There is `ValueDef` and this one is the opposite of it.

PR #202, opened 2019-08-13
feat: deprecate old SuperChart API that accepts chartProps

kristw:
💔 Breaking Changes

* No longer accept `chartProps` as a single prop in `<SuperChart>`
* Must specify each field in `chartProps` individually. e.g.

```
<SuperChart width={} height={} formData={} />
```

instead of 

```
<SuperChart chartProps={{width: ... , height: ...,  formData: ...}} />
```


kristw:
Please ignore `codecov` warning. The reduced percentage is due to code deletion.

PR #209, opened 2019-08-24
feat: add duration formatter

villebro:
🏆 Enhancements
- Add number formatter for time durations based on `pretty-ms` (MIT license). Example: 1000 ms = "1s", 90000 ms = "1m 30s". Supports both ms (`DURATION_MS`) and seconds (`DURATION_S`) as input values.
- Add unit tests.


villebro:
For context: We needed a queue time chart on a dashboard, and all existing number formatting options were sub-optimal. For example, 1 min 30 seconds would show display as either "90" (seconds) or "1.5" (minutes). This uses [pretty-ms](https://www.npmjs.com/package/pretty-ms) to format either milliseconds or seconds into more human readable format:
- 500 milliseconds = "500ms"
- 90 seconds = "1m 30s"

Not sure if community PRs are welcomed yet, but thought this might be useful to others, too, so decided to open up a PR.

packages/superset-ui-number-format/src/factories/createDurationFormatter.ts
```diff
@@ -0,0 +1,21 @@
+import prettyMsFormatter from 'pretty-ms';
+import NumberFormatter from '../NumberFormatter';
+import NumberFormats from '../NumberFormats';
+
+export default function createDurationFormatter(
+  config: {
```
kristw: Can also support options of `pretty-ms`

```ts
config: {
    description?: string;
    id?: string;
    label?: string;
    multiplier?: number;
  } & prettyMsFormatter.Options
```

then

```ts
const { description, id, label, multiplier = 1, ...prettyMsOptions } = config;

...
prettyMsFormatter(value * multiplier, prettyMsOptions)
...
```

packages/superset-ui-number-format/src/NumberFormats.ts
```diff
@@ -3,6 +3,10 @@ const DOLLAR_SIGNED = '+$,.2f';
 const DOLLAR_ROUND = '$,d';
 const DOLLAR_ROUND_SIGNED = '+$,d';
 
+const DURATION = 'DURATION';
```
kristw: Please revert the changes to this file. The duration formatter should be optional and not registered by default. There are other apps using this package and some do not need duration formatter. You can register these formatters in `incubator-superset`'s `setupFormatter.js`.  That way other apps can take advantage of tree-shaking and won't have to bundle `pretty-ms`.

packages/superset-ui-number-format/src/NumberFormatterRegistry.ts
```diff
@@ -14,6 +15,9 @@ export default class NumberFormatterRegistry extends RegistryWithDefaultKey<
       overwritePolicy: OverwritePolicy.WARN,
     });
 
+    this.registerValue(NumberFormats.DURATION, createDurationFormatter());
```
kristw: Please revert the changes to this file. The duration formatter should be optional and not registered by default. There are other apps using this package and some do not need duration formatter. You can register these formatters in `incubator-superset`'s `setupFormatter.js`.  That way other apps can take advantage of tree-shaking and won't have to bundle `pretty-ms` when they are not using.

kristw: You can add a section in `README` to instruct how to make duration formatter available.

villebro:
Thanks @kristw for the review and helpful comments. Before I go about making the proposed changes, do you think the name is good (`durationFormatter`), or should it perhaps be more explicitly related to `pretty-ms`, i.e. be called `prettyMsFormatter`? Especially after introducing `prettyMsOptions` this might seem more appropriate.

kristw:
@villebro I think `durationFormatter` is quite clear.

packages/superset-ui-number-format/src/NumberFormatterRegistry.ts
villebro: Ok that makes sense. I wasn't able to find `setupFormatter.js` or any reference to `getNumberFormatterRegistry()` in `incubator-superset`, but I think I have a good idea of how it should be done. So if it's ok I'll open a new PR to add instructions to `README.md` once `0.12` is released and `durationFormatter` gets implemented in the backend.

kristw: Sorry I made a typo. It is `setupFormatters.js`. For `README` I meant `packages/superset-ui-number-format/README.md`

PR #215, opened 2019-08-28
feat: add Wrapper support and bounding box for dynamic width/height

kristw:
### 1. Add bounding box for dynamic `width`/`height` calculation

When either `width` or `height` is already known, such as `width="100%" height={500}`.
Default `<ParentSize>` behavior will set the container style to `{width: '100%', height: '100%'}`, which may exceed the known `width` or `height` values, resulting in undesired blank space on the page. This PR add another `<div>` to wrap it (only when necessary). If both `width` and `height` are dynamic then no bounding box will be added. 

e.g. when want 100% width and 500px height.

```jsx
<div style={{width: '100%', height: 500}}>
  <ParentSize>
    {...}
  </ParentSize>
</div>
```

#### Alternative

Consider including these functionalities in the `@vx/responsive` rewrite.

#### Problem

Inspired by use case in `dataportal`. 

```jsx
// SuperChart with dynamic width/height
<SuperChart width="100%" height="400" chartType="line /> 
```
```jsx
// Rough hierarchy of the code above in the DOM is like this. 
<SuperChart width="100%" height="400"> // can compute dynamic width/height
  <ChartPlugin /> // for illustration purpose only, this is abstracted by SuperChart
</SuperChart>
```

There is a situation when a wrapper needs to be added to fix `z-index` issue such as tooltip.
This won't work because the wrapper will mess with the dynamic width/height calculation.

```jsx
<div style={{position: 'fixed'}}>
  // with the wrapper, now width and height become 0.
  // so the chart will disappear 😱
  <SuperChart width="100%" height="100%" chartType="line" /> 
</div>
```

If we want to insert something in between `SuperChart` and `ChartPlugin` to avoid messing with dynamic `width`/`height`, you either have to 

A. Do not use the built-in `SuperChart` way of determining auto `width` and `height`. Use your own way of computing dynamic `width` & `height`, which is pretty much re-implementing a lot of logic in `SuperChart`.

```jsx
<div style={{ width="100%", height: 400 }}>
  <ParentSize>
    // ...
    <div style={{position: 'fixed'}}>
      <SuperChart width={width} height={height} chartType="line" /> // can compute dynamic width/height
    </div>
    // ...
  </ParentSize>
</div>
```
 
B. Modify the `ChartPlugin` to add the wrapper. Sounds like an overkill.

Both options are undesirable extra work. 

#### Solution

With the changes in this PR.

```jsx
const Wrapper = ({children}) => (
  <div style={{position: 'fixed'}}>{children}</div>
);

return (
  <SuperChart Wrapper={Wrapper} width="100%" height="400" /> 
);
```

## 🏠 Internal

* Add unit test for one missing branch that makes the coverage ~99%. Now the `chart` package is 100% again.

### DIFF
--- a/packages/superset-ui-chart/src/components/SuperChart.tsx
+++ b/packages/superset-ui-chart/src/components/SuperChart.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 import ErrorBoundary, { ErrorBoundaryProps, FallbackProps } from 'react-error-boundary';
-import { parseLength } from '@superset-ui/dimension';
+import { parseLength, Dimension } from '@superset-ui/dimension';
 import { ParentSize } from '@vx/responsive';
+import { createSelector } from 'reselect';
 import SuperChartCore, { Props as SuperChartCoreProps } from './SuperChartCore';
 import DefaultFallbackComponent from './FallbackComponent';
 import ChartProps, { ChartPropsConfig } from '../models/ChartProps';
@@ -13,16 +14,40 @@ const defaultProps = {
   width: '100%' as string | number,
 };
 
-export type FallbackPropsWithDimension = FallbackProps & { width?: number; height?: number };
+export type FallbackPropsWithDimension = FallbackProps & Partial<Dimension>;
+
+export type WrapperProps = Dimension & {
+  children: ReactNode;
+};
 
 export type Props = Omit<SuperChartCoreProps, 'chartProps'> &
   Omit<ChartPropsConfig, 'width' | 'height'> & {
+    /**
+     * Set this to true to disable error boundary built-in in SuperChart
+     * and let the error propagate to upper level
+     * and handle by yourself
+     */
     disableErrorBoundary?: boolean;
+    /** debounceTime to check for container resize */
     debounceTime?: number;
+    /** Component to render when there are unexpected errors */
     FallbackComponent?: React.ComponentType<FallbackPropsWithDimension>;
+    /** Event listener for unexpected errors from chart */
     onErrorBoundary?: ErrorBoundaryProps['onError'];
+    /** Chart width */
     height?: number | string;
+    /** Chart height */
     width?: number | string;
+    /**
+     * Component to wrap the actual chart
+     * after the dynamic width and height are determined.
+     * This can be useful for handling tooltip z-index, etc.
+     * e.g. <div style={{ position: 'fixed' }} />
+     * You cannot just wrap this same component outside of SuperChart
+     * when using dynamic width or height
+     * because it will clash with auto-sizing.
+     */
+    Wrapper?: React.ComponentType<WrapperProps>;
   };
 
 type PropsWithDefault = Props & Readonly<typeof defaultProps>;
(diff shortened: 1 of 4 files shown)


packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -54,26 +100,29 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
...
-          height,
-          width,
-        })}
-        preTransformProps={preTransformProps}
-        overrideTransformProps={overrideTransformProps}
-        postTransformProps={postTransformProps}
-        onRenderSuccess={onRenderSuccess}
-        onRenderFailure={onRenderFailure}
-      />
+      <Wrapper width={width} height={height}>
```
williaster: `width` and `height` are invalid props for `React.Fragment` right? Ideally we don't get `propType` errors in that case 🤔

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -37,6 +62,27 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
...
+  private parseDimension = createSelector(
+    ({ width }: { width: string | number; height: string | number }) => width,
+    ({ height }) => height,
+    (width, height) => {
+      // Parse them in case they are % or 'auto'
+      const widthInfo = parseLength(width);
+      const heightInfo = parseLength(height);
+
+      const style = {
+        height: heightInfo.isDynamic ? '100%' : heightInfo.value,
```
williaster: could this be simplified so `parseLength` sets `value` to `100%` if `isDynamic` is true?

williaster: if not, below we use `multiplier`, why not use `multiplier` here instead of `100%` or is this slightly different?

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -92,27 +141,34 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
...
-          {({ width, height }) =>
-            width > 0 &&
-            height > 0 &&
-            this.renderChart(
-              widthInfo.isDynamic ? Math.floor(width * widthInfo.multiplier) : widthInfo.value,
-              heightInfo.isDynamic ? Math.floor(height * heightInfo.multiplier) : heightInfo.value,
-            )
-          }
-        </ParentSize>
+        // bounding box will ensure that when
```
williaster: could probably consolidate to fewer lines here? I also wonder if this could be moved to the `BoundingBox` selector to give more context to the `Fragment` vs `div` logic?

packages/superset-ui-chart/src/components/SuperChart.tsx
kristw: https://stackoverflow.com/questions/48705910/does-react-warn-of-additional-props-that-a-component-wasnt-expecting

packages/superset-ui-chart/src/components/SuperChart.tsx
kristw: Great idea.

packages/superset-ui-chart/src/components/SuperChart.tsx
kristw: Comments moved as suggested.

PR #230, opened 2019-10-02
feat: Add Encoder

kristw:
🏆 Enhancements

Add `Encoder` to `@superset-ui/encodable`. An `Encoder` is a utility class created once per chart type. 
* It consists of one or more `ChannelEncoder`
* It takes users' definitions of the channels (`vega-lite` syntax), then resolves ambiguity and produce complete definition and actionable utility functions. 

### Example Usage

The component developer creates an encoder factory once.

```ts
type LineChartEncodingConfig = {
  // channelName: [ChannelType, output type, support multiple fields or not]
  x: ['X', number];
  y: ['Y', number];
  color: ['Color', string];
  tooltip: ['Text', string, 'multiple'];
};

const lineChartEncoderFactory = createEncoderFactory<LineChartEncodingConfig>({
    // This part is manual work to convert the `ChannelType` defined in `LineChartEncodingConfig` above
    // from `type` to `value`. `LineChartEncodingConfig` is the source of truth 
    // and because it is generic for this function, type checking will ensure the only compatible value
    // for channelTypes is exactly like the one below.
    channelTypes: {
      x: 'X',
      y: 'Y',
      color: 'Color',
      tooltip: 'Text',
    },
    // Define default definition for each channel
    defaultEncoding: {
      x: { type: 'quantitative', field: 'x' },
      y: { type: 'quantitative', field: 'y' },
      color: { value: 'black' },
      tooltip: [],
    },
  });

type LineChartEncoding = DeriveEncoding<LineChartEncodingConfig>;
```

The `factory` encapsulates the awkward `channelTypes` and `defaultEncoding`
* which are constants across all `Encoder` instance of this chart
* making it convenient to create a new `Encoder` from `encoding` because `factory.create(encoding)` only needs one argument: `encoding`.

This is how user-specified `encoding` may look like.

```ts
const encoding: LineChartEncoding = {
  x: { type: 'quantitative', field: 'speed', scale: { domain: [0, 100] } },
  y: { type: 'quantitative', field: 'price' },
  color: { type: 'nominal', field: 'brand' },
  tooltip: [{ field: 'make' }, { field: 'model' }],
};
```

Then create an `Encoder` for the incoming `encoding` and use it to facilitate rendering. 

```ts
const encoder = lineChartEncoderFactory.create(encoding);
encoder.channels.x.getValueFromDatum({ speed: 100 }); // 100
encoder.channels.x.encodeDatum({ speed: 100 }); // 1
encoder.getGroupBys(); // ['brand', 'make', 'model']
```

### DIFF
--- a/packages/superset-ui-encodable/src/encoders/Encoder.ts
+++ b/packages/superset-ui-encodable/src/encoders/Encoder.ts
@@ -1,0 +1,163 @@
+import { flatMap } from 'lodash';
+import { Value } from '../types/VegaLite';
+import { ChannelType, ChannelInput } from '../types/Channel';
+import { ChannelDef } from '../types/ChannelDef';
+import { Dataset } from '../types/Data';
+import { Unarray, MayBeArray } from '../types/Base';
+import { isFieldDef, isTypedFieldDef } from '../typeGuards/ChannelDef';
+import { isArray, isNotArray } from '../typeGuards/Base';
+import ChannelEncoder from './ChannelEncoder';
+import { Encoding } from '../types/Encoding';
+
+type AllChannelEncoders<Encoding extends Record<string, MayBeArray<ChannelDef>>> = {
+  readonly [k in keyof Encoding]: Encoding[k] extends any[]
+    ? ChannelEncoder<Unarray<Encoding[k]>>[]
+    : ChannelEncoder<Unarray<Encoding[k]>>;
+};
+
+export default class Encoder<
+  ChannelTypes extends Record<string, ChannelType>,
+  CustomEncoding extends Encoding<keyof ChannelTypes>
+> {
+  readonly encoding: CustomEncoding;
+  readonly channelTypes: ChannelTypes;
+  readonly channels: AllChannelEncoders<CustomEncoding>;
+
+  readonly legends: {
+    [key: string]: (keyof CustomEncoding)[];
+  };
+
+  constructor({
+    channelTypes,
+    encoding,
+  }: {
+    channelTypes: ChannelTypes;
+    encoding: CustomEncoding;
+  }) {
+    this.channelTypes = channelTypes;
+    this.encoding = encoding;
+    const channelNames = this.getChannelNames();
+
+    const channels: { [k in keyof CustomEncoding]?: MayBeArray<ChannelEncoder<ChannelDef>> } = {};
+
+    channelNames.forEach(name => {
+      const channelEncoding = encoding[name];
+      if (isArray(channelEncoding)) {
+        const definitions = channelEncoding;
+        channels[name] = definitions.map(
+          (definition, i) =>
+            new ChannelEncoder({
+              channelType: 'Text',
+              definition,
+              name: `${name}[${i}]`,
+            }),
+        );
+      } else if (isNotArray(channelEncoding)) {
+        const definition = channelEncoding;
+        channels[name] = new ChannelEncoder({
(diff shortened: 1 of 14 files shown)


packages/superset-ui-encodable/src/encoders/createEncoderFactory.ts
```diff
@@ -0,0 +1,43 @@
+import Encoder from './Encoder';
+import { EncodingConfig, DeriveChannelTypes, DeriveEncoding } from '../types/Encoding';
+import mergeEncoding from '../utils/mergeEncoding';
+
+type CreateEncoderFactoryParams<Config extends EncodingConfig> = {
```
williaster: this `type` is beautiful 😻

packages/superset-ui-encodable/src/types/Encoding.ts
```diff
@@ -0,0 +1,27 @@
+import { ChannelType, ChannelTypeToDefMap } from './Channel';
+import { Value } from './VegaLite';
+import ChannelEncoder from '../encoders/ChannelEncoder';
+
+export type EncodingConfig = {
+  [k in string]: [ChannelType, Value, 'multiple'?];
+};
+
+export type DeriveChannelTypes<Config extends EncodingConfig> = {
+  readonly [k in keyof Config]: Config[k]['0'];
```
williaster: `#TIL`, did not know that you could index types like this. 

interesting it has to be a string and not a number.

packages/superset-ui-encodable/src/types/Encoding.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type DeriveEncoding<Config extends EncodingConfig> = {
+  [k in keyof Config]: Config[k]['2'] extends 'multiple'
+    ? ChannelTypeToDefMap<Config[k]['1']>[Config[k]['0']][]
+    : ChannelTypeToDefMap<Config[k]['1']>[Config[k]['0']];
+};
+
+export type DeriveChannelEncoders<Config extends EncodingConfig> = {
+  readonly [k in keyof Config]: Config[k]['2'] extends 'multiple'
+    ? ChannelEncoder<ChannelTypeToDefMap<Config[k]['1']>[Config[k]['0']]>[]
```
williaster: so advanced! 🤯

packages/superset-ui-encodable/src/encoders/Encoder.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    this.channels = channels as DeriveChannelEncoders<Config>;
+
+    // Group the channels that use the same field together
+    // so they can share the same legend.
+    this.legends = {};
+    channelNames
+      .map(name => this.channels[name])
+      .forEach(c => {
+        if (isNotArray(c) && c.hasLegend() && isFieldDef(c.definition)) {
```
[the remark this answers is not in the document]
kristw: At this point, the channels that are array are fields for `tooltip`. I tried to think if we want to support other channels that require legends such as `color` as array of multiple fields but that makes thing very complicated. Better create a derived field `field1,field2` in the dataset and use that derived field instead. (`vega` also has support for `formula` instead of `field` for this kind of purpose, which I haven't explored how to port)

packages/superset-ui-encodable/src/types/Encoding.ts
kristw: This is like 3rd-4th time I rewrote the whole thing lol.

PR #236, opened 2019-10-08
feat: add function for getting legend information

kristw:
🏆 Enhancements

Add `encoder.getLegendInformation(dataset)`, which generates metadata for rendering legend.

### Example usage

```ts
const legendInfo = factory
  .create({
    color: { 
      type: 'nominal', 
      field: 'brand', 
      scale: { range: ['red', 'green', 'blue'] } 
    },
    radius: { value: 5 },
    shape: { value: 'circle' },
  })
  .getLegendInformation([{ brand: 'Gucci' }, { brand: 'Prada' }]);

// [
//   // legend group information
//   {
//     field: 'brand',
//     type: 'nominal',
//     items: [
//       {
//         input: 'Gucci',
//         output: {
//           color: 'red',
//           radius: 5,
//           shape: 'circle',
//         },
//       },
//       {
//         input: 'Prada',
//         output: {
//           color: 'green',
//           radius: 5,
//           shape: 'circle',
//         },
//       },
//     ],
//   }
// ]
```

This also works with quantitative fields, with small difference. 

```ts
const legendInfo = factory
  .create({
    color: {
      type: 'quantitative',
      field: 'price',
      scale: { domain: [0, 20], range: ['#fff', '#f00'] },
    },
    radius: {
      type: 'quantitative',
      field: 'price',
      scale: { domain: [0, 20], range: [0, 10] },
    },
    shape: { value: 'circle' },
  })
  .getLegendInformation();
```

Unlike the nominal field, `items` is not included in the output, so you cannot use `legendInfo[0].items` directly. Have to create legend items manually from inputs. 

In the future we can implement logic to infer the appropriate inputs from continuous domain and output items the same way the nominal field does.

```ts
legendInfo[0].createLegendItems([0, 10, 20]);

// [
//   {
//     input: 0,
//     output: {
//       color: 'rgb(255, 255, 255)',
//       radius: 0,
//       shape: 'circle',
//     },
//   },
//   {
//     input: 10,
//     output: {
//       color: 'rgb(255, 128, 128)',
//       radius: 5,
//       shape: 'circle',
//     },
//   },
//   {
//     input: 20,
//     output: {
//       color: 'rgb(255, 0, 0)',
//       radius: 10,
//       shape: 'circle',
//     },
//   },
// ]
```

### DIFF
--- a/packages/superset-ui-encodable/src/encoders/Encoder.ts
+++ b/packages/superset-ui-encodable/src/encoders/Encoder.ts
@@ -1,7 +1,7 @@
 import { flatMap } from 'lodash';
 import { ChannelDef, TypedFieldDef } from '../types/ChannelDef';
 import { MayBeArray } from '../types/Base';
-import { isTypedFieldDef } from '../typeGuards/ChannelDef';
+import { isTypedFieldDef, isValueDef } from '../typeGuards/ChannelDef';
 import { isNotArray } from '../typeGuards/Base';
 import ChannelEncoder from './ChannelEncoder';
 import {
@@ -95,7 +95,27 @@ export default class Encoder<Config extends EncodingConfig> {
     return Array.from(new Set(fields));
   }
 
-  getLegendInformation(data: Dataset) {
+  private createLegendItemsFactory(field: string) {
+    const channelEncoders = flatMap(
+      this.getChannelEncoders().filter(e => isNotArray(e) && isValueDef(e.definition)),
+    ).concat(this.legends[field]);
+
+    return (domain: ChannelInput[]) =>
+      domain.map((input: ChannelInput) => ({
+        input,
+        output: channelEncoders.reduce(
+          (prev: Partial<{ [k in keyof Config]: Config[k]['1'] }>, curr) => {
+            const map = prev;
+            map[curr.name as keyof Config] = curr.encodeValue(`${input}`) as Value;
+
+            return map;
+          },
+          {},
+        ),
+      }));
+  }
+
+  getLegendInformation(data: Dataset = []) {
     return (
       Object.keys(this.legends)
         // for each field that was encoded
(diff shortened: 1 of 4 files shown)


packages/superset-ui-encodable/src/encoders/Encoder.ts
```diff
@@ -92,6 +95,55 @@ export default class Encoder<Config extends EncodingConfig> {
...
+          const firstEncoder = channelEncoders[0];
+          const definition = firstEncoder.definition as TypedFieldDef;
+          const createLegendItems = this.createLegendItemsFactory(field);
+
+          if (definition.type === 'nominal') {
+            return {
+              createLegendItems,
+              field,
+              items: createLegendItems(firstEncoder.getDomain(data)),
+              type: 'nominal' as const,
```
williaster: could set as `definition.type` like below, or could refactor to ternary just for `items` field

packages/superset-ui-encodable/src/encoders/Encoder.ts
```diff
@@ -92,6 +95,55 @@ export default class Encoder<Config extends EncodingConfig> {
...
+      this.getChannelEncoders().filter(e => isNotArray(e) && isValueDef(e.definition)),
+    ).concat(this.legends[field]);
+
+    return (domain: ChannelInput[]) =>
+      domain.map((input: ChannelInput) => ({
+        input,
+        output: channelEncoders.reduce(
+          (prev: Partial<{ [k in keyof Config]: Config[k]['1'] }>, curr) => {
+            const map = prev;
+            map[curr.name as keyof Config] = curr.encodeValue(`${input}`) as Value;
```
williaster: this works even if `ChannelInput` is not a `string`? why coerce to one? (could add comment)

packages/superset-ui-encodable/src/encoders/Encoder.ts
williaster: can you also add a comment for why we can't create items for non-nominal types?

williaster: what's the advantage of creating the `items` if the consumer can just invoke `createLegendItems` anyway and it only works if they pass `data`? seems simpler to just have a consistent behavior for different types?

packages/superset-ui-encodable/test/encoders/Encoder.test.ts
```diff
@@ -33,19 +44,157 @@ describe('Encoder', () => {
...
+              },
+            },
+          ],
+        },
+      ]);
+    });
+    it('ignore channels that are ValueDef', () => {
+      const legendInfo = factory
+        .create({
+          color: { type: 'nominal', field: 'brand', scale: { range: ['red', 'green', 'blue'] } },
```
williaster: should the input be a non-nominal type? (the above test tests this same thing to some extent right?)

kristw: It has some similarity with the test above it, but this one focuses on the omitted channels (`shape` which defaults to `ValueDef` `{ value: 'circle' }`)

packages/superset-ui-encodable/src/encoders/Encoder.ts
kristw: not sure why I did that. removed now.

packages/superset-ui-encodable/src/encoders/Encoder.ts
kristw: We can create items for non-nominal types but it will take time to implement. basically getting the right `ticks` from `domain` for linear, quantize, quantile, log, etc. scale. I kind of punt it for now.

PR #254, opened 2019-11-14
feat(encodable): make applyDomain() able to handle domain from dataset

kristw:
🏆 Enhancements

`applyDomain()` was applying `domain` field specified in scale config to `scale`. However, in real use case the developer needs to update the domain with domain from dataset. 

**Flow**

(1) user specifies `scaleConfig` (`json` or `js` object)
(2) passes it to `createScaleFromConfig` which creates `d3` scale and try to set `domain` if possible at this point. 
(3) set `domain` from dataset with respect to `domain` in `scaleConfig`, which is used as bounds. Now we have ready-to-use d3 scale.

The work before this PR stops at step 2. This PR enable step 3.

### DIFF
--- a/packages/superset-ui-encodable/src/utils/combineContinuousDomains.ts
+++ b/packages/superset-ui-encodable/src/utils/combineContinuousDomains.ts
@@ -1,0 +1,39 @@
+import { isEveryElementDefined, isDefined } from '../typeGuards/Base';
+
+/**
+ * Combine two continuous domain and ensure that the output
+ * does not go beyond fixedDomain
+ * @param fixedDomain
+ * @param inputDomain
+ */
+export default function combineContinuousDomains(
+  fixedDomain: (number | Date | null | undefined)[],
+  inputDomain?: (number | Date)[],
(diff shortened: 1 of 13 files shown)


packages/superset-ui-encodable/src/utils/combineContinuousDomains.ts
```diff
@@ -0,0 +1,39 @@
...
+) {
+  if (fixedDomain.length > 0 && isEveryElementDefined(fixedDomain)) {
+    return fixedDomain;
+  } else if (inputDomain) {
+    if (
+      fixedDomain.length === 2 &&
+      inputDomain.length === 2 &&
+      fixedDomain.filter(isDefined).length > 0
+    ) {
+      const [min1, max1] = fixedDomain;
```
felixcodes: maybe rename `min1, max1` to something along the lines of  `userSpecifiedMin` and `userSpecifiedMax` and `min2, max2` to `datasetMin`, and `datasetMax` for better readability

PR #305, opened 2020-03-03
feat: add NoResultsComponent to charts

etr2460:
🏆 Enhancements
Introduces new behavior for when no queryData is passed to a Superset Chart. Currently this never happens because Superset renders it's own error message, but this will be used in the future once we allow the backend to return empty query data as valid data.

Test plan:
Add new storybook examples and unit tests

Message on a small chart:
Screen Shot 2020-03-03 at 11 32 21 AM

Message on a large chart:
Screen Shot 2020-03-03 at 11 32 13 AM

### DIFF
--- a/packages/superset-ui-chart/src/components/NoResultsComponent.tsx
+++ b/packages/superset-ui-chart/src/components/NoResultsComponent.tsx
@@ -1,0 +1,45 @@
+import React, { CSSProperties, useMemo } from 'react';
+
+const MESSAGE_STYLES: CSSProperties = { maxWidth: 800 };
+const TITLE_STYLES: CSSProperties = { fontSize: 20, fontWeight: 'bold', paddingBottom: 8 };
+const BODY_STYLES: CSSProperties = { fontSize: 16 };
+
+const generateContainerStyles: (
+  height: number | string,
+  width: number | string,
+) => CSSProperties = (height: number | string, width: number | string) => ({
+  alignItems: 'center',
+  display: 'flex',
+  flexDirection: 'column',
+  height,
+  justifyContent: 'center',
+  padding: 16,
+  textAlign: 'center',
+  width,
+});
+
+type Props = {
+  className?: string;
+  height: number | string;
+  id?: string;
+  width: number | string;
+};
+
+const NoResultsComponent = ({ className, height, id, width }: Props) => {
+  const containerStyles = useMemo(() => generateContainerStyles(height, width), [height, width]);
+
+  return (
+    <div className={className} id={id} style={containerStyles}>
+      <div style={MESSAGE_STYLES}>
+        <div style={TITLE_STYLES}>No Results</div>
+        <div style={BODY_STYLES}>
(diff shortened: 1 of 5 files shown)


packages/superset-ui-chart/src/components/NoResultsComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+});
+
+type Props = {
+  height: number | string;
+  width: number | string;
+};
+
+const NoResultsComponent = ({ height, width }: Props) => (
+  <div style={generateContainerStyles(height, width)}>
+    <div style={{ maxWidth: 800 }}>
```
kristw: Please extract the style constants.

packages/superset-ui-chart/src/components/NoResultsComponent.tsx
```diff
@@ -0,0 +1,35 @@
...
+  width,
+});
+
+type Props = {
+  height: number | string;
+  width: number | string;
+};
+
+const NoResultsComponent = ({ height, width }: Props) => (
+  <div style={generateContainerStyles(height, width)}>
```
kristw: We can use `useMemo` hook for this to avoid creating new style object every time.

packages/superset-ui-demo/storybook/stories/index.js
```diff
@@ -1,3 +1,5 @@
+import 'core-js/stable';
```
kristw: Are these needed?

packages/superset-ui-chart/src/components/SuperChart.tsx
```diff
@@ -110,35 +111,48 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
...
+    });
+
+    let chart;
+    // Render the no results component if the query data is null or empty
+    if (
+      queryData == null ||
+      queryData.data === null ||
+      (Array.isArray(queryData.data) && queryData.data.length === 0)
+    ) {
+      chart = <NoResultsComponent height={height} width={width} />;
```
kristw: Perhaps it should take `id` and `className` as well?

packages/superset-ui-demo/storybook/stories/index.js
etr2460: oops, my bad. that must've gotten carried along from when i was trying to get the storybook to work

packages/superset-ui-chart/src/components/NoResultsComponent.tsx
etr2460: done

packages/superset-ui-chart/src/components/NoResultsComponent.tsx
etr2460: done

packages/superset-ui-chart/src/components/SuperChart.tsx
etr2460: makes sense, added

PR #307, opened 2020-03-04
feat: make formatTime and formatNumber accept undefined type

kristw:
🏆 Enhancements

* Make `formatTime` and `formatNumber` handle `undefined` type. The logic is already there, but the argument type was still `string`

cc @ktmud

### DIFF
--- a/packages/superset-ui-number-format/test/NumberFormatterRegistry.test.ts
+++ b/packages/superset-ui-number-format/test/NumberFormatterRegistry.test.ts
@@ -43,5 +43,8 @@ describe('NumberFormatterRegistry', () => {
       expect(registry.format('.2f', 100)).toEqual('100.00');
       expect(registry.format(',d', 100)).toEqual('100');
     });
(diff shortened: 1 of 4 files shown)


packages/superset-ui-number-format/test/NumberFormatterRegistry.test.ts
```diff
@@ -43,5 +43,8 @@ describe('NumberFormatterRegistry', () => {
       expect(registry.format('.2f', 100)).toEqual('100.00');
       expect(registry.format(',d', 100)).toEqual('100');
     });
+    it('falls back to the default formatter if the format is undefined', () => {
+      expect(registry.format(undefined, 1000)).toEqual('1k');
+    });
```
ktmud: Thanks for adding the test case!

ktmud:
Actually, this file needs to be updated, too:
https://github.com/apache-superset/superset-ui/blob/master/packages/superset-ui-number-format/src/NumberFormatterRegistrySingleton.ts#L12-L13

PR #308, opened 2020-03-04
feat: add more support for undefined format to number and time formatters

kristw:
🏆 Enhancements

* Follow up from #307 , make `formatTime`, `getTimeFormatter`, `formatNumber`, `getNumberFormatter` handle `undefined` format.

* Bring back support for empty string as format. This was due to the recent change from `||` to `??`.

The code used to be `format || defaultFormat`, and it was changed by `eslint` auto-fix to `format ?? defaultFormat`.

```ts
'' || 'abc' = 'abc'
'' ?? 'abc' = ''
```

### DIFF
--- a/packages/superset-ui-number-format/src/NumberFormatterRegistry.ts
+++ b/packages/superset-ui-number-format/src/NumberFormatterRegistry.ts
@@ -23,7 +23,11 @@ export default class NumberFormatterRegistry extends RegistryWithDefaultKey<
   }
 
   get(formatterId?: string) {
-    const targetFormat = `${formatterId ?? this.defaultKey}`.trim();
+    const targetFormat = `${
+      formatterId === null || typeof formatterId === 'undefined' || formatterId === ''
+        ? this.defaultKey
+        : formatterId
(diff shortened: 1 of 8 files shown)


packages/superset-ui-number-format/src/NumberFormatterRegistry.ts
```diff
@@ -23,7 +23,11 @@ export default class NumberFormatterRegistry extends RegistryWithDefaultKey<
   }
 
   get(formatterId?: string) {
-    const targetFormat = `${formatterId ?? this.defaultKey}`.trim();
+    const targetFormat = `${
+      formatterId === null || typeof formatterId === 'undefined' || formatterId === ''
+        ? this.defaultKey
+        : formatterId
+    }`.trim();
```
ktmud: I appreciate the explicitness, but could this be simplified as `formatterId ?? (formatterId || this.defaultKey)`.

kristw: eslint will try to change `||` to `??`

PR #322, opened 2020-03-28
feat: add validator package

kristw:
🏆 Enhancements

* Move validators from `incubator-superset` for usage in packages

```ts
import { legacyValidateInteger, legacyValidateNumber, validateNonEmpty } from '@superset-ui/validator';
```

There is a non-legacy `validateInteger` and `validateNumber` which has the logical behavior, but will break superset because it seems to expect the wrong output from the legacy functions at the moment.

@rusackas @villebro

### DIFF
--- a/packages/superset-ui-validator/src/validateNonEmpty.ts
+++ b/packages/superset-ui-validator/src/validateNonEmpty.ts
@@ -1,0 +1,8 @@
+import { t } from '@superset-ui/translation';
+
+export default function isNotEmpty(v: unknown) {
+  if (v === null || typeof v === 'undefined' || v === '' || (Array.isArray(v) && v.length === 0)) {
+    return t('cannot be empty');
+  }
+  return false;
+}
--- a/packages/superset-ui-validator/src/validateNonEmpty.ts
+++ b/packages/superset-ui-validator/src/validateNonEmpty.ts
@@ -1,6 +1,6 @@
 import { t } from '@superset-ui/translation';
 
-export default function isNotEmpty(v: unknown) {
+export default function validateNonEmpty(v: unknown) {
   if (v === null || typeof v === 'undefined' || v === '' || (Array.isArray(v) && v.length === 0)) {
     return t('cannot be empty');
   }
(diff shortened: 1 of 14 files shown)


packages/superset-ui-validator/src/validateNonEmpty.ts
```diff
@@ -0,0 +1,8 @@
+import { t } from '@superset-ui/translation';
+
+export default function isNotEmpty(v: unknown) {
```
ktmud: should the function name be consistent with the file name?

kristw:
> Can you give an example of how these validators are used in Superset? I'm also not sure about coupling translations with the validators. Most patterns I saw put the validation function and error messages in separate config fields.

They are from https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/explore/validators.js and used in https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/explore/controls.jsx and the control panel config files in https://github.com/apache/incubator-superset/tree/master/superset-frontend/src/explore/controlPanels

I am also not sure why they were designed this way. Trying to do minimum work to faciliate @rusackas and @villebro 's work as mentioned in one of the migration PR

rusackas:
> > Can you give an example of how these validators are used in Superset? I'm also not sure about coupling translations with the validators. Most patterns I saw put the validation function and error messages in separate config fields.
> 
> They are from https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/explore/validators.js and used in https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/explore/controls.jsx and the control panel config files in https://github.com/apache/incubator-superset/tree/master/superset-frontend/src/explore/controlPanels
> 
> I am also not sure why they were designed this way. Trying to do minimum work to faciliate @rusackas and @villebro 's work as mentioned in one of the migration PR

It is a little funny to couple translation (of error messages) with the validators. There are probably a few ways to address this, but for now it at least centralizes the burden of providing the string to the UI (i.e. it's DRY, at least). It's not something I'm particularly worried about, but I'm making a laundry list of tickets, and will put it on that list for discussion of how/when to remove this reliance. If you have the bandwidth to tackle it, that's fantastic too :D

PR #340, opened 2020-04-03
fix(legacy-plugin-chart-table): time column formatting

ktmud:
🐛 Bug Fix

Table chart displays time column incorrectly when "include time" is selected. This PR fixes it. It also fixes a bunch of linting issues surfaced by migrating the plugin packages to the `superset-ui` repo.

Snip20200403_50

### Before

Snip20200403_49

### After

image

## Test Plan

Added unit test.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,7 +2,7 @@ coverage:
   status:
     patch:
       default:
-        target: 50%
+        target: 40%
         threshold: 0%
     project:
       default:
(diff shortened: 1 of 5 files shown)


ktmud:
@kristw This is a quick fix to address the "NaN"s. Do you think it's still worth porting in `getTimeFormatterForGranularity` for the Table chart after this fix?

ktmud:
Merging this quick for now. Will adapt formatter by granularity later.

PR #344, opened 2020-04-06
feat(table): enable table filter and better typing

ktmud:
🏆 Enhancements
🐛 Bug Fix

- Add back "emit filter" support to Table chart.
- Refactor typing for both Table and Big Number, make more strict or give more hints when possible.

### DIFF
--- a/packages/superset-ui-chart-composition/package.json
+++ b/packages/superset-ui-chart-composition/package.json
@@ -26,12 +26,12 @@
     "access": "public"
   },
   "dependencies": {
-    "@types/react": "^16.7.17",
+    "@types/react": "^16.9.34",
     "@vx/responsive": "^0.0.195",
     "csstype": "^2.6.4"
   },
   "peerDependencies": {
     "@superset-ui/core": "^0.12.0",
-    "react": "^16.7.17"
+    "react": "^16.13.1"
   }
 }
--- a/packages/superset-ui-chart/src/clients/ChartClient.ts
+++ b/packages/superset-ui-chart/src/clients/ChartClient.ts
@@ -9,7 +9,7 @@ import {
 import { QueryFormData, Datasource } from '@superset-ui/query';
 import getChartBuildQueryRegistry from '../registries/ChartBuildQueryRegistrySingleton';
 import getChartMetadataRegistry from '../registries/ChartMetadataRegistrySingleton';
-import { QueryData } from '../models/ChartProps';
+import { QueryData } from '../types/QueryResponse';
 import { AnnotationLayerMetadata } from '../types/Annotation';
 import { PlainObject } from '../types/Base';
 
@@ -94,7 +94,10 @@ export default class ChartClient {
           },
           ...options,
         } as RequestConfig)
-        .then(response => response.json as Json);
+        .then(response => {
+          // let's assume response.json always has the shape of QueryData
+          return response.json as QueryData;
+        });
     }
 
     return Promise.reject(new Error(`Unknown chart type: ${visType}`));
--- a/packages/superset-ui-chart/src/models/ChartProps.ts
+++ b/packages/superset-ui-chart/src/models/ChartProps.ts
@@ -15,8 +15,12 @@ type ChartPropsSelector = (c: ChartPropsConfig) => ChartProps;
 
 /** Optional field for event handlers, renderers */
 type Hooks = {
+  /**
+   * sync active filters between chart and dashboard, "add" actually
+   * also handles "change" and "remove".
+   */
   onAddFilter?: (newFilters: DataRecordFilters, merge?: boolean) => void;
-  /** handle errors  */
+  /** handle errors */
   onError?: HandlerFunction;
   /** use the vis as control to update state */
   setControlValue?: HandlerFunction;
--- a/packages/superset-ui-chart/src/models/ChartProps.ts
+++ b/packages/superset-ui-chart/src/models/ChartProps.ts
@@ -1,30 +1,28 @@
 import { createSelector } from 'reselect';
 import { convertKeysToCamelCase } from '@superset-ui/core';
+import { Datasource } from '@superset-ui/query';
 import { HandlerFunction, PlainObject } from '../types/Base';
+import { QueryData, DataRecordFilters } from '../types/QueryResponse';
 
 // TODO: more specific typing for these fields of ChartProps
 type AnnotationData = PlainObject;
-type CamelCaseDatasource = PlainObject;
 type SnakeCaseDatasource = PlainObject;
 type CamelCaseFormData = PlainObject;
 type SnakeCaseFormData = PlainObject;
-export type QueryData = PlainObject;
-/** Initial values for the visualizations, currently used by only filter_box and table */
-type InitialValues = PlainObject;
+type RawFormData = CamelCaseFormData | SnakeCaseFormData;
+
 type ChartPropsSelector = (c: ChartPropsConfig) => ChartProps;
 
 /** Optional field for event handlers, renderers */
 type Hooks = {
-  /** handle adding filters  */
-  onAddFilter?: HandlerFunction;
+  onAddFilter?: (newFilters: DataRecordFilters, merge?: boolean) => void;
   /** handle errors  */
   onError?: HandlerFunction;
   /** use the vis as control to update state */
   setControlValue?: HandlerFunction;
   /** handle tooltip */
   setTooltip?: HandlerFunction;
-  [key: string]: any;
-};
+} & PlainObject;
 
 /**
  * Preferred format for ChartProps config
@@ -37,9 +35,9 @@ export interface ChartPropsConfig {
    * Formerly called "filters", which was misleading because it is actually
    * initial values of the filter_box and table vis
    */
-  initialValues?: InitialValues;
+  initialValues?: DataRecordFilters;
   /** Main configuration of the chart */
-  formData?: SnakeCaseFormData;
+  formData?: RawFormData;
   /** Chart height */
   height?: number;
   /** Programmatic overrides such as event handlers, renderers */
@@ -53,22 +51,20 @@ export interface ChartPropsConfig {
 const DEFAULT_WIDTH = 800;
 const DEFAULT_HEIGHT = 600;
 
-export default class ChartProps<
-  FormDataType extends CamelCaseFormData | SnakeCaseFormData = CamelCaseFormData
-> {
+export default class ChartProps {
   static createSelector: () => ChartPropsSelector;
 
   annotationData: AnnotationData;
 
-  datasource: CamelCaseDatasource;
+  datasource: Datasource;
 
   rawDatasource: SnakeCaseDatasource;
 
-  initialValues: InitialValues;
+  initialValues: DataRecordFilters;
 
   formData: CamelCaseFormData;
 
-  rawFormData: SnakeCaseFormData | CamelCaseFormData;
+  rawFormData: RawFormData;
 
   height: number;
 
@@ -82,7 +78,7 @@ export default class ChartProps<
     const {
       annotationData = {},
       datasource = {},
-      formData = {} as FormDataType,
+      formData = {},
       hooks = {},
       initialValues = {},
       queryData = {},
--- a/packages/superset-ui-query/src/types/Datasource.ts
+++ b/packages/superset-ui-query/src/types/Datasource.ts
@@ -14,6 +14,7 @@ export interface Datasource {
   columns: Column[];
   metrics: QueryObjectMetric[];
   description?: string;
+  // key is column names (labels)
   columnFormats?: {
     [key: string]: string;
   };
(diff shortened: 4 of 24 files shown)


packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -37,7 +35,7 @@ export interface ChartPropsConfig {
    * Formerly called "filters", which was misleading because it is actually
    * initial values of the filter_box and table vis
    */
-  initialValues?: InitialValues;
+  activeFilters?: DataRecordFilters;
```
ktmud, line 38: Renamed `initialValues` to `activeFilters` as it's not super clear what things' initial values it is. Superset frontend will update accordingly.

packages/superset-ui-chart-composition/package.json
```diff
@@ -26,12 +26,12 @@
...
   "dependencies": {
-    "@types/react": "^16.7.17",
+    "@types/react": "^16.9.32",
     "@vx/responsive": "^0.0.195",
     "csstype": "^2.6.4"
   },
   "peerDependencies": {
     "@superset-ui/core": "^0.12.0",
-    "react": "^16.7.17"
+    "react": "^16.13.1"
```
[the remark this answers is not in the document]
ktmud: I saw some typing errors related to `@types/react` and thought it'd be easier to just upgrade everything instead of hunting down which plugin imported the incompatible version. Will upgrade Superset main repo, too, when this is merged.

packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -82,10 +80,10 @@ export default class ChartProps<
     const {
       annotationData = {},
       datasource = {},
-      formData = {} as FormDataType,
+      formData = {},
       hooks = {},
-      initialValues = {},
-      queryData = {},
+      activeFilters = {},
+      queryData = { data: { records: [], columns: [] } },
```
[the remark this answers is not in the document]
ktmud: Yes, I realized that later and has since updated `QueryResponse`, but forgot about this.

packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -1,30 +1,28 @@
...
-type InitialValues = PlainObject;
+
 type ChartPropsSelector = (c: ChartPropsConfig) => ChartProps;
 
 /** Optional field for event handlers, renderers */
 type Hooks = {
-  /** handle adding filters  */
-  onAddFilter?: HandlerFunction;
+  /** handle adding and removing filters  */
+  onChangeFilter?: (newFilters: DataRecordFilters, merge?: boolean) => void;
```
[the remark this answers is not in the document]
ktmud: Whether to merge the values of the selected filter to existing values, or override them. Seems not actively in use though.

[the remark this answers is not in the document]
ktmud: Re field name change: the original name was not used in any other plugin anyway so I thought it was a fairly safe change.

packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -82,10 +80,10 @@ export default class ChartProps<
     const {
       annotationData = {},
       datasource = {},
-      formData = {} as FormDataType,
+      formData = {},
```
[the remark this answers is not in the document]
ktmud: This is intentional as TS is already able to infer `formData` from `ChartPropsConfig` above.

plugins/big-number/src/BigNumber/transformProps.ts
```diff
@@ -19,25 +19,58 @@
...
+  };
+  metric?:
+    | {
+        label: string;
+      }
+    | string;
+  compareLag?: string | number;
+  yAxisFormat?: string;
+  timeGrainSqla?: TimeGranularity;
+} & ChartProps['formData'];
```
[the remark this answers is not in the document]
ktmud: Thought it was more semantically correct to include it. Can't probably drop since we've been exhaustive listing all the fields.

packages/superset-ui-chart/src/models/ChartProps.ts
ktmud: Reverted it back just to be safe.

packages/superset-ui-chart/src/models/ChartProps.ts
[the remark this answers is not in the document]
ktmud: Reverted it back also just to be safe and also to make the diff on main repo smaller.

packages/superset-ui-chart/src/clients/ChartClient.ts
```diff
@@ -94,7 +94,10 @@ export default class ChartClient {
           },
           ...options,
         } as RequestConfig)
-        .then(response => response.json as Json);
+        .then(response => {
+          // let's assume response.json always has the shape of QueryData
+          return response.json as QueryData;
```
williaster, line 99: is this needed since the method return type is `Promise<QueryData>`?

packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -1,30 +1,28 @@
...
-type InitialValues = PlainObject;
+type RawFormData = CamelCaseFormData | SnakeCaseFormData;
+
 type ChartPropsSelector = (c: ChartPropsConfig) => ChartProps;
 
 /** Optional field for event handlers, renderers */
 type Hooks = {
-  /** handle adding filters  */
-  onAddFilter?: HandlerFunction;
+  onAddFilter?: (newFilters: DataRecordFilters, merge?: boolean) => void;
```
williaster, line 18: should we add back the doc string?

packages/superset-ui-chart/src/models/ChartProps.ts
```diff
@@ -1,30 +1,28 @@
...
   /** handle errors  */
   onError?: HandlerFunction;
   /** use the vis as control to update state */
   setControlValue?: HandlerFunction;
   /** handle tooltip */
   setTooltip?: HandlerFunction;
-  [key: string]: any;
-};
+} & PlainObject;
```
williaster, line 25: why is the `PlainObject` needed?

packages/superset-ui-query/src/types/Datasource.ts
```diff
@@ -10,8 +10,14 @@ export enum DatasourceType {
 export interface Datasource {
   id: number;
   name: string;
-  description?: string;
   type: DatasourceType;
   columns: Column[];
   metrics: QueryObjectMetric[];
+  description?: string;
+  columnFormats?: {
+    [key: string]: string;
```
williaster, line 18: could we be any more specific about what `key` is here + `verboseMap` below? like `name` or `id` or something?

plugins/big-number/src/BigNumber/transformProps.ts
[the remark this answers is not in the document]
williaster: two questions:
- should it be `Partial<ChartProps['formData']>`? or all of them?
- are there duplicate keys in `ChartProps['formData']`? I'm not sure order of override for TS, just wanted to make sure if there are duplicates that the `BigNumberFormData` values take precedent.

plugins/big-number/src/BigNumber/transformProps.ts
```diff
@@ -19,25 +19,56 @@
...
 };
 
-export default function transformProps(chartProps: ChartProps) {
-  const { width, height, formData, queryData } = chartProps;
+export default function transformProps(chartProps: BignumberChartProps) {
+  const { width, height, queryData, formData } = chartProps;
   const {
     colorPicker,
-    compareLag: compareLagInput,
+    compareLag: compareLag_,
```
williaster, line 68: readability nit – I think `compareLagInput` is a clearer name, not sure what `_` means tho I guess it's used a few places here

plugins/table/src/ReactDataTable.tsx
```diff
@@ -109,16 +110,16 @@ export default function ReactDataTable(props: DataTableProps) {
...
+    if (isTimeColumn(key)) {
       let value = val;
       if (typeof val === 'string') {
         // force UTC time zone if is an ISO timestamp without timezone
         // e.g. "2020-10-12T00:00:00"
         value = val.match(/T(\d{2}:){2}\d{2}$/) ? `${val}Z` : val;
         value = new Date(value);
       }
-      return formatTimestamp(value) as string;
+      return formatTimestamp(value as Date | number | null) as string;
```
williaster, line 122: can you set this above instead of coercing? `let value: Date | number | null = val; ...`

plugins/table/src/Table.css
```diff
@@ -25,6 +25,16 @@
 .superset-legacy-chart-table .dt-metric {
   text-align: right;
 }
+.superset-legacy-chart-table td.dt-is-filter {
+  cursor: pointer;
+}
+.superset-legacy-chart-table td.dt-is-filter:hover {
+  background-color: linen;
```
williaster, line 32: curious how these were chosen? should we get any design feedback?

packages/superset-ui-chart/src/clients/ChartClient.ts
ktmud: Yes, it is, because `response.json` might be `undefined`, then it becomes unassignable to `Promise<QueryData>`. Previously it was hinted as `Json`, I just changed it to `QueryData` to be more strict.

packages/superset-ui-chart/src/models/ChartProps.ts
ktmud: The doc string doesn't seem very helpful to me. `onAddFilter` is basically synonymous to `handle adding filter`.

packages/superset-ui-chart/src/models/ChartProps.ts
ktmud: `PlainObject` is basically `[key: string]: any`. Probably not needed, but didn't wanna break anything.

packages/superset-ui-query/src/types/Datasource.ts
ktmud: I always thought it's good to be generic in indexable types so that new TypeScript users don't confuse it with an actual property name.

ktmud: I'll add some inline comments instead.

plugins/big-number/src/BigNumber/transformProps.ts
ktmud: I use dangling "_" in for arguments that will be immediately overridden by simple argument processing. It indicates a sense of temporariness and keeps the variable names clean.

plugins/table/src/ReactDataTable.tsx
ktmud: Actually tried that, but ran into error when passing it to `new Date`. Obviously `new Date` doesn't accept `null`.

plugins/table/src/Table.css
ktmud: The [original color](https://github.com/apache/incubator-superset/issues/8273) is:

```
  background-color: lighten(desaturate(@brand-primary, 50%), 50%);
```
https://github.com/apache/incubator-superset/pull/9593/commits/cbf38ff466e81f33efd828d80bdf291391437651#diff-9479060a8cb28b444d0ce7f5df0a33ccL577


image

Which translates to `#bce9e5`. I just picked a named color that is close to it and does not look bad. Wanted to use the original color, but it kind of looked too dark to me.

This will probably eventually become customizable with the introduction of the [superset-ui/style](https://github.com/apache-superset/superset-ui/tree/master/packages/superset-ui-style
) package.

plugins/big-number/src/BigNumber/transformProps.ts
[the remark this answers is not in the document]
ktmud: <strike>Actually, it'd be easier to just do this because some shared properties are indeed useful to keep in ChartProps and it doesn't make sense to exhaustively list them for every chart type.</strike>

Strike again. `formData` and `queryData` are basically `PlainObject`, which accepts any key with any values. `Partial` or not it doesn't make any difference, at least for now. I was thinking `& ChartProps`.

@williaster thanks for the reminder about overrides. I should have placed `ChartProps['formData']` at the front.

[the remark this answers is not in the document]
ktmud: I think I'm going to avoid `Partial` as it makes all properties optional, which is probably not what we want in this case.

plugins/big-number/src/BigNumber/transformProps.ts
```diff
@@ -19,25 +19,58 @@
...
+}
+
 // we trust both the x (time) and y (big number) to be numeric
-type BigNumberDatum = {
-  [TIME_COLUMN]: number;
+export interface BigNumberDatum {
   [key: string]: number | null;
+}
+
+export type BigNumberData = BigNumberDatum[];
```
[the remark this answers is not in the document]
ktmud: Done.

PR #352, opened 2020-04-10
feat: allow toggling of table viz's bar chart backgrounds

rusackas:
🏆 Enhancements

Fixes, in coordination with [this parallel PR](https://github.com/apache/incubator-superset/pull/9500), Superset issue #9499

### DIFF
--- a/plugins/table/src/ReactDataTable.tsx
+++ b/plugins/table/src/ReactDataTable.tsx
@@ -244,7 +244,7 @@ export default function ReactDataTable(props: DataTableProps) {
                   // only set innerHTML for actual html content, this saves time
                   dangerouslySetInnerHTML={isHtml ? { __html: text } : undefined}
                   data-sort={val}
-                  className={showCellBar ? 'text-right' : ''}
+                  className={keyIsMetric ? 'text-right' : ''}
                   style={{
                     backgroundImage: showCellBar ? cellBar(key, val as number) : undefined,
                   }}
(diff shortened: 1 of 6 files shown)


rusackas:
@krist this one is minor, but would love to know how you prefer to approach these double-PRs as a coordinated effort. We can discuss on Slack if it's easier :D

plugins/table/src/ReactDataTable.tsx
```diff
@@ -236,15 +237,16 @@ export default function ReactDataTable(props: DataTableProps) {
...
               const isHtml = !keyIsMetric && isProbablyHTML(text);
+              const showCellBar = keyIsMetric && !hideCellBars;
               return (
                 <td
                   key={key}
                   // only set innerHTML for actual html content, this saves time
                   dangerouslySetInnerHTML={isHtml ? { __html: text } : undefined}
                   data-sort={val}
-                  className={keyIsMetric ? 'text-right' : ''}
+                  className={showCellBar ? 'text-right' : ''}
```
ktmud, line 247: It probably makes sense to always align metrics to the right even if we don't need to show the bar chart?

rusackas: Sure, that's fine... it looks more wrong than right to me to have them right-aligned in many cases, but that all depends on the number formatting. I guess there's no need to squeeze that extra change, as it wasn't part of the original issue. I'll change it back.

rusackas:
Thanks for the review, @ktmud :)

If you don't mind reviewing/approving the [superset PR](https://github.com/apache/incubator-superset/pull/9500), I'll merge both, unless you see any deployment problems in doing so.

kristw:
published as `0.12.14`

PR #357, opened 2020-04-11
feat: make CategoricalScale compatible with D3's ScaleOrdinal

kristw:
🏆 Enhancements

`CategoricalScale` was originally written to provide similar functionality with `ScaleOrdinal` but allow color overrides (both at the scale itself and from its parent namespace).

However, due to its different class signature (function names, return types, etc.) from `ScaleOrdinal`, this caused some frictions when adopting it in place of `ScaleOrdinal`, both in Superset and other consumer apps. 

This PR add function `copy`, `domain`, `range` and `unknown` to `CategoricalScale` so it is fully compatible with `ScaleOrdinal` signature. 

```ts
// This is a valid statement in typescript. 
// Developer then can use scale similar to how they use D3's scaleOrdinal
const scale: ScaleOrdinal<{ toString(): string }, string> = new CategoricalColorScale([]);
```

This PR also expose new feature `.unknown()` similar to `D3`'s which allow bucketing of unknown value that are not in the specified domain.

### DIFF
--- a/packages/superset-ui-color/src/CategoricalColorScale.ts
+++ b/packages/superset-ui-color/src/CategoricalColorScale.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-dupe-class-members */
 import { ExtensibleFunction } from '@superset-ui/core';
+import { scaleOrdinal, ScaleOrdinal } from 'd3-scale';
 import { ColorsLookup } from './types';
 import stringifyAndTrim from './stringifyAndTrim';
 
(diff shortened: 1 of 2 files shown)


packages/superset-ui-color/src/CategoricalColorScale.ts
```diff
@@ -78,4 +79,85 @@ export default class CategoricalColorScale extends ExtensibleFunction {
       ...this.parentForcedColors,
```
ktmud, line 79: It is intentional to allow color map being overridden by parent?

kristw: Yes, a scale can be in a namespace. We allow override at namespace level. so that should override scale-level setting.

PR #367, opened 2020-04-15
feat: migrate heatmap, horizon, iframe and markup

kristw:
🏆 Enhancements

* Migrate `heatmap`, `horizon`, `iframe` and `markup` plugin


packages/superset-ui-demo/.storybook/main.js
```diff
@@ -20,6 +24,17 @@ module.exports = {
...
+    // These are excluded by the default rule
+    // because they reside in node_modules
+    config.module.rules.push({
+      include: PLUGIN_PACKAGES_PATH_REGEXP,
+      exclude: /node_modules/,
+      test: /\.jsx?$/,
+      use: config.module.rules[0].use,
+    });
+
     config.module.rules.push({
```
kristw, line 38: This rule only applies to typescript. but there are older packages that are jsx.

plugins/legacy-plugin-chart-horizon/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```js
+import HorizonChartPlugin from '@superset-ui/legacy-plugin-chart-horizon';
+
+new HorizonChartPlugin()
+  .configure({ key: 'horizon' })
+  .register();
+```
+
+Then use it via `SuperChart`. See [storybook](https://apache-superset.github.io/superset-ui-plugins/?selectedKind=plugin-chart-horizon) for more details.
```
ktmud, line 20: These storybook links would also need an update. But we can probably do it in a batch later.

kristw: Sounds good. I'll focus on making them compile and storybook work first.

PR #368, opened 2020-04-15
feat: migrate 4 chart plugins

kristw:
🏆 Enhancements

Migrate `paired-t-test`, `parallel-coordinates`, `partition` and `pivot-table`.


plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
```diff
@@ -0,0 +1,110 @@
...
+  $container.find('tbody tr').each(function eachRow() {
+    $(this)
+      .find('td')
+      .each(function each(i) {
+        const metric = cols[i];
+        const format = columnFormats[metric] || numberFormat || '.3s';
+        const tdText = $(this)[0].textContent;
+        if (!Number.isNaN(tdText) && tdText !== '' && tdText.trim().toLowerCase() !== 'null') {
+          $(this)[0].textContent = formatNumber(format, tdText);
+          $(this).attr('data-sort', tdText);
```
ktmud, line 77: Note: null values are handled differently in charts. I just changed Big Number to render it as "N/A". Probably makes sense to consolidate the behaviors as we move along refactoring charts.

kristw: Umm, perhaps we should discuss with Erik before changing that. He was doing work on missing data earlier.

PR #369, opened 2020-04-15
build: add github workflow for build and test

kristw:
🏠 Internal

Try using github actions for ci

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         fail_ci_if_error: true
 
     # - if: github.ref == 'refs/heads/master'
-    - if: github.ref == 'refs/heads/kristw-ci'
+    if: github.ref == 'refs/heads/kristw-ci'
       steps:
       - name: Configure npm and git
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: Node.js CI
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,57 @@
+name: Node.js CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
ktmud, line 6: Should probably enable pull requests on non-master branch (i.e, feature branch), too.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
+name: Node.js CI
+
+on: [ push ]
```
ktmud, line 3: Does this mean the job will not run on pull requests?

kristw: It does. I had both `push` and `pull_request` at first and it ends up running the test twice.

kristw: Looked up a few repos and `vega-lite` use this for the tests. This should mean any push to any branch.

ktmud: `pull_request` is still different though because it runs on the merged code. If the PR branch didn't rebase before merge, the merged code could still break.

kristw: umm, that makes sense now. i will add back the `pull_request`

kristw:
@ktmud green now. If approve, i'll merge and disable travis. Then work on the auto-publish in separate pr.

PR #375, opened 2020-04-17
feat: migrate chart plugins

kristw:
🏆 Enhancements

Migrate
* force directed
* histogram
* mapbox
* rose
* sankey
* sankey loop
* sunburst
* treemap
* world map

Most of the changes are `eslint` auto fix via `yarn lint:fix`.
And update storybook structure to match the new way.


plugins/legacy-plugin-chart-sunburst/src/Sunburst.js
```diff
@@ -58,6 +58,75 @@ function getAncestors(node) {
   return path;
 }
 
+function buildHierarchy(rows) {
```
kristw, line 61: Just move function outside

kristw:
Removal PR: https://github.com/apache-superset/superset-ui-plugins/pull/440

PR #384, opened 2020-04-21
feat: add retry to callApi

etr2460:
🏆 Enhancements

Adds the `fetch-retry` package to automatically retry any requests that fail due to network issues or 503 errors. This seems like a pretty safe approach to apply to all requests, but would appreciate feedback, especially if you think it's likely to break something.

Test Plan:
add/update unit tests

to: @kristw @williaster @ktmud @rusackas

### DIFF
--- a/packages/superset-ui-connection/src/SupersetClientClass.ts
+++ b/packages/superset-ui-connection/src/SupersetClientClass.ts
@@ -5,6 +5,7 @@ import {
   Credentials,
   CsrfPromise,
   CsrfToken,
+  FetchRetryOptions,
   Headers,
   Host,
   Mode,
@@ -12,11 +13,13 @@ import {
   RequestConfig,
   SupersetClientResponse,
 } from './types';
+import { DEFAULT_FETCH_RETRY_OPTIONS } from './constants';
 
 export default class SupersetClientClass {
   credentials: Credentials;
   csrfToken?: CsrfToken;
   csrfPromise?: CsrfPromise;
+  fetchRetryOptions?: FetchRetryOptions;
   protocol: Protocol;
   host: Host;
   headers: Headers;
@@ -27,6 +30,7 @@ export default class SupersetClientClass {
     protocol = 'http:',
     host = 'localhost',
     headers = {},
+    fetchRetryOptions = {},
     mode = 'same-origin',
     timeout,
     credentials = undefined,
@@ -40,6 +44,7 @@ export default class SupersetClientClass {
     this.credentials = credentials;
     this.csrfToken = csrfToken;
     this.csrfPromise = undefined;
+    this.fetchRetryOptions = { ...DEFAULT_FETCH_RETRY_OPTIONS, ...fetchRetryOptions };
 
     if (typeof this.csrfToken === 'string') {
       this.headers = { ...this.headers, 'X-CSRFToken': this.csrfToken };
@@ -80,6 +85,7 @@ export default class SupersetClientClass {
     body,
     credentials,
     endpoint,
+    fetchRetryOptions,
     headers,
     host,
     method,
@@ -95,6 +101,7 @@ export default class SupersetClientClass {
       callApi({
         body,
         credentials: credentials ?? this.credentials,
+        fetchRetryOptions,
         headers: { ...this.headers, ...headers },
         method,
         mode: mode ?? this.mode,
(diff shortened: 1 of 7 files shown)


packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -1,7 +1,12 @@
 import 'whatwg-fetch';
+import fetchRetry from 'fetch-retry';
 import { CallApi } from '../types';
 import { CACHE_AVAILABLE, CACHE_KEY, HTTP_STATUS_NOT_MODIFIED, HTTP_STATUS_OK } from '../constants';
 
+const RETRY_COUNT = 3;
+const RETRY_DELAY_MS = 1000;
+const RETRY_ON_HTTP_STATUS_CODES = [503];
```
williaster, line 8: do you think we should make this configurable?  like make this the default but override-able? 

we could do that for all of the constants actually. no reason not to expose them for more granular customization?

packages/superset-ui-connection/test/callApi/callApi.test.ts
```diff
@@ -437,7 +439,7 @@ describe('callApi()', () => {
     });
   });
 
-  it('rejects if the request throws', () => {
+  it('rejects after retrying thrice if the request throws', () => {
```
williaster, line 442: could update the tests to match the retry config variables specified if we expose them

williaster: this is pretty awesome tho that the throwing is delayed

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -1,7 +1,12 @@
 import 'whatwg-fetch';
+import fetchRetry from 'fetch-retry';
 import { CallApi } from '../types';
 import { CACHE_AVAILABLE, CACHE_KEY, HTTP_STATUS_NOT_MODIFIED, HTTP_STATUS_OK } from '../constants';
 
+const RETRY_COUNT = 3;
```
williaster, line 6: since this is a pretty big change, I wonder if we should make this `2` by default for a _total_ of 3 requests? not really sure what's standard.

etr2460: 3 is the default for the package and is a standard i've seen before. if we make this configurable i think it's fine to default it though

packages/superset-ui-connection/src/callApi/callApi.ts
```diff
@@ -16,6 +21,12 @@ export default function callApi({
   stringify = true,
   url,
```
ktmud, line 22: Can we add the retry options to `callApi` arguments, too, in case people want to override the defaults?

It'd also be preferred to make this an opt-in rather than opt-out. Can also add an API to `superset-ui/connection` to change the global defaults.

packages/superset-ui-connection/src/callApi/callApi.ts
etr2460: That's a good call, I'll move it into the SupersetClient class.

packages/superset-ui-connection/test/callApi/callApi.test.ts
etr2460: updated

packages/superset-ui-connection/src/callApi/callApi.ts
etr2460: I've added the retry options to the SupersetClient config. However, I think it's fair to make this opt out. I can add a comment in incubator-superset that this was added, but I think the common path should include QOL features like automatic retried on network issues

packages/superset-ui-connection/src/SupersetClientClass.ts
```diff
@@ -40,6 +44,7 @@ export default class SupersetClientClass {
     this.credentials = credentials;
     this.csrfToken = csrfToken;
     this.csrfPromise = undefined;
+    this.fetchRetryOptions = fetchRetryOptions;
```
ktmud, line 47: Maybe `this.fetchRetryOptions = { ...DEFAULT_FETCH_RETRY_OPTIONS, ...fetchRetryOptions }` and     `fetchRetryOptions = {}` in `L33` in case people only want to override one of the options.

etr2460: good call

PR #387, opened 2020-04-22
fix: unset row_limit when it's not a number

etr2460:
🐛 Bug Fix

Recently Superset was changed to not allow row_limit to be null, it can only be unset or a number. This fixes the library to match it

to: @williaster @ktmud @villebro @kristw

### DIFF
--- a/packages/superset-ui-query/src/buildQueryObject.ts
+++ b/packages/superset-ui-query/src/buildQueryObject.ts
@@ -33,6 +33,7 @@ export default function buildQueryObject<T extends QueryFormData>(formData: T): 
   } = formData;
 
   const groupbySet = new Set([...columns, ...groupby]);
+  const numericRowLimit = Number(row_limit);
 
   const queryObject: QueryObject = {
     extras: processExtras(formData),
@@ -42,7 +43,7 @@ export default function buildQueryObject<T extends QueryFormData>(formData: T): 
     metrics: processMetrics(formData),
     order_desc: typeof order_desc === 'undefined' ? true : order_desc,
     orderby: [],
-    row_limit: !Number(row_limit) && row_limit !== 0 ? undefined : Number(row_limit),
+    row_limit: row_limit == null || isNaN(numericRowLimit) ? undefined : numericRowLimit,
     since,
     time_range,
     timeseries_limit: limit ? Number(limit) : 0,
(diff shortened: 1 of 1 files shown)


packages/superset-ui-query/src/buildQueryObject.ts
```diff
@@ -42,7 +42,7 @@ export default function buildQueryObject<T extends QueryFormData>(formData: T):
     metrics: processMetrics(formData),
     order_desc: typeof order_desc === 'undefined' ? true : order_desc,
     orderby: [],
-    row_limit: Number(row_limit),
+    row_limit: !Number(row_limit) && row_limit !== 0 ? undefined : Number(row_limit),
```
williaster, line 45: if we're converting `row_limit` to a `number`, won't `row_limit !== 0` _always_ be `false`?

could this be something like 
```ts
const numericRowLimit = Number(row_limit);
...
row_limit: row_limit == null || isNaN(numericRowLimit) ? undefined : numericRowLimit,
```

etr2460: I'm catching the case where if row_limit = 0, we don't want to set it to undefined.

`!Number(0) is true`, but we don't want row_limit to be undefined

etr2460: maybe this should be Number(row_limit) though...

etr2460: i see your point, updated

villebro:
My apologies, apparently one of my changes caused this problem. I'll try not to make breaking changes to `QueryContext`/`QueryObject` going forward without including all central stakeholders in the review process. Who would be the right person(s) to tag when making changes to these classes?

williaster:
thanks @villebro. I think tagging @kristw or myself or @ktmud would be good 👍 (apologies for not keeping a close pulse on the refactoring, I think you may already be tagging Krist + Jesse)

williaster:
I think **ideally** there would be some integration tests between the main repo + these packages, but I think that's likely easiest once the API is refactored and the app itself is using these client calls.

packages/superset-ui-query/src/buildQueryObject.ts
```diff
@@ -42,7 +43,7 @@ export default function buildQueryObject<T extends QueryFormData>(formData: T):
     metrics: processMetrics(formData),
     order_desc: typeof order_desc === 'undefined' ? true : order_desc,
     orderby: [],
-    row_limit: Number(row_limit),
+    row_limit: row_limit == null || isNaN(numericRowLimit) ? undefined : numericRowLimit,
```
kristw, line 46: Should use `Number.isNaN`, instead of `isNaN`

ktmud: I'm fixing this in my test coverage PR: https://github.com/apache-superset/superset-ui/pull/390/files

kristw: Thank you!

PR #388, opened 2020-04-22
build: clean up and test auto release

ktmud:
🏆 Enhancements

Some minor changes to test auto-release setup by @kristw .

Reading from the code, it seems we are only [bumping versions](https://github.com/lerna/lerna/tree/master/commands/publish#positionals) from manually updated `package.json`.

1. A minor fix to speed up commit hook
2. A minor change in BigNumber to test auto release
3. Clean up some irrelevant code that is not used in the current auto-release process.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,12 +2,11 @@ name: release-workflow
 
 on:
   push:
-    branches:
-      - 'master'
+    branches: ['master']
 
 jobs:
   build:
-    name: Bump version and publish package(s)
+    name: Build and release
 
     runs-on: ubuntu-latest
 
@@ -22,7 +21,7 @@ jobs:
         fetch-depth: 0
     - name: Get tags and filter trigger tags
       run: |
-        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        git fetch --depth=1 origin "+refs/tags/*:refs/tags/*"
         git fetch --prune --unshallow
         git tag -d `git tag | grep -E '^trigger-'`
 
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -54,34 +54,17 @@ jobs:
...
-        git update-index --assume-unchanged .npmrc
-        git tag -d `git tag | grep -E '^trigger-'`
-        yarn ci:release-from-tag
-      env:
-        NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        GH_TOKEN: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+        echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" >> $HOME/.npmrc 2> /dev/null
+        npm whoami
+        yarn ci:release-from-tag  # lerna publish from-package
```
kristw, line 70: Please do not change these yet. 
`yarn ci:release-from-tag` actually did not bump the package version, only publish to npm by comparing the versions in `package.json` and the latest version on `npm`.

I had called a command that actually bump the package version called before, but removed that temporarily due to conflict with protected `master` branch and I didn't have time to fix and test. These credentials that you have removed in the PR are necessary for the version bump, git push and create github release to succeed.

ktmud: Thought we'd be manually update package versions and only use `lerna publish from-package`. If we are going to automatically update versions and make commits in CI, should we publish with `--canary` instead?

ktmud: I sense it'd be quite troublesome if we let CI automatically make new commits. It'd basically mean every merge will create another commit in Git.

I like what [zeit/now is doing](https://github.com/zeit/now/blob/master/utils/publish.sh): publish stable versions from git tags, but canary versions from pushes to the master branch.

kristw: Canary sounds good. We can discuss more. I was hesitating about publishing every version as well so I pulled it out.

ktmud:
This is no longer needed for now since `yarn ci:create-patch-version` seems to work quite well.

PR #362, opened 2020-04-14
feat: style package for theming and styled components

suddjian:
💔 Breaking Changes

🏆 Enhancements

Adds a new package. For now the package contents are just porting the theme from Superset. This is so it can be used by viz plugins and other superset-ui components.

📜 Documentation

See readme

🐛 Bug Fix

🏠 Internal

PR #389, opened 2020-04-22
chore: add shared plugin controls utilities

suddjian:
💔 Breaking Changes

🏆 Enhancements

This moves several files from incubator-superset into `@superset-ui/chart`. These will allow moving the plugin controls config out of Superset's `setupPlugins.ts` and into the individual plugins for all the plugins. This covers the dependencies for all plugin controls except the DeckGL plugins. Those will come later.

The new modules are not exported directly by `@superset-ui/chart`, and must be imported by their directory paths under `@superset-ui/chart/controls/[...]`. This is to keep things organized and to avoid bloating the package size.

📜 Documentation

Added a paragraph to the readme. It's not enough documentation really, but things are very much in flux and I'd prefer to write more comprehensive docs once the plugin situation is more locked down.

🐛 Bug Fix

🏠 Internal


packages/superset-ui-chart/src/controls/D3Formatting.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+// D3 specific formatting config
+
+export const D3_FORMAT_DOCS = 'D3 format syntax: https://github.com/d3/d3-format';
+
+// input choices & options
+export const D3_FORMAT_OPTIONS = [
+  ['SMART_NUMBER', 'Adaptative formating'],
+  [' ', 'Original value'],
```
ktmud, line 27: This has been updated to `"~g"`. See: https://github.com/apache/incubator-superset/pull/9608

packages/superset-ui-chart/src/controls/NVD3.ts
```diff
@@ -0,0 +1,299 @@
+/**
```
rusackas, line 1: I think this could be a part of @superset-ui/legacy-preset-chart-nvd3 since it's written for (and only used by) that one plugin. 

Same applies to the DeckGL equivalent - those shared controls could live with that plugin.

packages/superset-ui-chart/src/controls/sections.tsx
```diff
@@ -0,0 +1,143 @@
...
+  controlSetRows: [['granularity_sqla'], ['time_range']],
+};
+
+export const annotations = {
+  label: t('Annotations and Layers'),
+  expanded: true,
+  controlSetRows: [['annotation_layers']],
+};
+
+export const NVD3TimeSeries = [
```
rusackas, line 61: I wish this was called something other than NVD3... that makes me feel like it should live in the shared NVD3 controls module. In fact, it can, and may be should. This is used only by NVD3 plugins, but also used by the Horizon chart. If you were to manually copy the two objects (NVD3TimeSeries[0] and NVD3TimeSeries[1]) into that Horizon chart control file, then this stuff could be moved to the Shared NVD3 control panel

packages/superset-ui-chart/yarn.lock
```diff
@@ -0,0 +1,273 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
ktmud, line 2: We only use `yarn.lock` in the root directory. https://github.com/apache-superset/superset-ui/pull/362#discussion_r408291488

I'm creating a PR to add `yarn.lock` to `.gitignore`.

packages/superset-ui-chart/src/controls/NVD3.ts
suddjian: Oh thanks for the pointer, if it's only used there then that's way better.

packages/superset-ui-chart/yarn.lock
suddjian: oops, thanks

packages/superset-ui-chart/src/controls/NVD3.ts
kristw: Agree with @rusackas

kristw:
> Maybe create a new package superset-ui/controls

Agree with @ktmud.

packages/superset-ui-chart/package.json
```diff
@@ -27,6 +27,8 @@
     "access": "public"
   },
   "dependencies": {
+    "@superset-ui/translation": "^0.12.21",
```
kristw, line 30: These should be `peerDependencies`. (`translation` needs to be a singleton. `validator` not necessariliy, but I have been making all `@superset-ui/xxx` `peerDependencies` by convention to avoid having multiple copies of some singleton by mistake)

rusackas:
@kristw @ktmud I'm inclined to say this is OK as-is, and push it through for now. We should sync up about the plan for remaining controls and their dependencies. If they need to be in a new package (I'm not sure, but I'm open to it), we can address that in a separate PR. I would just want to call it something like `controlUtils` rather than `controls` since the latter sounds exactly like what we're all trying to kill off :)

ktmud:
> I would just want to call it something like `controlUtils` rather than `controls` since the latter sounds exactly like what we're all trying to kill off :)

@rusackas , can you clarify on this statement?

I was creating a new control type today (a [radio button](https://user-images.githubusercontent.com/335541/84557390-9f637000-acdf-11ea-8c62-14691115eeb6.png), to be specific) and thought it'd be beneficial to create it as a shared control.

A `@superset-ui/controls` or `@superset-ui/chart-controls` package seems a great place to store these shared control components, as well as those currently reside in [superset-frontend/src/explore/component/controls](https://github.com/apache/incubator-superset/blob/291306392443a5a0d0e2ee0cc4a95d37c56d4589/superset-frontend/src/explore/components/controls/index.js#L84). What are our plans regarding those?

PR #390, opened 2020-04-26
fix(superset-ui-query): add unit tests for #387

ktmud:
🏆 Enhancements

Add unit tests for #387, increase test coverage for core packages back to 100%.

cc @etr2460 @williaster @kristw @villebro

### DIFF
--- a/packages/superset-ui-query/test/buildQueryContext.test.ts
+++ b/packages/superset-ui-query/test/buildQueryContext.test.ts
@@ -1,6 +1,6 @@
 import { buildQueryContext } from '../src';
 
-describe('queryContextBuilder', () => {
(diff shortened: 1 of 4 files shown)


packages/superset-ui-query/test/buildQueryContext.test.ts
```diff
@@ -1,6 +1,6 @@
 import { buildQueryContext } from '../src';
 
-describe('queryContextBuilder', () => {
+describe('buildQueryContect', () => {
```
kristw, line 3: typo `Contect` => `Context`

etr2460:
Thanks for the unit tests! Sorry I didn't add them back when fixing this...

PR #391, opened 2020-04-27
fix: export superset theme props

nytai:
🏆 Enhancements
Export SupersetThemeProps interface so that it can be extended when using additional props in styled components.

### DIFF
--- a/packages/superset-ui-style/src/index.ts
+++ b/packages/superset-ui-style/src/index.ts
@@ -42,3 +42,7 @@ const defaultTheme = {
 export default styled as CreateStyled<typeof defaultTheme>;
 
 export const supersetTheme = defaultTheme;
+
+export interface SupersetThemeProps {
+  theme: typeof defaultTheme;
+}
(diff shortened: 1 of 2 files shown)


packages/superset-ui-style/src/index.ts
```diff
@@ -42,3 +42,7 @@ const defaultTheme = {
 export default styled as CreateStyled<typeof defaultTheme>;
 
 export const supersetTheme = defaultTheme;
+
+export interface SupersetThemeProps {
```
kristw, line 46: Not sure if this problem still persists with exporting types. We can try this first if you'd like, but if you run into issue later in the consuming app about "reexporting type", please come back here and move the definition of the interface to a new file such as `types.ts`

**types.ts**
```ts
export interface SupersetThemeProps {
 theme: typeof defaultTheme;
}
```

then export all from the type file like this in the **index.ts**

```ts
export default styled as CreateStyled<typeof defaultTheme>;
export const supersetTheme = defaultTheme;
export * from './types'
```

PR #398, opened 2020-04-28
chore: bump pretty-ms

villebro:
🐛 Bug Fix

Bump `pretty-ms` dependency to fix rounding inconsistencies/errors in version 6: https://github.com/sindresorhus/pretty-ms/pull/49


ktmud:
Better bump `yarn.lock` at the same time so that the actual upgrade doesn't happen in a random PR like this: https://github.com/apache-superset/superset-ui/pull/412/files#diff-8ee2343978836a779dc9f8d6b794c3b2

villebro:
Thanks @ktmud for the heads up, will keep this in mind next time.

PR #419, opened 2020-04-29
feat(legacy-plugin-chart-big-number): add control panel config for the BigNumber charts

suddjian:
💔 Breaking Changes

🏆 Enhancements

Allows BigNumber to register its own controls, so that the control configs can be removed from superset.

📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/big-number/src/BigNumber/index.ts
```diff
@@ -18,6 +18,7 @@
  */
 import { t } from '@superset-ui/translation';
 import { ChartMetadata, ChartPlugin } from '@superset-ui/chart';
+import controlPanel from './controls';
```
kristw, line 21: Let's name the file `controlPanel` so it is the same when imported.

plugins/legacy-preset-chart-nvd3/src/NVD3Controls.tsx
```diff
@@ -19,9 +19,12 @@
...
 import { t } from '@superset-ui/translation';
-import { controls } from '@superset-ui/chart';
-
-const { formatSelectOptions, D3_TIME_FORMAT_OPTIONS, D3_FORMAT_DOCS, D3_FORMAT_OPTIONS } = controls;
+import {
+  formatSelectOptions,
+  D3_TIME_FORMAT_OPTIONS,
+  D3_FORMAT_DOCS,
+  D3_FORMAT_OPTIONS,
+} from '@superset-ui/chart';
```
kristw, line 27: Please see #417. May have to change reference after that one is merged.

plugins/legacy-preset-chart-nvd3/src/NVD3Controls.tsx
```diff
@@ -338,6 +341,7 @@ export const timeSeriesSection = [
         'of query results',
     ),
     controlSetRows: [
+      // eslint-disable-next-line react/jsx-key
```
ktmud, line 344: Can move this to top of file to disable the rule for the whole file.

suddjian: I'm wary of disabling rules for a whole file, it's safer to do it per-line.

plugins/legacy-preset-chart-nvd3/src/NVD3Controls.tsx
suddjian: Let's merge that and I'll rebase

plugins/legacy-preset-chart-nvd3/src/NVD3Controls.tsx
ktmud: For this particular case, I don't understand why the rule needs to be disabled though. I see no real code change and it certainly has passed previous CI check?

kristw: previously it was in the `temporary-plugins` dir which is excluded from CI build and tests

suddjian:
Addressed PR feedback

plugins/big-number/package.json
```diff
@@ -37,6 +37,7 @@
   "peerDependencies": {
     "@superset-ui/chart": "^0.13.0",
     "@superset-ui/color": "^0.13.0",
+    "@superset-ui/control-utils": "^0.0.0",
```
kristw, line 40: This should be `^0.13.0` now.

PR #420, opened 2020-04-29
fix: lint errors

kristw:
🐛 Bug Fix

* Resolve type warnings from `eslint`
* Reduced from 274 errors to 198 errors


kristw:
@rusackas The one I'm disabling here broke when it is changed to `unknown`. I'm not sure how best to define the plain object type `{ [key: string]: any }` that won't break. If you have suggestion, please let me know.

rusackas:
> @rusackas The one I'm disabling here broke when it is changed to `unknown`. I'm not sure how best to define the plain object type `{ [key: string]: any }` that won't break. If you have suggestion, please let me know.

Ok, I figured you tried it, but thought I should ask :P I don't have any better suggestions, sadly.

PR #442, opened 2020-05-05
feat(control-utils): add infotooltipwithtrigger

pkdotson:
🏆 Enhancements

Adding infotooltipwithtrigger to superset ui. This is required for migrating partition chart control panels.


packages/superset-ui-control-utils/package.json
```diff
@@ -28,5 +28,10 @@
   "peerDependencies": {
     "@superset-ui/translation": "^0.13",
     "@superset-ui/validator": "^0.13"
+  },
+  "dependencies": {
+    "lodash": "^4.17.15",
+    "react": "^16.13.1",
+    "react-bootstrap": "^1.0.1"
```
ktmud, line 35: Re: dependencies

1. These should be `peerDependencies`, as they are shared by multiple `superset-ui` packages and should be upgraded all at once.
2. We are not using `react-bootstrap@1` yet. It's for Bootstrap v4, but we are still stuck with Boostrap v3. Check `incubator-superset` for property versions for `react-bootstrap` and `@types/react-bootstrap`.
3. Can we get rid of `lodash`? Is `kebabCase` absolutely necessary? Could you investigate why do we needed an ID for the tooltip in the first place? It's the best if we could just remove it. Less dependencies the better.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { kebabCase } from 'lodash';
+// @ts-ignore
+import { Tooltip, OverlayTrigger } from 'react-bootstrap';
+
+const defaultProps = {
+  className: 'text-muted',
+  icon: 'info-circle',
+  placement: 'right',
+};
+const tooltipStyle = { wordWrap: 'break-word' } as any;
```
ktmud, line 29: Try avoid `any` if possible:

```ts
const tooltipStyle: React.CSSProperties = { wordWrap: 'break-word' };
```

Since this rule is so simple, you can probably also just inline the style in `<Tooltip ... style={}>`.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React from 'react';
+import { kebabCase } from 'lodash';
+// @ts-ignore
+import { Tooltip, OverlayTrigger } from 'react-bootstrap';
```
ktmud, line 22: Prefer to remove `// @ts-ignore` and add proper typing support by installing `@types/react-bootstrap`.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const defaultProps = {
+  className: 'text-muted',
+  icon: 'info-circle',
+  placement: 'right',
+};
```
ktmud, line 28: Prefer to use ES6 function argument defaults.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+interface Props {
+  label: string;
+  tooltip: string;
+  icon: string;
+  onClick: () => void;
+  placement: string;
+  bsStyle: string;
+  className: string;
+}
```
ktmud, line 39: Make those with defaults optional.

```ts
interface Props {
  label: string;
  icon: string;
  tooltip?: string;
  onClick?: () => void;
  placement?: string;
  bsStyle?: string;
  className?: string;
}
```

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
suddjian: If I remember correctly, bootstrap 3 doesn't have types support. We ran into this problem in incubator-superset as well.

packages/superset-ui-control-utils/package.json
suddjian: Regarding 3, I think that kind of effort is better left until after we've moved things from superset. This should be as clean a refactor as possible, just moving code into superset-ui.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
kristw: having constant here is better so react won't re-render. when passing object to `style`, `react` only compares reference.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
kristw: Instead of @ts-ignore, you can add a declaration file. Even if it is an empty declaration

Example: 
https://github.com/apache-superset/superset-ui/blob/master/packages/superset-ui-translation/types/external.d.ts

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
kristw: +1 @ktmud with the small change to use `type` for `Props` instead of `interface`. (Just convention from guidelines at Airbnb and used throughout the codebase at the moment)

packages/superset-ui-control-utils/package.json
kristw: `react` in particular should always be `peerDependencies` in any library.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+export default function InfoTooltipWithTrigger({
```
kristw, line 41: I am actually not sure why this `react` component is in this package?

kristw:
Please also add unit test to ensure 100% coverage

packages/superset-ui-control-utils/package.json
suddjian: I also don't agree with making these peer dependencies, because the code directly depends on them. Peer dependencies aren't meant to be used for optimization, npm does that automatically when version numbers of dependencies are compatible with each other.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
ktmud: I think `@types/react-bootstrap @ 0.32.2` works with Bootstrap v3. There might be cases with suboptimal typings, where you can then do `@ts-ignore` per module.

packages/superset-ui-control-utils/package.json
kristw: I agree with making them `dependencies` as-is except `react`

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
suddjian: Decided to put it here because it's primarily used by controls, but open to moving it if you think a different package would be more appropriate. We're just trying to move forward on migrating the control panels into the chart plugins, and having this component accessible is a prerequisite for that.

packages/superset-ui-control-utils/package.json
ktmud: True, npm might be able to remove duplicates, but I think there is also the implication that what if it is not compatible? Are we really OK with having multiple versions of the dependent libraries in our app?

These `superset-ui` packages cannot work outside of Superset or Superset UI demos, so they should not introduce their own large non-optional dependencies. Using `peerDependencies` kind of forces package maintainers to keep the dependencies up to date.

kristw: > True, npm might be able to remove duplicates, but I think there is also the implication of what if it is not compatible? Are we really OK with having multiple versions of the dependent libraries in our app?

I think that is how all packages work tho. I am ok having multiple versions of the dependent instead of worrying about breaking anything if bump version of one dependencies.

> These superset-ui packages cannot work outside of Superset or Superset UI demos, so they should not introduce their own large non-optional dependencies. Using peerDependencies kind of forces package maintainers to keep the dependencies up to date.

These packages can work outside of Superset and Superset UI demos. They are independent packages on their own (number format, time format, color, etc.) which do not need Superset to function. I have them as dependencies in other projects and have always think of them independent from `incubator-superset`. They should have their own dependencies.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
kristw: > it's primarily used by controls

Are they referrred to in the `controlPanel.ts`? Just asking

packages/superset-ui-control-utils/package.json
suddjian: All other superset-ui packages have things listed under `dependencies`, including lodash. Is this different somehow?

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
suddjian: Nice! That means we can add types for bootstrap back in incubator-superset too

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
suddjian: Yes, that's where the need to move it came from

packages/superset-ui-control-utils/package.json
kristw: I am proposing we proceed with moving `react` to `peerDependencies` and leave the remaining in `dependencies` unless there is objection?

suddjian: No objection here. I actually already moved react, but haven't pushed it yet.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
suddjian: @ktmud I tried adding the types for bootstrap, but that leads to an issue in CI. It looks like our version of typescript isn't happy with those type definitions, any ideas on how to fix?

ktmud: Sorry, should've used `@types/react-bootstrap @ ^0.32.21`. This is also what's used in `incubator-superset`.

Btw, it's preferable to have `@types/*` in `dependencies` instead of `devDependencies` so that dependent packages can have easier access to typings when the library is used in `npm link`.

packages/superset-ui-control-utils/src/InfoTooltipWithTrigger.tsx
ktmud: `interface` means the properties are immutable which often runs into problems when you want to extend the component and override the prop types.

suddjian:
Addressed PR feedback, should be ready for a re-review

PR #443, opened 2020-05-06
chore: pulling in Sunburst controls from incubator-superset

rusackas:
💔 Breaking Changes

🏆 Enhancements

This pulls in the controls from `incubator-superset`, and bumps the version number of the package. [Another PR](https://github.com/apache/incubator-superset/pull/9746) has been filed on `incubator-superset` to pull in this new package and remove the old controls and shims.

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/plugins/legacy-plugin-chart-sunburst/package.json
+++ b/plugins/legacy-plugin-chart-sunburst/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/legacy-plugin-chart-sunburst",
-  "version": "0.13.6",
+  "version": "0.13.5",
   "description": "Superset Legacy Chart - Sunburst",
   "sideEffects": [
     "*.css"
--- a/plugins/legacy-plugin-chart-sunburst/package.json
+++ b/plugins/legacy-plugin-chart-sunburst/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/legacy-plugin-chart-sunburst",
-  "version": "0.13.5",
+  "version": "0.13.6",
   "description": "Superset Legacy Chart - Sunburst",
   "sideEffects": [
     "*.css"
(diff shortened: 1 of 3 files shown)


plugins/legacy-plugin-chart-sunburst/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@superset-ui/legacy-plugin-chart-sunburst",
-  "version": "0.13.5",
+  "version": "0.13.6",
```
rusackas, line 3: @kristw maybe I shouldn't be doing this manually, if that CI job does it for me?

rusackas: Actually, perhaps this should actually be a minor version bump, not a patch bump? ¯\_(ツ)_/¯

kristw: The version 0 is a bit weird since it is `0.major.minor` while the lerna command consider it a patch

kristw: You should not bump the version manually. The `ci:create-patch-version` script should do it.

plugins/legacy-plugin-chart-sunburst/src/index.js
```diff
@@ -20,6 +20,7 @@ import { t } from '@superset-ui/translation';
 import { ChartMetadata, ChartPlugin } from '@superset-ui/chart';
 import transformProps from './transformProps';
 import thumbnail from './images/thumbnail.png';
+import controlPanel from './controlPanel.ts';
```
kristw, line 23: please remove `.ts` This may break in consumer app because it may refer to the `.js` transpiled file.

rusackas: Hmm... originally I didn't have the .ts there, but CI was failing due to a linting error about it.

plugins/legacy-plugin-chart-sunburst/package.json
rusackas: Un-bumped :)

plugins/legacy-plugin-chart-sunburst/src/index.js
rusackas: removed, and eslint-disable'd

kristw: hmm. we should consider changing that to warning or sth. thanks for the note.

PR #446, opened 2020-05-06
chore: bumping package versions

rusackas:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal
Just bumping packages with the lerna script 🤞


rusackas:
Apparently this PR should not be necessary and the script is magic when wielded correctly.

PR #447, opened 2020-05-07
fix: lint again

kristw:
🏠 Internal

Fix another hundred or so of lint warnings. Down from 199 warnings to < 10.

### DIFF
--- a/packages/superset-ui-chart/src/models/ChartPlugin.ts
+++ b/packages/superset-ui-chart/src/models/ChartPlugin.ts
@@ -1,4 +1,4 @@
-import { FunctionComponent, ComponentType } from 'react';
+import { ComponentType } from 'react';
 import { isRequired, Plugin } from '@superset-ui/core';
 import { QueryFormData } from '@superset-ui/query';
 import ChartMetadata from './ChartMetadata';
@@ -18,7 +18,7 @@ const EMPTY = {};
 
 export type PromiseOrValue<T> = Promise<T> | T;
 export type PromiseOrValueLoader<T> = () => PromiseOrValue<T>;
-export type ChartType = ComponentType<any> | FunctionComponent<any>;
+export type ChartType = ComponentType;
 type ValueOrModuleWithValue<T> = T | { default: T };
 
 interface ChartPluginConfig<T extends QueryFormData> {
--- a/packages/superset-ui-chart/src/models/ChartPlugin.ts
+++ b/packages/superset-ui-chart/src/models/ChartPlugin.ts
@@ -18,7 +18,7 @@ const EMPTY = {};
 
 export type PromiseOrValue<T> = Promise<T> | T;
 export type PromiseOrValueLoader<T> = () => PromiseOrValue<T>;
-export type ChartType = ComponentType;
+export type ChartType = ComponentType<any>;
 type ValueOrModuleWithValue<T> = T | { default: T };
 
 interface ChartPluginConfig<T extends QueryFormData> {
--- a/packages/superset-ui-connection/test/SupersetClient.test.ts
+++ b/packages/superset-ui-connection/test/SupersetClient.test.ts
@@ -13,14 +13,14 @@ describe('SupersetClient', () => {
   afterEach(SupersetClient.reset);
 
   it('exposes reset, configure, init, get, post, isAuthenticated, and reAuthenticate methods', () => {
-    expect(SupersetClient.configure).toEqual(expect.any(Function));
-    expect(SupersetClient.init).toEqual(expect.any(Function));
-    expect(SupersetClient.get).toEqual(expect.any(Function));
-    expect(SupersetClient.post).toEqual(expect.any(Function));
-    expect(SupersetClient.isAuthenticated).toEqual(expect.any(Function));
-    expect(SupersetClient.reAuthenticate).toEqual(expect.any(Function));
-    expect(SupersetClient.request).toEqual(expect.any(Function));
-    expect(SupersetClient.reset).toEqual(expect.any(Function));
+    expect(typeof SupersetClient.configure).toBe('function');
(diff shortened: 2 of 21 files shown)


packages/superset-ui-chart/src/models/ChartPlugin.ts
```diff
@@ -18,7 +18,7 @@ const EMPTY = {};
 
 export type PromiseOrValue<T> = Promise<T> | T;
 export type PromiseOrValueLoader<T> = () => PromiseOrValue<T>;
-export type ChartType = ComponentType<any> | FunctionComponent<any>;
+export type ChartType = ComponentType<any>;
```
kristw, line 21: `ComponentType` already includes `FunctionComponent`

packages/superset-ui-connection/test/callApi/callApi.test.ts
```diff
@@ -8,6 +8,10 @@ import throwIfCalled from '../utils/throwIfCalled';
 import { CallApi } from '../../src/types';
 import { DEFAULT_FETCH_RETRY_OPTIONS } from '../../src/constants';
 
+function getObjectKeys<T>(object: T) {
+  return Object.keys(object) as (keyof T)[];
+}
```
ktmud, line 13: If this function is just for typing, maybe it makes sense to just inline it:

```ts
        (Object.keys(object) as (keyof T)[]).forEach(key => {
          expect(body.get(key)).toBe(JSON.stringify(postPayload[key]));
        });
```

So future readers don't have to wonder what's the purpose of this function.

Another option is to use `Object.entries`:

```ts
        Object.entries(postPayload).forEach(([key, value]) => {
          expect(body.get(key)).toBe(JSON.stringify(value));
        });
```

plugins/preset-chart-xy/src/BoxPlot/legacy/transformProps.ts
```diff
@@ -22,7 +22,9 @@ import { RawBoxPlotDataRow, BoxPlotDataRow } from '../../components/BoxPlot/type
 export default function transformProps(chartProps: ChartProps) {
   const { width, height, datasource, formData, queryData } = chartProps;
   const { verboseMap = {} } = datasource;
-  const { colorScheme, groupby, metrics } = formData;
+  const { colorScheme } = formData;
+  const groupby = formData.groupby as string[];
+  const metrics = formData.metrics as string[];
```
ktmud, line 27: Maybe expand the `ChartProps` like [BigNumber](https://github.com/apache-superset/superset-ui/blob/20e70244548f167e206a6f822569305321513ec5/plugins/legacy-preset-chart-big-number/src/BigNumber/transformProps.ts#L41-L62) for consistency?

I'm wondering whether we can add these generic `formData` properties to the global `formData`, too. They would go hand-in-hand with the shared controls.

packages/superset-ui-connection/test/callApi/callApi.test.ts
kristw: Using the function can infer `T`. Otherwise it has to be

```ts
(Object.keys(object) as (keyof typeof object)[]).forEach(...)
```

Not sure if having casting in every occurrence like this is better or having a helper is better. I had to cast the return type of `Object.keys()` in many places already. Now considering adding this function as part of the `core`.

`Object.entries()` has the same issue with `Object.keys()` that `key` is `string`

ktmud: `Object.entries` at least works for this case because `value` would not be `any` but rather the union of all value types in `object`.

ktmud:
Btw, why does `superset-ui-connection` has it's own `eslintrc`?

kristw:
> Btw, why does `superset-ui-connection` has it's own `eslintrc`?

Good question. I am not aware of this either. cc @williaster

packages/superset-ui-connection/test/SupersetClient.test.ts
```diff
@@ -13,14 +13,14 @@ describe('SupersetClient', () => {
...
-    expect(SupersetClient.request).toEqual(expect.any(Function));
-    expect(SupersetClient.reset).toEqual(expect.any(Function));
+    expect(typeof SupersetClient.configure).toBe('function');
+    expect(typeof SupersetClient.init).toBe('function');
+    expect(typeof SupersetClient.get).toBe('function');
+    expect(typeof SupersetClient.post).toBe('function');
+    expect(typeof SupersetClient.isAuthenticated).toBe('function');
+    expect(typeof SupersetClient.reAuthenticate).toBe('function');
+    expect(typeof SupersetClient.request).toBe('function');
+    expect(typeof SupersetClient.reset).toBe('function');
```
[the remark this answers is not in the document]
kristw: not sure. I only do the minimal to remove lint warning.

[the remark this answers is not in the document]
kristw: lemme check

packages/superset-ui-connection/test/callApi/callApi.test.ts
kristw: let me give this a try.

kristw: converted to `Object.entries`

packages/superset-ui-connection/test/SupersetClient.test.ts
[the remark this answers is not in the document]
kristw: This file may benefit from more refactor. Will leave it like this for this PR.

plugins/preset-chart-xy/src/BoxPlot/legacy/transformProps.ts
kristw: Great idea.

PR #455, opened 2020-05-08
feat(legacy-preset-big-number): use emotion styling

rusackas:
💔 Breaking Changes

🏆 Enhancements
Adding emotion styling, pulling from Superset theme, particularly for typographic purposes in Big Number plugin. This also removes the CSS file, and puts ALL the styling INSIDE the component where it's exported. Emotion includes SASS out of the box, so I've condensed things a bit accordingly.

This also adds a ThemeProvider to the storybook, so the plugins work/look right there.

Before screenshots:
Screen Shot 2020-05-20 at 3 11 46 PM
Screen Shot 2020-05-20 at 3 11 56 PM

After screenshots:
Screen Shot 2020-05-20 at 3 09 13 PM
Screen Shot 2020-05-20 at 3 12 17 PM

📜 Documentation

🐛 Bug Fix

🏠 Internal


rusackas:
Font change is intentional, bringing it inline with the rest of Incubator-Superset's UI. Other viz pugins will follow suit, changing the fonts for axes, tooltips, whatever. Eventually we'll do the same for colors, border radii, and all the other little bits and baubles that come from the theme. This will be a long tail, but this PR show the pattern.

PR #456, opened 2020-05-08
feat(generator-superset): add scaffolder for chart plugin

kristw:
🏆 Enhancements

Add `yeoman` generator for chart plugin

image


packages/generator-superset/generators/plugin-chart/templates/src/MyChart.erb
```diff
@@ -0,0 +1,49 @@
...
+            cy={y * height}
+            r={Math.random() * 20}
+            fill="#fff"
+            opacity="0.5"
+          />
+        ))}
+        <text x={width / 2} y={height / 2} textAnchor="middle" fontWeight="bold" fontSize="36">
+          Hello!
+        </text>
+      </svg>
```
ktmud, line 46: Could this be simplified? People would need to delete it anyway, right?

kristw: sure. I was copy pasting from the dummy chart. 

https://apache-superset.github.io/superset-ui-plugins-template/?path=/story/plugin-chart-dummy-dummychartplugin--basic

PR #459, opened 2020-05-08
feat(control-utils): add shared controls + dependencies, convert to typescript

suddjian:
🏆 Enhancements

Brought in the remaining shared controls from incubator-superset. This is a prerequisite for moving many of the controlPanels into the chart plugins.

When typing the `shared-controls` module, I did my best but had to use `unknown` in several places. I propose merging the types as-is, and putting together a set of properly fleshed-out types for the controls system separately, since that will be very useful for plugin writers/maintainers. I don't feel well-equipped to do that myself.

There are many subtle changes to these files due to differences in typescript/linter requirements between projects. Looking at the diff between these files and their originals in incubator-superset might be the best way to review.

I know this brings test coverage down, but I don't have the bandwidth to write tests for everything in shared-controls.tsx. I don't think adding tests would actually provide any real quality improvements anyway, since those functions are so closely tied to external behavior and have almost no behavior of their own.

### DIFF
--- a/packages/superset-ui-control-utils/package.json
+++ b/packages/superset-ui-control-utils/package.json
@@ -34,6 +34,7 @@
     "enzyme": "^3.11.0"
   },
   "dependencies": {
+    "@superset-ui/color": "^0.13.3",
     "@types/react-bootstrap": "0.32.21",
     "lodash": "^4.17.15",
     "react-bootstrap": "^0.33.1"
--- a/packages/superset-ui-control-utils/package.json
+++ b/packages/superset-ui-control-utils/package.json
@@ -26,6 +26,7 @@
     "access": "public"
   },
   "peerDependencies": {
+    "@superset-ui/color": "^0.13.3",
     "@superset-ui/translation": "^0.13",
     "@superset-ui/validator": "^0.13",
     "react": "^16.13.1"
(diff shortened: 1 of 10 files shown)


suddjian:
This should be the last piece necessary to add to `control-utils` in order to allow moving the rest of the controlPanels over.

(except the DeckGL charts, they have some other dependency stuff going on that will need to be handled separately)

ktmud:
I'm making changes to the code coverage check: https://github.com/apache-superset/superset-ui/pull/468

packages/superset-ui-control-utils/package.json
```diff
@@ -34,6 +34,7 @@
     "enzyme": "^3.11.0"
   },
   "dependencies": {
+    "@superset-ui/color": "^0.13.3",
```
kristw, line 37: Please move this to `peerDependencies`

kristw:
Please rebase from master and that should resolve the coverage

suddjian:
Rebased and should be ready for merge pending CI

suddjian:
This is still failing coverage after rebasing. I'd like to just merge it since tests are passing and we've agreed to forego coverage for these files.

ktmud:
Damn, the coverage rule still does not work, as `.ts` now becomes partial match that also matches `.tsx`.

ktmud:
Merged and created a PR to fix the coverage in the meantime: #482

PR #463, opened 2020-05-08
feat(style): adding typographic variables to theme

rusackas:
💔 Breaking Changes

🏆 Enhancements
Adding a few styles to the superset theme object, just to fiddle around with in other packages

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-style/src/index.ts
+++ b/packages/superset-ui-style/src/index.ts
@@ -36,6 +36,27 @@ const defaultTheme = {
       light5: '#F5F5F8',
     },
   },
+  typography: {
+    families: {
+      sansSerif: `'Inter', Helvetica, Arial`,
+      serif: `Georgia, 'Times New Roman', Times, serif`,
+      monospace: `'Fira Code', 'Courier New', monospace`,
+    },
+    weights: {
+      light: 200,
+      normal: 400,
+      bold: 700,
+    },
+    sizes: {
+      xxs: '9px',
+      xs: '10px',
+      s: '12px',
+      m: '14px',
+      l: '16px',
+      xl: '21px',
+      xxl: '28px',
+    },
+  },
   gridUnit: '4px',
 };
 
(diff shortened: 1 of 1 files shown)


packages/superset-ui-style/src/index.ts
```diff
@@ -36,6 +36,27 @@ const defaultTheme = {
...
+      xxs: '9px',
+      xs: '10px',
+      s: '12px',
+      m: '14px',
+      l: '16px',
+      xl: '21px',
+      xxl: '28px',
+    },
+  },
   gridUnit: '4px',
```
ktmud, line 60: Can we make these numeric so possible future computation can be easier? I think it makes sense to make base units always be pixels.

ktmud: For example, the new `react-select` uses baseUnit arithmetics directly in custom styles:

https://github.com/apache/incubator-superset/pull/9628/files#diff-a6194bf15854838f02e26b41cac7cdebR217-R244

rusackas: I've gone back and forth on this, and so have others in the discussions. The numeric bits do make things simpler in one sense, but being explicit about units in the theme also opens you up to having heterogeneous units, e.g. some things in `rem`s or viewport units, or percentages, or whatever, without having to interpret them when looking at the theme file.

rusackas: But yes, I'll make the change... this is really just to support a little POC work... we can adjust as this thing grows.

rusackas: I guess the other downside is that we'd have to tack on the `px` in exponentially more places in the code (every instance of the style bing used), where it's centralized here. But we can try it out this way and see how it goes.

@ktmud Change is committed, if you don't mind approving, so I can work on the new POC PR.

ktmud: I’m in favor of using relative units whenever possible, but baseUnits should be pixels as the computation needs to start from somewhere. Which are categorized as baseUnits is debatable, but I’m imagine most of those in the core theme definition should be.

Another choice is to use CSS variables. Facebook just did that in their new website: https://engineering.fb.com/web/facebook-redesign/

PR #468, opened 2020-05-08
test: be more lenient on jsx/tsx test coverage

ktmud:
React modules don't have to have 100% unit-test coverages.

🏆 Enhancements

🏠 Internal

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,27 +1,21 @@
 coverage:
   status:
-    patch:
-      default:
-        target: 0%
-        threshold: 0%
-      core-packages:
-        target: 0%
-        paths:
-          - packages
-      plugins:
-        target: 0%
-        paths:
-          - plugins
+    patch: off
     project:
       default:
         target: 0%
-        # send notification if coverage change exceed this threshold
-        threshold: 0%
-      core-packages:
+        threshold: 10%
+      core-packages-ts:
         target: 100%
         paths:
-          - packages
+          - packages/**/*.js
+          - packages/**/*.ts
+      core-packages-tsx:
+        target: 50%
+        paths:
+          - packages/**/*.jsx
+          - packages/**/*.tsx
       plugins:
-        target: 0%
+        target: 0
         paths:
           - plugins


ktmud:
Kind of feel maybe we should just get rid of the `patch` check?

codecov.yml
```diff
@@ -1,26 +1,20 @@
...
+      core-packages-ts:
         target: 100%
         paths:
-          - packages
+          - packages/*.ts$
+          - packages/*.js$
+      core-packages-tsx:
+        target: 10%
+        paths:
+          - packages/*.jsx$
```
kristw, line 16: would something like this work? 

```
- packages/!(superset-ui-control-utils)/**/*.{j,t}sx
```

https://www.digitalocean.com/community/tools/glob?comments=true&glob=packages%2F%21%28superset-ui-control-utils%29%2F%2A%2A%2F%2A.%7Bj%2Ct%7Dsx&matches=false&tests=%2F%2F%20This%20will%20match%20as%20it%20ends%20with%20%27.js%27&tests=packages%2Fsuperset-ui-control-utils%2Fworld.js&tests=packages%2Fsuperset-ui-control-utils%2Fworld.jsx&tests=packages%2Fsuperset-ui-control-utils%2Fworld.ts&tests=packages%2Fsuperset-ui-control-utils%2Fworld.tsx&tests=packages%2Fsuperset-ui-chart%2Fchart.tsx&tests=packages%2Fsuperset-ui-chart%2Fplugin%2Fchart.tsx&tests=%2F%2F%20This%20won%27t%20match%21&tests=%2Ftest%2Fsome%2Fglobs

codecov.yml
```diff
@@ -1,26 +1,20 @@
...
         target: 0%
-        # send notification if coverage change exceed this threshold
-        threshold: 0%
-      core-packages:
+        threshold: 10%
+      core-packages-ts:
         target: 100%
         paths:
-          - packages
+          - packages/*.ts$
```
kristw, line 11: > The pattern folder/* will not match recursively in the folder.
> Please use this folder/**/*.

https://docs.codecov.io/docs/ignoring-paths

ktmud: yeah, I was referring to https://docs.codecov.io/docs/path-regexp and thought it’d work.

codecov.yml
ktmud: I thought codecovv glob works differently. Their documentation was not super clear.

kristw: Yeah I'm always confused about the glob pattern. Let's give it a few iterations.

codecov.yml
```diff
@@ -1,26 +1,18 @@
...
-        threshold: 0%
-      core-packages:
+        threshold: 10%
+      core-packages-ts:
         target: 100%
         paths:
-          - packages
+          - packages/**/*.{js|ts}$
+      core-packages-tsx:
+        target: 10%
```
kristw, line 13: `100%` ?

Could we combine the `core-packages-ts` and `core-packages-tsx` section then add the `ignore` path for `control-utils`'s `{t|j}sx`?

rusackas: Probably not 100% since relaxing this constraint on `{j|t}sx` is the point. That said, the number can be negotiated/adjusted at any point. 50%? ¯\\\_(ツ)_/¯

codecov.yml
ktmud: Looks like `{j,t}` operator doesn't really work for codecov.

kristw: does `@(j|t)` work?

kristw:
Not sure why

image

codecov.yml
ktmud: That looks even more advanced... I doubt codecov will support that.

ktmud:
> Not sure why
> 
> image

Apparently you can't mix the use of "**" and "$". So `projects/**/*.ts$` doesn't work, but `projects/**/*.ts` does.

rusackas:
@kristw @ktmud Seems like this is good to go, right? My trigger finger is itchy here, as this will unblock [this PR](https://github.com/apache-superset/superset-ui/pull/459).

PR #469, opened 2020-05-08
feat(legacy-plugin-chart-nvd3): add control panels

suddjian:
All the ones that are actually used in superset, anyway

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/src/Area/controlPanel.ts
+++ b/plugins/legacy-preset-chart-nvd3/src/Area/controlPanel.ts
@@ -1,0 +1,101 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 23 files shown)


plugins/legacy-preset-chart-nvd3/src/Area/controlPanel.ts
```diff
@@ -0,0 +1,101 @@
...
+      expanded: true,
+      controlSetRows: [
+        ['y_axis_format', yAxisBounds],
+        [yLogScale, null],
+      ],
+    },
+    timeSeriesSection[1],
+    sections.annotations,
+  ],
+  controlOverrides: {
```
kristw, line 88: Could we apply these overrides in-place instead of having this section here?

```
['color_scheme', 'label_colors']
```

```
[
  'color_scheme', // replace with the object and set renderTrigger to false
  'label_colors'
]
```

suddjian: I think that's a good idea, but for now I'm just trying to move the control panels while keeping them as intact as possible to minimize the chance of breaking things.

rusackas: @kristw I'm making an internal JIRA ticket to track this, and we can go through and audit and remove (where possible) all `controlOverrides`. If you think it's safe enough, we can then consider removing the code that enables them, as well. We'll discuss further with you :)

PR #363, opened 2020-04-14
build: disable commit lint

ktmud:
🏆 Enhancements
🏠 Internal

Disable commit lint in local development for smoother developer experience.

All changes must go through PR process and we already have [Semantic Pull Request](https://github.com/zeke/semantic-pull-requests) check. We almost always do squash merge, which means what developers committed locally doesn't really matter anyway.

PR #471, opened 2020-05-08
chore: enforce commit-lint on first commit in branch

ktmud:
🏆 Enhancements

🏠 Internal

Since disabled commit-lint in #363, people who didn't install [Refined GitHub ](https://github.com/sindresorhus/refined-github) had been merging PRs without conventional commits.

This re-applies commit-lint to the first commit in branches that are not `master`, so to make sure we will always have conventional commits in master.

Also fixes prettier check in `pre-commit`, as it was unnecessarily slowed down by checking all files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "vx"
   ],
   "license": "Apache-2.0",
-  "devDependencies": {
+  "dependencies": {
     "@airbnb/config-babel": "^3.1.0",
     "@airbnb/config-eslint": "^3.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "vx"
   ],
   "license": "Apache-2.0",
-  "devDependencies": {
+  "dependencies": {
```
ktmud, line 54: The root package.json doesn't have a production release anyway. This change is needed to appease linting in `scripts`.

PR #479, opened 2020-05-12
feat(legacy-plugin-chart-markup): add controls to markup chart

pkdotson:
💔 Breaking Changes
   
🏆 Enhancements
     Add controls to markup chart plugin.
📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/plugins/legacy-plugin-chart-markup/src/controlPanel.js
+++ b/plugins/legacy-plugin-chart-markup/src/controlPanel.js
@@ -20,7 +20,6 @@ import { t } from '@superset-ui/translation';
 import { validateNonEmpty } from '@superset-ui/validator';
 import { formatSelectOptions } from '@superset-ui/control-utils';
 
-console.log('what are going on with these controls')
 export default {
   controlPanelSections: [
     {
--- a/plugins/legacy-plugin-chart-markup/src/controlPanel.js
+++ b/plugins/legacy-plugin-chart-markup/src/controlPanel.js
@@ -1,0 +1,72 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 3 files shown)


plugins/legacy-plugin-chart-markup/src/controlPanel.js
```diff
@@ -0,0 +1,72 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { t } from '@superset-ui/translation';
+import { validateNonEmpty } from '@superset-ui/validator';
+import { formatSelectOptions } from '@superset-ui/control-utils';
+
+console.log('what are going on with these controls')
```
kristw, line 23: remove `console.log`

plugins/legacy-plugin-chart-markup/src/controlPanel.js
```diff
@@ -0,0 +1,72 @@
...
+                    : 'markdown',
+              }),
+              default: '',
+            },
+          },
+        ],
+      ],
+    },
+  ],
+  sectionOverrides: {
```
kristw, line 64: Is this section needed?

rusackas: @kristw I think they are needed (or at least relevant). There are a few plugins that do this... iframe, markup, and separator. It appears this basically removes the `time_range` control  that's seemingly injected automatically in some cases as part of the `druidTimeSeries`. Maybe we can find a better way of addressing this in the future, but it seems out of the scope of this PR, which is just to migrate the code that's there already.

kristw: got it

PR #485, opened 2020-05-13
feat: add control grouping functionality

villebro:
🏆 Enhancements
This adds support for control value grouping introduced in https://github.com/apache/incubator-superset/pull/9740 . If a control has specified the property `queryFields`, those are used to group control values together when the query is build. Example `formData`:

```json
{
  "series": ["gender", "nationality"],
  "primary_metric": "age",
  "secondary_metric": "height",
  "queryFields": {
    "primary_metric": "metrics",
    "secondary_metric": "metrics",
    "series': 'groupby"
  }
}
```

This will result in the following query object:

```json
{
  "metrics": ["primary_metric", "secondary_metric"],
  "groupby": ["gender", "nationality']
}
```

 This will help decouple `superset-ui` from individual plugins, as plugins can specify an arbitrary number of `metric`, `secondary_metric`, `series`, `groupby` without there being a need for `superset-ui` to know which is which. This will also bring the query logic closer to the control panel where controls are defined, as the `queryField` can be defined in the control metadata, hence won't need to be mapped over in `buildQuery.{js,ts}`.


packages/superset-ui-query/src/buildGroupedControls.ts
```diff
@@ -0,0 +1,38 @@
+import { ControlGroups, ResidualFormData } from './types/QueryFormData';
+import { GroupedControlData, ResidualGroupedControlData } from './types/ControlData';
+
+export default function buildGroupedControlData(
```
kristw, line 4: no need to specify return type if can be inferred from typescript

packages/superset-ui-query/src/buildQueryObject.ts
```diff
@@ -24,23 +25,25 @@ export default function buildQueryObject<T extends QueryFormData>(formData: T):
...
     order_desc,
     row_limit,
     limit,
     timeseries_limit_metric,
+    controlGroups,
+    ...residualFormData
   } = formData;
 
+  const groupedControls = buildGroupedControlData(residualFormData, controlGroups);
+  const { metrics, groupby, columns } = groupedControls;
```
kristw, line 37: nit

```ts
const { metrics, groupby, columns } = buildGroupedControlData(residualFormData, controlGroups);
```

packages/superset-ui-query/src/processMetrics.ts
```diff
@@ -1,25 +1,11 @@
...
+import { QueryFormDataMetric } from './types/QueryFormData';
 import { QueryObjectMetric } from './types/Query';
-import { MetricKey } from './types/Metric';
 import convertMetric from './convertMetric';
 
-export default function processMetrics(formData: QueryFormData) {
-  // Use Array to maintain insertion order
-  // for metrics that are order sensitive
+export default function processMetrics(rawMetrics: QueryFormDataMetric[]) {
   const metrics: QueryObjectMetric[] = [];
```
kristw, line 6:
```ts
return rawMetrics.map(convertMetric);
```

villebro:
Thanks for the review @kristw ; my apologies for the brief description, I meant to expand on the initial description but got sidetracked. Will add more context to describe why this change is necessary and pointers for some design decisions. Also, as this is my first real stab at TS, I fully expect there to be inconsistencies, so if something looks weird, it's probably because I've done it wrong.

ktmud:
I think the term "control group" is kind of confusing. It took me quite a while to understand what problem this PR tries to solve.

If I'm getting it right, this is about consolidating control fields required for constructing DB queries (`metrics`, `groupby`, and `columns`). They may have as arbitrary field names, but the backend needs to know which are `metrics` fields, which are `groupby`, etc.

It's not _grouped control_ as in UI components, but more like **grouped query API parameters** or **consolidated query fields**. The term _control group_ is also too commonly used in Data Science in the context of hypothesis testing, i.e., A/B tests.

Can we maybe rename the variables/functions as such:

```
controlGroups -> queryFields
buildGroupedControls -> extractQueryFields
defaultedControlGroups -> aliases or queryFieldAliases
```

packages/superset-ui-query/src/processMetrics.ts
suddjian: need flatmap

suddjian: oh wait I was looking at the wrong side of the diff 🤦‍♂️

packages/superset-ui-query/src/types/Query.ts
```diff
@@ -40,6 +40,11 @@ export type QueryObjectExtras = Partial<{
   where?: string;
 }>;
 
+export type ResidualQueryObjectData = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
suddjian, line 45: `unknown` might work better than `any` here. https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

packages/superset-ui-query/src/types/QueryFormData.ts
```diff
@@ -1,18 +1,17 @@
...
-// Define mapped type separately to work around a limitation of TypeScript
-// https://github.com/Microsoft/TypeScript/issues/13573
-// The Metrics in formData is either a string or a proper metric. It will be
-// unified into a proper Metric type during buildQuery (see `/query/Metrics.ts`).
-export type QueryFormDataMetrics = Partial<
-  Record<MetricKey, QueryFormDataMetric | QueryFormDataMetric[]>
->;
+export type ResidualFormData = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  [key: string]: any;
```
suddjian, line 10: ditto on `unknown`

packages/superset-ui-query/src/buildGroupedControls.ts
```diff
@@ -0,0 +1,38 @@
...
+    columns: [],
+    groupby: [],
+    metrics: [],
+  };
+
+  Object.entries(residualFormData).forEach(entry => {
+    const [key, residualValue] = entry;
+    if (Object.prototype.hasOwnProperty.call(defaultedControlGroups, key)) {
+      const controlGroup: string = defaultedControlGroups[key];
+      const controlValue = residualFormData[key];
```
suddjian, line 29: I'm a little confused by what this function is doing. Aren't `controlValue` and `residualValue` the same thing?

packages/superset-ui-query/src/types/QueryFormData.ts
suddjian: Actually, are the items in ResidualFormData a `QueryFormDataMetric`? `buildGroupedControlData` seems to imply that.

packages/superset-ui-query/src/buildGroupedControls.ts
suddjian: Here's what I _think_ this function is essentially doing, correct me if I'm wrong?

```
const normalizedKey = Object.prototype.hasOwnProperty.call(defaultedControlGroups, key)
      ? defaultedControlGroups[key]
      : key;
 groupedControls[normalizedKey] = (groupedControls[normalizedKey] || []).concat(residualValue);
```

villebro:
Thanks all for the comments! ❤️ Will implement changes and update description + comment where necessary.

packages/superset-ui-query/src/buildGroupedControls.ts
rusackas: I, too, was a bit puzzled here, so I've tried to write a simplified version of this block as well. Curious if this is missing something important (I'm a little tired, so... probably):
```
Object.keys(residualFormData).forEach(key => {
  const conformedKey = defaultedControlGroups[key] || key;
  groupedControls[conformedKey] = [...groupedControls[conformedKey] || [], residualFormData[key]]
});
```

packages/superset-ui-query/src/buildQueryObject.ts
rusackas: Is anything from the object returned by `buildGroupedControlData` expected to be useful besides `metrics, groupby, columns`? If not, it may be possible to simplify that method's logic to ignore the extraneous stuff.

packages/superset-ui-query/src/buildGroupedControls.ts
villebro: You're both right, this was poorly written 🤦 Will address.

packages/superset-ui-query/src/types/QueryFormData.ts
villebro: Yes, I was also puzzled by this. `ResidualFormData` (in retrospect should be `QueryFormResidualData`) is currently expected to contain either column names (`string`) or metrics (`QueryFormDataMetric`  which is the same as `string | AdhocMetric`). When these are combined, the union is reduced to `string | AdhocMetric` which is the same as `QueryFormDataMetric`. I'm tempted to remove `QueryFormDataMetric` and just pass around `QueryFormResidualData`, which can later be split into `QueryFormDataMetric` and `QueryFormDataColumn` when that need arises (Down the line we'll need to add `AdhocColumn` which is a column with a SQL expression). I will actually do that, we'll see how it looks.

PR #486, opened 2020-05-13
feat(time-format): add full-date to weekly time formatter

ktmud:
🏆 Enhancements

BigNumber trendline uses time formatter based on granularity. Currently the formatter weekly periods is not very helpful. We add full date to the formatted string so users don't have to compute which month/day the week is on their own.

## Before

image

## After

image


packages/superset-ui-time-format/src/factories/getTimeFormatterForGranularity.ts
```diff
@@ -39,16 +41,30 @@ const formats = {
...
   P1Y: '%Y', // year
-  // d3-time-format weeks does not support weeks start on Sunday
   '1969-12-28T00:00:00Z/P1W': SUNDAY_BASED_WEEK, // 'week_start_sunday'
   '1969-12-29T00:00:00Z/P1W': MONDAY_BASED_WEEK, // 'week_start_monday'
   'P1W/1970-01-03T00:00:00Z': SUNDAY_BASED_WEEK, // 'week_ending_saturday'
   'P1W/1970-01-04T00:00:00Z': MONDAY_BASED_WEEK, // 'week_ending_sunday'
 };
 
-export type TimeGranularity = keyof typeof formats;
+export type TimeGranularity =
```
kristw, line 50: There is a known issue with re-exporting type for library. 

```
export {
  default as getTimeFormatterForGranularity,
  TimeGranularity,
} from './factories/getTimeFormatterForGranularity';
```
there will be issue with `TimeGranularity` in the consumer app.

Have to create `./types.ts` and put this type there 
then in index 

```ts
export * from './types.ts'
```

packages/superset-ui-demo/storybook/stories/plugins/legacy-preset-chart-big-number/BigNumber/BigNumberStories.tsx
```diff
@@ -74,7 +74,10 @@ export const nullInTheMiddle = () => (
     width={400}
     height={400}
     queryData={{ data: withNulls(testData, 3) }}
-    formData={formData}
+    formData={{
+      ...formData,
+      timeGrainSqla: 'P1W', // weekly
```
kristw, line 79: do you want to make another story for this or remove the change?

packages/superset-ui-time-format/src/factories/getTimeFormatterForGranularity.ts
ktmud: Does it mean the consumer apps will have to `import from '@sueprset-ui/time-format/src/types` too?

packages/superset-ui-demo/storybook/stories/plugins/legacy-preset-chart-big-number/BigNumber/BigNumberStories.tsx
ktmud: Added a new story.

packages/superset-ui-time-format/src/factories/getTimeFormatterForGranularity.ts
kristw: no they can 

`import { TimeGranularity } from '@superset-ui/time-format';`

PR #488, opened 2020-05-13
fix(legacy-plugin-chart-nvd3): disable bad linting rule

suddjian:
🐛 Bug Fix

The `import/extensions` rule makes it impossible to correctly import ts from js.


suddjian:
It's approved, tests pass, but codecov is confused. Merging!

PR #493, opened 2020-05-15
fix(time-format): remove week number

ktmud:
🐛 Bug Fix

Most databases, including [ISO standards](https://en.wikipedia.org/wiki/ISO_8601#Durations), actually uses Monday as the first day of a week. The default week time grain in Superset (`P1W`) should translate to Monday-based weeks, too.

### DIFF
--- a/packages/superset-ui-time-format/src/factories/getTimeFormatterForGranularity.ts
+++ b/packages/superset-ui-time-format/src/factories/getTimeFormatterForGranularity.ts
@@ -23,8 +23,6 @@ import { TimeGranularity } from '../types';
 
 // Translate time granularity to d3-format
 const MINUTE = '%Y-%m-%d %H:%M';
-const SUNDAY_BASED_WEEK = '%Y-%m-%d W%U';
-const MONDAY_BASED_WEEK = '%Y-%m-%d W%W';
 const { DATABASE_DATE, DATABASE_DATETIME } = TimeFormats;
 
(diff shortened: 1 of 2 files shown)


michellethomas:
Thanks I checked presto and it looks like the logic for presto is `date_trunc('week', CAST({col} AS TIMESTAMP))` and when I tested it out it looks like week starts on Monday. Do we know if this is true for all supported `db_engine_specs`? It seems like if any of them evaluate to week start Sunday, this would be incorrect. Can we just show the date for the start of the week (to make sure we will always be accurate)?

ktmud:
> Can we just show the date for the start of the week (to make sure we will always be accurate)?

That was fixed in #486 .

> Do we know if this is true for all supported db_engine_specs?

I also checked Druid, MySQL, PostgreSQL, SQL Lite and some other databases, with the exception of MySQL and Drill, all others are using Monday-based weekdays.

Given how most businesses operate and that Monday-based is also the ISO standard, I propose we move forward with this change and change MySQL & Drill to be Monday-based, too.

cc @mistercrunch @john-bodley @kristw

ktmud:
Another option is to simply remove the week numbers, as: 1) it is not useful for most business settings anyway; 2) different businesses may have different definition of week-of-years.

kristw:
I agree with removing `Wxxx`

michellethomas:
Thanks for resolving this! This seems like a good solution.

PR #509, opened 2020-05-19
feat(time-format): improve support for formatting with granularity in mind

kristw:
🏆 Enhancements

#### Extends core time formatting functions to include granularity support

* `getTimeFormatter(formatId)` to `getTimeFormatter(formatId?: string, granularity?: TimeGranularity)`
* `formatTime(formatId, value)` to `formatTime(formatId: string | undefined, value: Date | null | undefined, granularity?: TimeGranularity)`

#### New APIs

* `getTimeRangeFormatter(formatId?: string)`
* `formatTimeRange(formatId: string | undefined, range: (Date | null | undefined)[])`

#### Others

Convert the `granularity` string into constants.

### Explanation of technical changes

When a `time` includes `granularity`, it means the given `time` is not a single point in time, but a time range. The changes in this PR includes logic for reconstructing the time range from `time` and `granularity`. The current logic assumes that the input time will be a valid output from database based on the `granularity`. For example, if the `granularity` is `P1M` (monthly), the database will always return the first day of the month.

Formatting a time point with granularity then follows these steps:
1. Convert `time` and `granularity` into a time range.
2. Pass the output time range to a `TimeRangeFormatter`, along with `formatId` (~d3 format string) or if no format is specified, look up default format string from a lookup table based on `granularity` (`TimeFormatsForGranularity`)
3. The `TimeRangeFormatter` at the moment is quite naive. Just try to apply same formatting to the `start` and `end` of the time range. If both `start` and `end` have the same output then return just `start`, otherwise return `${start} - ${end}`. 

See more concrete examples in the unit tests.

### Future plans

* `TimeRangeFormatter` can also have its own registry in case developer wants to override how certain format works for range.
* Same with the `TimeFormatsForGranularity`, which each application may have different opinion about default settings.


packages/superset-ui-time-format/src/factories/createMultiFormatter.ts
```diff
@@ -1,5 +1,5 @@
 import { utcFormat, timeFormat } from 'd3-time-format';
-import { utcUtils, localTimeUtils } from '../utils';
+import { utcUtils, localTimeUtils } from '../utils/d3Time';
```
kristw, line 2: move and rename file

packages/superset-ui-time-format/src/types.ts
```diff
@@ -1,20 +1,30 @@
...
-  | '1969-12-28T00:00:00Z/P1W'
-  | '1969-12-29T00:00:00Z/P1W'
-  | 'P1W/1970-01-03T00:00:00Z'
-  | 'P1W/1970-01-04T00:00:00Z';
+export type TimeRangeFormatFunction = (values: (Date | number | undefined | null)[]) => string;
+
+/**
+ * search for `builtin_time_grains` in incubator-superset/superset/db_engine_specs/base.py
+ */
+export const TimeGranularity = {
```
kristw, line 8: Make this also a constant to avoid having to type the awkward granularity string for `WEEK_ENDING_xxx`

packages/superset-ui-time-format/src/types.ts
```diff
@@ -1,20 +1,30 @@
...
+  DAY: 'P1D',
+  WEEK: 'P1W',
+  WEEK_STARTING_SUNDAY: '1969-12-28T00:00:00Z/P1W',
+  WEEK_STARTING_MONDAY: '1969-12-29T00:00:00Z/P1W',
+  WEEK_ENDING_SATURDAY: 'P1W/1970-01-03T00:00:00Z',
+  WEEK_ENDING_SUNDAY: 'P1W/1970-01-04T00:00:00Z',
+  MONTH: 'P1M',
+  QUARTER: 'P0.25Y',
+  YEAR: 'P1Y',
+} as const;
```
ktmud, line 26: why not string enum?

packages/superset-ui-time-format/src/utils/createDate.ts
```diff
@@ -0,0 +1,13 @@
...
+  date = 1,
+  useLocalTime = false,
+}: {
+  year: number;
+  month?: number;
+  date?: number;
+  useLocalTime?: boolean;
+}): Date {
+  return useLocalTime ? new Date(year, month, date) : new Date(Date.UTC(year, month, date));
+}
```
ktmud, line 13: Will this ever need hour minutes seconds?

packages/superset-ui-time-format/src/utils/createTimeRangeFromGranularity.ts
```diff
@@ -0,0 +1,94 @@
+import { TimeGranularity } from '../types';
+import createDate from './createDate';
+
+const MS_IN_SECOND = 1000;
+const MS_IN_MINUTE = 60 * MS_IN_SECOND;
+const MS_IN_HOUR = 60 * MS_IN_MINUTE;
+const MS_IN_DAY = 24 * MS_IN_HOUR;
+const MS_IN_WEEK = 7 * MS_IN_DAY;
```
ktmud, line 8: Can probably remove `MS_IN_` to make it easier to read?

packages/superset-ui-time-format/src/utils/createTimeRangeFromGranularity.ts
```diff
@@ -0,0 +1,94 @@
...
+      createDate(
+        quarter <= 3
+          ? { year, month: quarter * 3, useLocalTime }
+          : { year: year + 1, useLocalTime },
+      ),
+    );
+  }
+
+  // Year
+  return deductOneMs(createDate({ year: year + 1, useLocalTime }));
```
ktmud, line 66: How you thought of utilize `d3-time` intervals? https://github.com/d3/d3-time#interval_round

packages/superset-ui-time-format/src/utils/createTimeRangeFromGranularity.ts
kristw: To avoid confusion with `TimeGranularity.HOUR`, etc. Also make it clear that the value is milliseconds.

packages/superset-ui-time-format/src/utils/createTimeRangeFromGranularity.ts
kristw: That could work. Will look into it.

packages/superset-ui-time-format/src/utils/createDate.ts
kristw: This was used internally for dates at the moment. I could rename to `createTime` and add support for hour, minutes, seconds, milliseconds.

packages/superset-ui-time-format/src/types.ts
kristw: babel has some issue with enum.

kristw:
> In theory, this is definitely a nice feature to have, but I'm worried whether it will lead to potential confusions if we do date truncating on both backend and frontend----users might see inconsistent datetime in charts and raw data.

Could you elaborate on the truncation? I am not sure I understand what you mean.

ktmud:
> > In theory, this is definitely a nice feature to have, but I'm worried whether it will lead to potential confusions if we do date truncating on both backend and frontend----users might see inconsistent datetime in charts and raw data.
> 
> Could you elaborate on the truncation? I am not sure I understand what you mean.

`computeEndTimeFromGranularity` seems to be truncating the `startTime` based on granularity. I was worried since the backend already truncates `timestamp` based on time granularities, would there be a case when some piece of code passes the truncated timestamps to the frontend API with a different granularity?

kristw:
> `computeEndTimeFromGranularity` seems to be truncating the startTime based on granularity. I was worried since the backend already truncates timestamp based on time granularities, would there be a case when some piece of code passes the truncated timestamps to the frontend API with a different granularity?

`computeEndTimeFromGranularity` computes the `endTime`. I don't think it touches `startTime` at all. This is computing true time range for each data point. 

For example, when a data point is returned from database, it seems like a single point in time.

`{ time: 2010-04-01, value: 300 }`

However, it is rarely a single point in time. If this query has `MONTH` granularity, the data point above actually means the entire April.

This is what the actual data point is.

`{ time: [2010-04-01, 2010-04-30], value: 300 }`

so `computeEndTimeFromGranularity` takes `2010-04-01` and `MONTH` and returns the end time `2010-04-30` just for formatting. 

Now we know the time range is `[2010-04-01, 2010-04-30]`, it depends on the format string. 
* If the format is `%Y-%m` this will be `2010-04 - 2010-04` which looks redundant, so we can merge and return `2010-04`
* If the format string is `%Y-%m-%d` this will be `2010-04-01 - 2010-04-30`

ktmud:
> `computeEndTimeFromGranularity` computes the `endTime`. I don't think it touches `startTime` at all. This is computing true time range for each data point.

I mean It calculates an `endTime` based on truncated `startTime`, since it only takes parts of `startTime`. I can see why this function can be useful, but would there be a case the truncating is done differently on the frontend and backend? For example, some db engine may round dates up instead of round-down? Or when backend rounded the dates to weeks, but the frontend still uses days because some chart forgot to pass the granularity to the formatter?

Maybe I was just overthinking.

kristw:
> For example, some db engine may round dates up instead of round-down? 

The granularity `WEEK_ENDING_xxx` rounds date up and this function will react accordingly. I am not aware of other cases that it does not round down. (This would be an even broader issue that already exists if some database do round up by default.) The current code handles all granularity listed in the superset `db_engine`.

> Or when backend rounded the dates to weeks, but the frontend still uses days because some chart forgot to pass the granularity to the formatter?

This is already the current behavior I think. This feature allows it to be more correct if the granularity is passed to the formatter. It is an opt-in feature.

packages/superset-ui-time-format/src/utils/createTimeRangeFromGranularity.ts
kristw: The `floor` and `ceil` functions work if the value is in the middle of the range but not so useful for the case we are handlings that the inputs are already at the beginning of the range. It also ceils to `YYYY-mm-dd 00:00:00.000` instead of `YYYY-mm-dd 23.59.59.999` that I would like.

PR #522, opened 2020-05-21
fix(legacy-plugin-chart-table): parse numeric pageLength

ktmud:
🐛 Bug Fix

Pagination for table chart had stopped working because `pageLength` param somehow is not string any more.

This fix makes `transformProps` handle both numeric and string values and added unit tests.


ktmud:
`git bisect` says bug was introduced in https://github.com/apache/incubator-superset/pull/9757

kristw:
Hmm. Not sure what was changed in that mass bump.

PR #539, opened 2020-05-26
fix: resolves query issue where wordcloud had no words

rusackas:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
Word cloud chart was not getting any words back from the API. This resolves that issue. I think it was an accidental deletion in another PR. 

🏠 Internal


villebro:
This was intentional, should be fixed now as https://github.com/apache/incubator-superset/pull/9908 is merged!

rusackas:
OK, closing for now, will rebase and check :D

PR #540, opened 2020-05-26
fix: nvd3 annotation tooltip

etr2460:
🐛 Bug Fix

It looks like a long time ago, the `sourceType` field for annotations was migrated from being `table` to `NATIVE`. This broke the tooltip for annotations. Since I'm quite unfamiliar with this code, and for backwards compatibility, I've left the old selector in but also added the new one which should resolve the issue.

New class name in superset:
image

to: @ktmud @graceguo-supercat @kristw


etr2460:
After researching a bit more, it seems like the old class is still applied for annotation events built from a table as opposed to an annotation layer. Therefore, we need both classes here for the styles to apply properly

PR #547, opened 2020-05-27
feat(color): support better color interpolation for sequential schemes

kristw:
🏆 Enhancements

* add `modifyRange` option to `createLinearScale(domain, modifyRange = false)`. This extends the original behavior that always modify `domain` if needed.

```
   * Return a linear scale with a new domain interpolated from the input domain
   * to match the number of elements in the color scheme
   * because D3 continuous scale uses piecewise mapping between domain and range.
   * This is a common use-case when the domain is [min, max]
   * and the palette has more than two colors.
   *
   * @param domain domain of the scale
   * @param modifyRange Set this to true if you don't want to modify the domain and
   * want to interpolate range to have the same number of elements with domain instead.
```

* add `extent` option for `.getColors(numColors, extent)` to reduce the range of the color scheme, inspired `SchemeParams` option in [vega-lite](https://vega.github.io/vega-lite/docs/scale.html)

```
   * Get colors from this scheme
   * @param numColors number of colors to return.
   * Will interpolate the current scheme to match the number of colors requested
   * @param extent The extent of the color range to use.
   * For example [0.2, 1] will rescale the color scheme
   * such that color values in the range [0, 0.2) are excluded from the scheme.
```

* Use HCL color space interpolation instead of RGB interpolation https://bl.ocks.org/mbostock/3014589
* Add more code comments for auto-complete
* Add more unit tests

### DIFF
--- a/packages/superset-ui-color/src/SequentialScheme.ts
+++ b/packages/superset-ui-color/src/SequentialScheme.ts
@@ -1,4 +1,5 @@
 import { scaleLinear } from 'd3-scale';
+import { interpolateHcl } from 'd3-interpolate';
 import ColorScheme, { ColorSchemeConfig } from './ColorScheme';
 
 function range(count: number) {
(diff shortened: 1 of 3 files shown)


packages/superset-ui-color/src/SequentialScheme.ts
```diff
@@ -10,6 +11,11 @@ function range(count: number) {
   return values;
 }
 
+function rangeZeroToOne(steps: number) {
+  const denominator = steps - 1;
+  return range(steps).map(i => i / denominator);
```
ktmud, line 16: Can replace `range` with `[...new Array(steps).keys()]`

kristw:
@ktmud `d3.quantize` and `d3.piecewise` are awesome.

ktmud:
`d3` API is full of gems like this.

PR #548, opened 2020-05-27
feat(plugin-chart-choropleth-map): add tooltip

kristw:
**NOTE: This is merging into a feature branch, not `master`, and will have subsequent PRs into the feature branch until completion.**

## 🏆 Enhancements

* Create `plugin-chart-choropleth-map` to replace the `country-map` and `world-map` charts.

### New features

Support tooltip

### Screenshots

image

### DIFF
--- a/plugins/plugin-chart-choropleth-map/src/chart/ChoroplethMap.tsx
+++ b/plugins/plugin-chart-choropleth-map/src/chart/ChoroplethMap.tsx
@@ -21,9 +21,11 @@ import { t } from '@superset-ui/translation';
 import { Zoom } from '@vx/zoom';
 import { localPoint } from '@vx/event';
 import { RectClipPath } from '@vx/clip-path';
+import { withTooltip } from '@vx/tooltip';
 import { keyBy } from 'lodash';
 import { geoPath } from 'd3-geo';
 import type { FeatureCollection } from 'geojson';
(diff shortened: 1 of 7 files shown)


plugins/plugin-chart-choropleth-map/src/chart/ChoroplethMap.tsx
```diff
@@ -58,31 +60,30 @@ export type ChoroplethMapVisualProps = {
...
+  };
 
 const defaultProps = {
   data: [],
   encoding: {},
   map: 'world',
 };
 
-const missingItem: ChoroplethMapChannelOutputs = DefaultChannelOutputs;
+const missingItem = DefaultChannelOutputs;
```
ktmud, line 76: Is this variable necessary?

kristw: was planning to have a different style for missing data. Now using the default output as placeholder.

kristw:
> Just one small feature request: is it possible to add "double-click to zoom in"? Currently it the map jumps a little when double clicked, which is not very smooth.

Now the double-click does the zoom in by 10%. Agree with the not smooth zooming and 10% is probably unclear whether the zoom level was changed or was it just translated. 

Smooth zooming transition was easy to do when implementing with pure `d3`. Now with the zoom being handled in react, I have to figure out how to do the transition with `react-spring` or alike. Will do a follow-up PR.

PR #553, opened 2020-05-30
feat(superset-ui-style): export ThemeProvider and useTheme

ktmud:
Export `ThemeProvider`, `useTheme` and `withTheme` from `emotion-theming` in `@superset-ui/style`, since they are almost always used together with `supersetTheme`.

This makes importing and package dependencies in Superset app (and dependent plugins) a little easier.

🏆 Enhancements


ktmud:
> lgtm. wonder if we should reexport `styled` as well.

Done.

PR #554, opened 2020-05-30
fix(chart): update chart data query

kristw:
🐛  Bug Fix

Update chart data queries to fit the new requirements of API `/api/v1/chart/data`
(logic copied from `incubator-superset`)

### DIFF
--- a/packages/superset-ui-chart/src/clients/ChartClient.ts
+++ b/packages/superset-ui-chart/src/clients/ChartClient.ts
@@ -85,7 +85,7 @@ export default class ChartClient {
       const buildQuery = (await buildQueryRegistry.get(visType)) ?? (() => formData);
 
       if (useLegacyApi) {
-        this.client
+        return this.client
           .post({
             ...options,
             endpoint: '/superset/explore_json/',
--- a/packages/superset-ui-chart/src/clients/ChartClient.ts
+++ b/packages/superset-ui-chart/src/clients/ChartClient.ts
@@ -84,19 +84,34 @@ export default class ChartClient {
       const { useLegacyApi } = metaDataRegistry.get(visType)!;
       const buildQuery = (await buildQueryRegistry.get(visType)) ?? (() => formData);
 
-      return this.client
-        .post({
-          headers: { 'Content-Type': 'application/json' },
-          endpoint: useLegacyApi ? '/superset/explore_json/' : '/api/v1/chart/data',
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart/src/clients/ChartClient.ts
```diff
@@ -84,19 +84,34 @@ export default class ChartClient {
...
+          .then(({ json }) => json as QueryData);
+      }
+
+      return (
+        this.client
+          .post({
+            ...options,
+            endpoint: '/api/v1/chart/data',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify(buildQuery(formData)),
```
ktmud, line 105: I wonder if we can add a method `postJson` or option `post({ json: ... })` to `SupersetClient`. I'd imagine we are going to see more and more POST json requests.

suddjian: I think it'd be best to have a `SupersetClientV1` that uses conventions for the v1 api by default across the board.

kristw: This will be refactored to use functions from #555.

ktmud: @suddjian If we want something like `SupersetClientV1`, then we should probably remove the flexibility of specifying arbitrary API endpoint.

@kristw I think a `postJson` or `post({ json: ... })` is still needed, for new API endpoints like seen here: https://github.com/apache/incubator-superset/pull/10104/files#r442641384 .

There should be two layer of abstraction: 1) a utility client that make requests to arbitrary endpoints 2) a true Superset client with schema enforced.

For now with all the unstable new APIs, people should start using 1, so they don't have to go back and forth editing two repos. After backend API becomes more stable, we migration them to 2.

@villebro @rusackas @etr2460  thoughts?

rusackas:
Seems like there's some refactor work ahead (I like the SupersetClientV1 idea with schema'd & freeform endpoints), but this seems mergeable for now, if rebased.

ktmud:
I think this PR has been rendered obsolete by https://github.com/apache-superset/superset-ui/commit/b7903428e9469cb0980471b57a1033f9c4bb9602

There are more refactoring work going on with `buildQuery` et al. but I was distracted by something else. Will come back to this in the next couple of weeks,

PR #555, opened 2020-05-30
feat(query): add functions to wrap api calls with typings

kristw:
🏆  Enhancements

Facilitate the calls to `explore_json` and `api/v1/chart/data` by specify required params and provide typescript type for their `response`. Logics are based on the [chartAction.js](https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/chart/chartAction.js) code in `incubator-superset`. 

* `postChartData()`
* `fetchExploreJson()`

#### Future plan

* Add more functions for other APIs. 
* Replace calls in `incubator-superset` and `@superset-ui/chart` to use these functions.


packages/superset-ui-query/src/api/legacy/fetchExploreJson.ts
```diff
@@ -0,0 +1,28 @@
+import { SupersetClientInterface, SupersetClient, RequestConfig } from '@superset-ui/connection';
+import { QueryFormData } from '../../types/QueryFormData';
```
felixcodes, line 2: thank god for this custom type... i have mild ptsd from `formData` lol

packages/superset-ui-query/src/api/legacy/fetchExploreJson.ts
```diff
@@ -0,0 +1,28 @@
...
+  method = 'POST',
+  requestConfig,
+  url = '/superset/explore_json/',
+  formData,
+}: Params) {
+  const fetchFunc = method === 'GET' ? client.get : client.post;
+
+  return fetchFunc({
+    ...requestConfig,
+    // TODO: Have to transform formData as query string for GET
```
felixcodes, line 24: Do you have an example for what a query string could look like?

packages/superset-ui-query/src/api/v1/types.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable camelcase */
+export interface V1ChartDataResponseResult {
```
ktmud, line 2: I'm not sure about the `V1` prefix. Having a prefix makes it look temporary or deprecated. Was `ChartDataResponseResult` used elsewhere?

kristw: Agree with the doubts about variable names. This was the hardest part about this PR.
I had trouble with naming and went back and forth whether to version or not. Still not fully satisfied TBH. On one hand I want to avoid naming collision if there is v2. On the other hand the `V1xxx` looks a bit weird.

packages/superset-ui-query/src/api/legacy/fetchExploreJson.ts
kristw: There is a utility function in `incubator-superset` that has to be ported here.

packages/superset-ui-query/src/api/v1/types.ts
kristw: cc @rusackas @suddjian if you have thoughts about naming.

suddjian: I think marking it as v1 is useful in cases where the interface specifically relates to behavior of the v1 api.

kristw: I still don't feel strongly either way so I will merge and adjust later if anybody has more comments.

PR #557, opened 2020-06-01
feat(demo): add helpers for resizable chart demo

ktmud:
🏆 Enhancements

📜 Documentation

Make chart demos resizable

Snip20200531_23

[Test URL](https://superset-ui-git-fork-ktmud-resizable-demo.superset.now.sh/?path=/story/core-packages-superset-ui-chart--resizable)

Snip20200531_22

### DIFF
--- a/packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
+++ b/packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
@@ -1,32 +1,45 @@
+import styled from '@superset-ui/style';
 import React, { useState, ReactNode } from 'react';
 import { DecoratorFunction } from '@storybook/addons';
 import ResizablePanel, { Size } from './ResizablePanel';
 
+export const SupersetBody = styled.div`
+  background: #f5f5f5;
+  padding: 16px;
+  min-height: 100%;
+
+  .panel {
+    margin-bottom: 0;
+  }
+`;
+
 export default function ResizableChartDemo({
   children,
+  panelPadding = 30,
   initialSize = { width: 500, height: 300 },
 }: {
-  children: (size: Size) => ReactNode;
+  children: (innerSize: Size) => ReactNode;
+  panelPadding?: number;
   initialSize?: Size;
 }) {
   // size are all inner size
   const [size, setSize] = useState(initialSize);
   return (
-    <div className="superset-body">
+    <SupersetBody>
       <ResizablePanel initialSize={initialSize} onResize={(e, data) => setSize(data.size)}>
-        {children(size)}
+        {children({ width: size.width - panelPadding, height: size.height - panelPadding })}
       </ResizablePanel>
-    </div>
+    </SupersetBody>
   );
 }
 
 export const withResizableChartDemo: DecoratorFunction<ReactNode> = (storyFn, context) => {
   const {
-    parameters: { initialSize },
+    parameters: { initialSize, panelPadding },
   } = context;
   return (
-    <ResizableChartDemo initialSize={initialSize as Size | undefined}>
-      {size => storyFn({ ...context, ...size })}
+    <ResizableChartDemo initialSize={initialSize as Size | undefined} panelPadding={panelPadding}>
+      {innerSize => storyFn({ ...context, ...innerSize })}
     </ResizableChartDemo>
   );
 };
--- a/packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
+++ b/packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
@@ -1,0 +1,32 @@
+import React, { useState, ReactNode } from 'react';
+import { DecoratorFunction } from '@storybook/addons';
+import ResizablePanel, { Size } from './ResizablePanel';
+
+export default function ResizableChartDemo({
+  children,
+  initialSize = { width: 500, height: 300 },
+}: {
+  children: (size: Size) => ReactNode;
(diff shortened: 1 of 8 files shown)


packages/superset-ui-demo/storybook/shared/components/ResizablePanel.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { PropsWithChildren, ReactNode } from 'react';
+import { ResizableBox, ResizableBoxProps } from 'react-resizable';
+
+import 'react-resizable/css/styles.css';
+
+export default function ResiablePanel({
```
felixcodes, line 6: small nit: You have a typo right here @ktmud 

`ResiablePanel` -> `ResizablePanel`

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
```diff
@@ -0,0 +1,24 @@
...
+  width: number;
+  height: number;
+}
+
+export default function ResizableChartDemo({ children }: { children: (size: Size) => ReactNode }) {
+  // size are all inner size
+  const initialSize = { width: 500, height: 300 };
+  const [size, setSize] = useState(initialSize);
+  return (
+    <div className="superset-body">
```
kristw, line 14: Is this className needed?

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useState, ReactNode } from 'react';
+import ResizablePanel from './ResizablePanel';
+
+interface Size {
```
kristw, line 4: minor. There is also a type `Dimension` in `@superset-ui/dimension`

```ts
import { Dimension } from '@superset-ui/dimension';
```

packages/superset-ui-demo/storybook/shared/components/ResizablePanel.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { ResizableBox, ResizableBoxProps } from 'react-resizable';
+
+import 'react-resizable/css/styles.css';
+
+export default function ResizablePanel({
+  children,
+  heading = undefined,
+  initialSize = { width: 500, height: 300 },
+  minConstraints = [100, 100] as [number, number],
+  panelPadding = 30,
```
kristw, line 11: not sure about adding padding here though. I would prefer not to add because the components when used in Superset and other places do not have this padding. It may be misleading during the component development that there are paddings.

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
ktmud: Yes, it's for having a light gray background in main body (as `panel` is white).

kristw: Could we do this in `js`/`css-in-js` to move away from declaring this in `storybook.css`?

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
ktmud: Semantically, this is from `react-resizable`. I'll import it from there instead. [https://github.com/apache-superset/superset-ui/pull/557/commits/b297fea9fd59b4b742243e5c13f40eba872a58e7 ](https://github.com/apache-superset/superset-ui/pull/557/commits/b297fea9fd59b4b742243e5c13f40eba872a58e7)#diff-e296f83c808a6bf79336f4e585f97bd7

kristw: sure. this is pretty much optional.

packages/superset-ui-demo/storybook/shared/components/ResizablePanel.tsx
ktmud: The chart containers in Superset do have some padding. I was simulating the environment when charts are rendered in dashboard panels or on the explore page.

Just moved the paddings to `ResizableChartDemo`, so users specify `initialSize` for the panel, instead of the content in the panel (i.e. `SupersetChart`). Not sure which ways is less misleading though...

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  .panel {
+    margin-bottom: 0;
+  }
+`;
+
+export default function ResizableChartDemo({
+  children,
+  panelPadding = 30,
+  initialSize = { width: 500, height: 300 },
```
kristw, line 19: I was using initialSize 400x400 (1:1) so we can take screenshots for thumbnails easily.

packages/superset-ui-demo/storybook/shared/components/ResizableChartDemo.tsx
```diff
@@ -0,0 +1,45 @@
...
+  min-height: 100%;
+
+  .panel {
+    margin-bottom: 0;
+  }
+`;
+
+export default function ResizableChartDemo({
+  children,
+  panelPadding = 30,
```
kristw, line 18: `padding` for cells in Superset dashboard is `16px`. should this be `32`?
I think `padding` usually means one side. May be better make this padding 16 and do the `*2` when computing children width/height.

PR #567, opened 2020-06-03
chore(Markup chart):  add emotion styling to Markup chart

pkdotson:
💔 Breaking Changes

🏆 Enhancements
add emotion styling to Markup chart
📜 Documentation

🐛 Bug Fix

🏠 Internal

Before
￼
C Superset o Security v

After
0 Superset o Security


plugins/legacy-plugin-chart-markup/package.json
```diff
@@ -28,6 +28,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@superset-ui/style": "^0.13.26",
```
kristw, line 31: `peerDependencies`

PR #571, opened 2020-06-03
build: allow building only one (or more) package

ktmud:
🏆 Enhancements

The script used for building one package (`scirpts/build.js`) is broken. This PR fixes it and adds it to `yarn build`.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -18,15 +18,17 @@ const run = cmd => {
   }
 };
 
+const BABEL_CONFIG = '--config-file=../../babel.config.js';
+
 if (glob) {
   // lint is slow, so not turning it on by default
   if (extraArgs.includes('--lint')) {
     run(`nimbus eslint {packages,plugins}/${glob}/{src,test}`);
   }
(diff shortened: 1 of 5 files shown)


scripts/copyAssets.js
```diff
@@ -1,6 +1,8 @@
-/* eslint-disable import/no-extraneous-dependencies, no-console */
```
ktmud, line 1: Renamed this file because of tab autocompletion... I would stuck at `vim ./script/build` when trying to edit `build.js` from terminal.

scripts/build.js
```diff
@@ -1,5 +1,7 @@
+#!/bin/env node
+/* eslint-disable no-console */
```
rusackas, line 2: NIT: could just use `eslint-disable-next-line` for the one existing use case, so this file won't get muddied up by accident :)

...but I'm not losing any sleep over it.

ktmud: I was thinking about that but felt a cli script should always allow `console.log` anyway.

PR #578, opened 2020-06-04
chore(chord chart): add emotion styling to chord chart

pkdotson:
💔 Breaking Changes

🏆 Enhancements
 Add emotion styling to chord chart 
📜 Documentation

🐛 Bug Fix

🏠 Internal

Before
il Charts

after
• Superset o Security


plugins/legacy-plugin-chart-chord/src/ReactChord.jsx
```diff
@@ -16,7 +16,44 @@
...
+  otherProps: {},
+};
+
+Chord.propTypes = {
+  className: PropTypes.string.isRequired,
+  otherProps: PropTypes.objectOf(PropTypes.any),
+};
+
+export default styled(Chord)`
+  .superset-legacy-chart-chord svg #circle circle {
```
rusackas, line 45: Just a note for all of these plugins: Emotion uses SASS, so you can condense these styles a bit, e.g.

```.superset-legacy-chart-chord svg{
  #circle {
    circle {
      fill: none;
      pointer-events: all;
    }
    &:hover path.fade {
      opacity: 0.2;
    }
  } 
  .group path {
    fill-opacity: 0.6;
  }
  path.chord {
    stroke: #000;
    stroke-width: 0.25px;
  }
}```

And I _think_ you might not even need the `.superset-legacy-chart-chord` there, since the CSS should be scoped to the component.

PR #591, opened 2020-06-08
feat: Superset observable

nytai:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/plugin-chart-observable/src/controlPanel.ts
```diff
@@ -0,0 +1,128 @@
+// @ts-nocheck
```
nytai, line 1: We should probably fix this

nytai:
closing as there hasn't been much activity lately. Maybe one day we'll actually get this in

PR #596, opened 2020-06-12
build: fix typing rebuild for single packages

ktmud:
🐛 Bug Fix

Sometimes build a single package via `yarn build "*package-name"` does not update rebuild the type declaration files. This PR fixes it by cleaning the existing built files first.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -2,6 +2,7 @@
 /**
  * Build plugins specified by globs
  */
+const rimraf = require('rimraf');
 const { spawnSync } = require('child_process');
 
 const glob = process.argv[2];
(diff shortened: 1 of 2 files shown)


scripts/build.js
```diff
@@ -27,6 +28,7 @@ if (glob) {
   if (extraArgs.includes('--lint')) {
     run(`nimbus eslint {packages,plugins}/${glob}/{src,test}`);
   }
+  rimraf.sync(`./{packages,plugins}/${glob}/{lib,esm,tsconfig.tsbuildinfo}`);
```
rusackas, line 31: Not super relevant here, but this reminds me, it might be nice to do something similar to remove package-lock.json files that sometimes sneak their way into plugins, which break the `yarn ci:create-patch-version` process.

ktmud: `pckage-lock.json` doesn't seem to matter for the build process. I'll create a PR to remove them before `ci:create-patch-version`.

PR #602, opened 2020-06-15
chore: enabling Emotion for styles on legacy table

rusackas:
💔 Breaking Changes

🏆 Enhancements
Moves CSS into Emotion (currently, with no changes)

Nothing dynamic in here _yet_, but there are a couple colors that can/should come from the theme in `@superset-ui/style` (as a peer dependency). That theme was just updated, and there's a PR to bump it in Superset. We can then circle back to using the new colors here, and in other plugins.

📜 Documentation

🐛 Bug Fix

🏠 Internal


ktmud:
Hmm, I did this in the new table chart, too: https://github.com/apache-superset/superset-ui/pull/584/files#diff-cf73fcc3f3abc557f6e65d2a9d5b5de3 , which currently is just a drop-in replacement for the table chart.

ktmud:
I think the new table chart is more or less ready to merge (pending final some review and testing). Do you mind waiting for it to be merged before working on the colors?

rusackas:
> I think the new table chart is more or less ready to merge (pending final some review and testing). Do you mind waiting for it to be merged before working on the colors?

No problem! I thought this legacy one was sticking around, so I thought I'd just throw this in, too.

PR #609, opened 2020-06-16
feat(legacy-table-chart): add query mode switch

ktmud:
🏆 Enhancements

 

Rearrange form controls for Table chart, replace GROUP BY and NOT GROUP BY with a "Query Mode" switch.

Added to both the legacy table and [the new table chart](https://github.com/apache-superset/superset-ui/pull/584).

Superset app also need to be updated to accept the frontend passing in both `all_columns` and `metrics`.

Related: https://github.com/apache/incubator-superset/pull/10113

### DIFF
--- a/packages/superset-ui-chart-controls/src/components/RadioButtonControl.tsx
+++ b/packages/superset-ui-chart-controls/src/components/RadioButtonControl.tsx
@@ -18,7 +18,7 @@
  */
 import React, { ReactText, ReactNode, MouseEvent, useCallback } from 'react';
 import styled from '@superset-ui/style';
-import { InfoTooltipWithTrigger } from '@superset-ui/chart-controls';
+import { InfoTooltipWithTrigger } from './InfoTooltipWithTrigger';
 
 export interface RadioButtonOption {
   label: string;
--- a/packages/superset-ui-chart-controls/src/shared-controls/components.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/components.tsx
@@ -1,0 +1,28 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import RadioButtonControl from '../components/RadioButtonControl';
+
+export * from '../components/RadioButtonControl';
+
+/**
+ * Aliases for Control Components
+ */
+export default {
+  RadioButtonControl,
+};
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -133,6 +133,9 @@ export type InternalControlType =
   | 'AdhocFilterControlVerifiedOptions'
   | keyof SharedControlComponents; // expanded in `expandControlConfig`
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ControlType = InternalControlType | React.ComponentType<any>;
+
 export type TabOverride = 'data' | boolean;
 
 /**
@@ -163,14 +166,14 @@ export type TabOverride = 'data' | boolean;
  *    be visibile.
  */
 export interface BaseControlConfig<
-  T = unknown,
+  T extends ControlType = ControlType,
   O extends SelectOption = SelectOption,
   V = unknown
 > {
   type: T;
   label?: ReactNode;
   description?: ReactNode;
-  default?: unknown;
+  default?: V;
   renderTrigger?: boolean;
   validators?: ControlValueValidator<T, O, V>[];
   warning?: ReactNode;
@@ -274,16 +277,11 @@ export interface OverrideSharedControlItem<A extends SharedControlAlias = Shared
   override: Partial<SharedControls[A]>;
 }
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type CustomControlItem<P = any> = P extends SharedControlAlias
-  ? {
-      name: P;
-      config: SharedControls[P];
-    }
-  : {
-      name: string;
-      config: CustomControlConfig<P>;
-    };
+export type CustomControlItem = {
+  name: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  config: BaseControlConfig<any, any, any>;
+};
 
 // use ReactElement instead of ReactNode because `string`, `number`, etc. may
 // interfere with other ControlSetItem types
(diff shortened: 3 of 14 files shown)


plugins/legacy-plugin-chart-table/src/RadioButtonControl.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <Styles>
+      {controlLabel && (
+        <div className="control-label">
+          {controlLabel}{' '}
+          {hovered && description && (
+            <InfoTooltipWithTrigger tooltip={description} placement="top" />
+          )}
+        </div>
+      )}
+      <div className="btn-group btn-group-sm">
```
mistercrunch, line 76: wondering if `react-bootstrap` components would be better here (?). Historically we've been pushing in that direction, and it will make a migration easier to have consistency in what we're replacing...

ktmud: For the most part, `react-bootstrap` is nothing but a generator for CSS classes. There are a lot of places that directly use bootstrap classes in Superset already, we'd have to do the migration for bare class names anyway. I'd rather not to use an API that I know will soon be deprecated.

Pure css classes would be actually easier to clean because you don't have to remember or look up all the API of `react-bootstrap`.

packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -253,28 +272,23 @@ export type SharedSectionAlias =
...
-export type ControlSetItem =
-  | SharedControlAlias
-  | OverrideSharedControlItem
-  | CustomControlItem
-  // use ReactElement instead of ReactNode because `string`, `number`, etc. may
-  // interfere with other ControlSetItem types
-  | ReactElement
-  | null;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  config: BaseControlConfig<any, any, any>;
```
ktmud, line 283: The typing is so tricky. Had to use a lot of `any`s to make it work. Happy to revise if people have better suggestions.

packages/superset-ui-chart-controls/src/utils/expandControlConfig.tsx
```diff
@@ -0,0 +1,82 @@
...
+ *     config: {
+ *        type: ...,
+ *        ...
+ *     }
+ *   }
+ */
+export function expandControlConfig(
+  control: ControlSetItem,
+  controlOverrides: ControlOverrides = {},
+): ExpandedControlItem {
```
ktmud, line 44: This util function is [used to consolidate the control configs](https://github.com/apache/incubator-superset/pull/10113/files#diff-e0e47e59023c2350722fd73d59588413) in `sectionsToRender`, such that initial rendering and subsequent state updates can refer to the the same processed control configs.

packages/superset-ui-chart-controls/src/shared-controls/components.tsx
```diff
@@ -0,0 +1,28 @@
...
+import RadioButtonControl from '../components/RadioButtonControl';
+
+export * from '../components/RadioButtonControl';
+
+/**
+ * Aliases for Control Components
+ */
+export default {
+  RadioButtonControl,
+};
```
ktmud, line 28: The intention is we should slowly move all control components in [superset-frontend/src/explore/components/controls/index.js](https://github.com/apache/incubator-superset/blob/291306392443a5a0d0e2ee0cc4a95d37c56d4589/superset-frontend/src/explore/components/controls/index.js#L44) here.

Currently the API is compatible to have them in both places.

packages/superset-ui-chart-controls/src/utils/expandControlConfig.tsx
```diff
@@ -0,0 +1,82 @@
...
+  if (!control || typeof control === 'string' || React.isValidElement(control)) {
+    return control as ReactElement;
+  }
+  // already fully expanded control config
+  if ('name' in control && 'config' in control) {
+    return {
+      ...control,
+      config: {
+        ...control.config,
+        type: expandControlType(control.config.type as ControlType),
```
ktmud, line 66: Control component aliases in `chart-controls` (currently only `RadioButtonControl`) are expanded here. The rest are expanded in [superset-frontend/src/explore/components/Control.jsx](https://github.com/apache/incubator-superset/blob/afd5c086937aac7569f1fcaded2c41779ba4c05d/superset-frontend/src/explore/components/Control.jsx#L81).

plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -21,45 +22,163 @@ import { t } from '@superset-ui/translation';
...
+  ControlConfig,
   ColumnOption,
+  ControlStateMapping,
+  ControlPanelConfig,
+  ControlPanelsContainerProps,
 } from '@superset-ui/chart-controls';
 import { validateNonEmpty } from '@superset-ui/validator';
 import { smartDateFormatter } from '@superset-ui/time-format';
 
 export const PAGE_SIZE_OPTIONS = formatSelectOptions<number>([[0, t('All')], 10, 20, 50, 100, 200]);
```
ktmud, line 34: This file is almost identical to `legacy-plugin-chart-table/src/controlPanel.tsx`, except updates on exporting `PAGE_SIZE_OPTIONS ` and default option values.

packages/superset-ui-chart-controls/src/components/RadioButtonControl.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+export default function RadioButtonControl({
+  label: controlLabel,
+  description,
+  value: initialValue,
+  hovered,
+  options,
+  onChange,
+}: RadioButtonControlProps) {
+  const currentValue = initialValue || options[0].value;
```
kristw, line 59: what if options.length === 0?

packages/superset-ui-chart-controls/src/types.ts
kristw: Would `unknown` work or does it have to be `any`?

plugins/legacy-plugin-chart-table/src/controlPanel.tsx
```diff
@@ -22,41 +22,160 @@ import { t } from '@superset-ui/translation';
...
+  ControlConfig,
   ColumnOption,
+  ControlStateMapping,
   ControlPanelConfig,
+  ControlPanelsContainerProps,
 } from '@superset-ui/chart-controls';
 import { validateNonEmpty } from '@superset-ui/validator';
 
+export enum QueryMode {
+  agg = 'agg',
```
kristw, line 34: Perhaps full word `AGGREGATE` for clarity

plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -21,45 +22,163 @@ import { t } from '@superset-ui/translation';
...
+            config: {
+              type: 'SelectControl',
+              label: t('Columns'),
+              description: t('Columns to display'),
+              multi: true,
+              freeForm: true,
+              allowAll: true,
+              commaChoosesOption: false,
+              default: [],
+              optionRenderer: (c: never) => <ColumnOption showType column={c} />,
```
kristw, line 126: why is it `never`?

packages/superset-ui-chart-controls/src/components/RadioButtonControl.tsx
ktmud: It will never be 0 since the options are manual. Can probably return null when `options` are invalid though.

plugins/plugin-chart-table/src/controlPanel.tsx
ktmud: It doesn't have to be. The general `BaseControlConfig` does not have `optionRenderer`. Just moved the whole config to a new typed variable and reused typing from sharedControls.

plugins/legacy-plugin-chart-table/src/controlPanel.tsx
ktmud: done

packages/superset-ui-chart-controls/src/types.ts
ktmud: It has to be `any`, as it is used as a catch-all type in `ControlSetRow`. `unknown` in instantiated generic types are not compatible with another instance of the same generic type.

PR #610, opened 2020-06-16
refactor(control-utils): reorganize files and refine typing

ktmud:
🏆 Enhancements

🏠 Internal

Move UI components in `@superset-ui/control-utils` to a separate folder and optimize typing for custom control configs.

While working on #609 , I saw the need to refactor how control panel configs are set up. This PR is the first step to achieve that.

My goal would be:

- Control configs allow ad-hoc override
- Differentiate overrides on shared controls with full custom controls
- Parse control panel config before rendering them
- Use the same parsed config for initial rendering and reading state


packages/superset-ui-control-utils/src/types.ts
```diff
@@ -228,34 +233,42 @@ export type SharedSectionAlias =
   | 'sqlaTimeSeries'
   | 'NVD3TimeSeries';
 
-export interface ControlItem {
+export interface OverrideSharedControlItem {
   name: SharedControlAlias;
-  config: Partial<ControlConfig>;
+  override: Partial<SharedControlConfig>;
```
ktmud, line 238: This is new API. Superset app needs to be updated to support this. 

Will file a separate PR.

ktmud:
cc @suddjian @pkdotson @kristw

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -164,24 +159,33 @@ export type TabOverride = 'data' | boolean;
...
-  validators?: Validator[];
+  validators?: ControlValueValidator[];
   warning?: ReactNode;
   error?: ReactNode;
   // override control panel state props
   mapStateToProps?: (
     state: ControlPanelState,
     control: ControlConfig,
-    actions: ControlPanelActionDispathers,
+    actions?: ControlPanelActionDispathers,
```
kristw, line 174: typo `Dispat"c"hers`

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -228,34 +233,42 @@ export type SharedSectionAlias =
...
 // Ref:
 //  - superset-frontend/src/explore/components/ControlPanelsContainer.jsx
 //  - superset-frontend/src/explore/components/ControlPanelSection.jsx
 export interface ControlPanelSectionConfig {
   label: ReactNode;
-  controlSetRows: ControlSetRow[];
   description?: ReactNode;
   expanded?: boolean;
   tabOverride?: TabOverride;
+  controlSetRows: ControlSetRow[];
```
kristw, line 264: Is this intentional? Seems like just changing order.

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -164,24 +159,33 @@ export type TabOverride = 'data' | boolean;
  *    be visibile.
  */
 export interface GeneralControlConfig {
-  type: InternalControlType | React.ComponentType;
+  type: unknown;
```
kristw, line 162: Why is this changed to `unknown`?

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -52,13 +52,8 @@ export interface DatasourceMeta {
 
 export interface ControlPanelState {
   form_data: QueryFormData;
-  datasource?: DatasourceMeta | null;
-  options?: ColumnMeta[];
```
kristw, line 56: What happen to `options`?

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -52,13 +52,8 @@ export interface DatasourceMeta {
...
-  controls?: {
-    comparison_type?: {
-      value: string;
-    };
-  };
+  datasource: DatasourceMeta | null;
+  controls: ControlStateMapping;
```
kristw, line 56: This field is now required and also less strict about its value.

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -228,34 +233,42 @@ export type SharedSectionAlias =
...
-  config: ControlConfig;
+  config: CustomComponentControlConfig<P>;
 }
 
-export type ControlSetItem = SharedControlAlias | ControlItem | CustomControlItem | ReactNode;
+export type ControlSetItem =
+  | SharedControlAlias
+  | OverrideSharedControlItem
+  | CustomControlItem
+  | ReactElement
```
kristw, line 251: would be nice to add explanation why 
`ReactNode` vs `ReactElement | null`

packages/superset-ui-control-utils/src/types.ts
ktmud: OMG, eagle eye Krist strike again! 😱

packages/superset-ui-control-utils/src/types.ts
ktmud: It's not actually a valid property for `ControlPanelState`, judging by the code in `superset-frontend` (I did `console.log` on when `mapStateToProps` is called to debug). Removing it doesn't break any current type checking so I assumed it was good.

packages/superset-ui-control-utils/src/types.ts
ktmud: Yes, it is a required field based on source code in superset-frontend. It's basically a mapping of the states of all current controls.

packages/superset-ui-control-utils/src/types.ts
ktmud: It is expanded later in more specific types. `GeneralControlConfig` is basically the base interface.

packages/superset-ui-control-utils/src/types.ts
ktmud: `ReactNode` also includes `ReactText` which contains `string`. I wanted `string` to be exclusive for `SharedControlAlias`, so just by `if (typeof controlItem !== 'string')`, TypeScript will be able to infer what the remaining types `controlItem` could be.

packages/superset-ui-control-utils/src/types.ts
ktmud: Yes, it's just changing order. `controlSetRows` is a larger object. Kind of feel it should be at the bottom because of the "weight".

packages/superset-ui-control-utils/src/types.ts
kristw: Then do we need this field? Perhaps can just remove if this is only the base interface?

packages/superset-ui-control-utils/src/types.ts
kristw: got it.

packages/superset-ui-control-utils/src/types.ts
kristw: cool

packages/superset-ui-control-utils/src/types.ts
kristw: ic.

packages/superset-ui-control-utils/src/types.ts
kristw: Also, this is broader question. I often seen `type` field being string literal or something used as differentiator of union type. 
* Curious how is this `type` field used? 
* What is `internalControlType`? Is it a string?

packages/superset-ui-control-utils/src/types.ts
kristw: perhaps can add comment in code. in case someone stop by and wanting to change it to `ReactNode` without knowing the context

packages/superset-ui-control-utils/src/types.ts
ktmud: `InternalControlType` is a union of strings, basically the name of all the controls seeing in [controls/index.js](https://github.com/apache/incubator-superset/blob/291306392443a5a0d0e2ee0cc4a95d37c56d4589/superset-frontend/src/explore/components/controls/index.js#L73).

It is used make sure users are using known control types when passing string literals.

ktmud: I'm creating separate types instead of unions in the general type because I want TypeScript be able to infer requirements for other props based on what's being specified for `type`.

Also wanted to make sure `type` a required field for `GeneralControlConfig` so I think I'm going to keep it...

packages/superset-ui-control-utils/src/types.ts
ktmud: good call, will do

packages/superset-ui-control-utils/src/types.ts
kristw: I don't mean making `type` a union type. Let me clarify. Usually I do something like this. (also not sure if your "base interface" and my "base interface" mean the same thing)

```ts
interface BaseHero {
  attack: number;
  defense: number;
  speed: number;
  HP: number
} 

export interface Barbarian extends BaseHero {
  type: 'barbarian';
  fury: number;
}

export interface Sorceress extends BaseHero {
  type: 'sorceress';
  mana: number;
}

export type Hero = Barbarian | Sorceress;
```

ktmud: Yeah, the idea is the same. Maybe I can rename `GeneralControlConfig` to `BaseControlConfig` so people would no not to import it directly, then it feels more OK to remove `type`.

kristw: Do we need to `export` the `BaseControlConfig`?

ktmud: Probably not

packages/superset-ui-control-utils/src/shared-controls.tsx
```diff
@@ -111,8 +111,8 @@ const groupByControl: ControlConfig = {
   default: [],
   includeTime: false,
   description: t('One or many controls to group by'),
-  optionRenderer: (c: ColumnMeta) => <ColumnOption showType column={c} />,
-  valueRenderer: (c: ColumnMeta) => <ColumnOption column={c} />,
+  optionRenderer: c => <ColumnOption showType column={c} />,
+  valueRenderer: c => <ColumnOption column={c} />,
   valueKey: 'column_name',
   allowAll: true,
   filterOption: (opt: ColumnMeta, text: string) =>
```
rusackas, line 118: @ktmud It seems that this code hasn't been updated with the changes from [this commit](https://github.com/preset-io/incubator-superset/commit/81ab8dd8b46f7291171e674cd48d498c1efde409). We can update that as a fast-follow to this PR, unless you have the wherewithall to update (and properly type) it in this PR.

packages/superset-ui-control-utils/src/index.ts
```diff
@@ -2,16 +2,20 @@ import * as constantsModule from './constants';
 import * as sharedControlsModule from './shared-controls';
 import * as sectionModules from './sections';
 
+// explore all available shared controls
+export { default as sharedControls } from './shared-controls';
```
rusackas, line 6: THANK YOU. I happen to be in need of this presently, so this has my full support :D

packages/superset-ui-control-utils/src/shared-controls.tsx
ktmud: I'll slip in.

packages/superset-ui-control-utils/src/types.ts
```diff
@@ -198,21 +193,44 @@ type SelectControlType =
...
+ * Custom controls
+ * --------------------------------------------- */
+export type CustomComponentControlConfig<P = unknown> = BaseControlConfig & {
+  type: InternalControlType | React.ComponentType<P>;
+} & Omit<P, 'onChange' | 'hovered'>; // two run-time properties from superset-frontend/src/explore/components/Control.jsx
+
+// Catch-all ControlConfig
 export type ControlConfig<T extends SelectOption = SelectOption> =
-  | GeneralControlConfig
+  | BaseControlConfig
```
kristw, line 232: Do I understand correctly that a control config is either `SharedControlConfig` or `CustomComponentControlConfig`? 
Should `BaseControlConfig` then be removed from this catch-all? 

```ts
export type ControlConfig<
  T extends InternalControlType = InternalControlType
  O extends SelectOption = SelectOption
> = 
  | SharedControlConfig<T, O>
  | CustomComponentConfig;
```

ktmud: True, this needs an update.

BTW, got this error when removed `BaseControlConfig` from the export:

```
Default export of the module has or is using private name 'BaseControlConfig'.
```

It seems typescript requires you to export all internal types used by a default export.

packages/superset-ui-control-utils/src/types.ts
ktmud: Actually `type: unknown` is also needed, otherwise `BaseControlConfig & { type: xxx }` will report type incompatible error.

I made `BaseControlConfig` a generic type instead.

PR #615, opened 2020-06-18
fix: pointing nvd3 package to nvd3-fork

mistercrunch:
`nvd3.multiChart` did not offer the `xScale` property the same way that many other `nvd3` charts do and that was causing the xAxis to look off. 

I did some work to fork and publish [nvd3-fork](https://github.com/preset-io/nvd3) to npm from the `master` branch, where the bug was fixed years ago.

As a good citizen I offered help here https://github.com/novus/nvd3/issues/2222, but without enough movement decided to fork for now.

On npm: [nvd3-fork](https://www.npmjs.com/package/nvd3-fork)

Longer term strategy is to move to something like eCharts, but in the meantime we have a fork where we can fix bugs and such

## before 
Screen Shot 2020-06-17 at 9 30 24 PM

## after
Screen Shot 2020-06-17 at 9 53 50 PM


rusackas:
Or, perhaps, `nvd3-superset` under the `apache-superset` org if this is more of a bespoke thing to maintain for the oss offering. But whatevs... it works, and it's merged :shipit:

mistercrunch:
Oh oops, missed that comment and merged. Could have been `nvd3-fork-superset`, but might as well own it. "USE THE FORK LUKE!"

mistercrunch:
Pushed the fork to apache-superset https://github.com/apache-superset/nvd3#this-is-a-fork

ktmud:
Another idea is to just publish via GitHub SHA, so no need to worry about publish accesses.

```
  "nvd3": "apache-superset/nvd3#ed5b2c8"
```

mistercrunch:
Yeah I wanted for the fork to be out/published to see if anyone rallies around the fork.

etr2460:
@mistercrunch, I suspect this introduced a regression in tooltips on nvd3 charts, see here: https://github.com/apache-superset/superset-ui/issues/650

mistercrunch:
Oh no, just saw this. Looks like the regression has been addressed (?)

PR #623, opened 2020-06-19
feat(plugin-chart-table): rewrite with react-table

ktmud:
🏆 Enhancements

Rewrite the whole table chart with `react-table`, replacing `@superset-ui/plugin-chart-table` previously implemented with [lunar](https://github.com/airbnb/lunar). This package will eventually replace `@superset-ui/legacy-plugin-chart-table`. It is a drop-in replacement for the legacy table chart with some minor performance and UX improvements. Most noticeably:

1. Enabling CSS via `style` attribute for common HTML tags in cell values
2. Change default pagination to NULL, which means no pagination when number of cells < 10,000 and automatically set page size to 200 when it's above that.

This PR merges the already reviewed feature branch (#556 and #584) to master.

Example: https://superset-ui-git-react-table.superset.vercel.app/?path=/story/legacy-chart-plugins-legacy-plugin-chart-table--big-table


plugins/plugin-chart-table/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+import controlPanel from './controlPanel';
+
+export * from './types';
+
+const metadata = new ChartMetadata({
+  canBeAnnotationTypes: ['EVENT', 'INTERVAL'],
+  description: '',
+  name: t('Table'),
+  thumbnail,
+  useLegacyApi: true,
```
villebro, line 32: Do you think it would be possible to use the new API for this? If there are any features that are missing I'm sure we could add them. Migrating the table viz to the new endpoint would be a really big step forward, as it's one of the more complex plugins in `viz.py`

ktmud: I took a dab earlier but it feels more complicated than I thought. I'd prefer to merge the frontend change first, then replace the API. The API migration seems to worth its own work as there might be need to change the query client as well.

PR #630, opened 2020-06-25
chore: migrate time table chart from superset-frontend

pkdotson:
💔 Breaking Changes

🏆 Enhancements
This MR will add time-table from superset to superset-ui with add types using tyepsecript. I tested these change using symlink.
📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/legacy-plugin-chart-time-table/src/controlPanel.ts
```diff
@@ -0,0 +1,73 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { t } from '@superset-ui/translation';
+import { validateNonEmpty } from '@superset-ui/validator';
+
+export default {
```
ktmud, line 22: Could use `import { ControlPanelConfig } from '@superset-ui/chart-controls'` for safer typing.

E.g. https://github.com/apache-superset/superset-ui/blob/d6da661a11094a9895c76c6faff6692e08ad8b16/plugins/plugin-chart-table/src/controlPanel.tsx#L125-L130

plugins/legacy-plugin-chart-time-table/src/transformProps.ts
```diff
@@ -0,0 +1,87 @@
...
+  groupby: string[];
+  metrics: Array<object>;
+  url: string;
+  columnCollection: Array<object> | [];
+}
+
+interface DataObj {
+  records: DataRecord[];
+  columns: string[];
+}
```
ktmud, line 31: Nit: can just do

```
interface QueryData {
  data: {
    records: DataRecord[];
    columns: string[];
  };
}
```

if `DataObj` is not going to be reused.

plugins/legacy-plugin-chart-time-table/types/external.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.png';
+declare module '@data-ui/sparkline';
+declare module 'reactable-arc'
+declare module 'mustache';
```
ktmud, line 4: Nit: missing new line (Prettier didn't catch this because `*.d.ts` is in `.prettierignore` -- maybe we should remove it).

plugins/legacy-plugin-chart-time-table/src/TimeTable.less
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+ .time-table {
+    overflow: auto;
+  }
```
ktmud, line 21: 1. I don't think `superset-ui` handles `.less` right now.
2. Can probably convert this to styled components.

plugins/legacy-plugin-chart-time-table/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "bugs": {
+    "url": "https://github.com/apache-superset/superset-ui/issues"
+  },
+  "homepage": "https://github.com/apache-superset/superset-ui#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@data-ui/sparkline": "^0.0.84",
+    "@superset-ui/chart-controls": "^0.14.3",
```
ktmud, line 32: Let's follow the convention and make `@superset-ui` packages `peerDependencies`.

plugins/legacy-plugin-chart-time-table/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "@superset-ui/dimension": "^0.13.21",
+    "@superset-ui/number-format": "^0.13.21",
+    "@superset-ui/time-format": "^0.13.22",
```
ktmud, line 35: The versions also need an update.

plugins/legacy-plugin-chart-time-table/src/transformProps.ts
```diff
@@ -0,0 +1,87 @@
...
+export default function transformProps(chartProps: TableChartProps) {
+  const { height, datasource, formData, queryData } = chartProps;
+  const { columnCollection, groupby, metrics, url } = formData;
+  const { records, columns } = queryData.data;
+  const isGroupBy = groupby?.length > 0;
+
+  // When there is a "group by",
+  // each row in the table is a database column
+  // Otherwise,``
```
ktmud, line 53:
```suggestion
  // Otherwise each row in the table is a metric
```

ktmud:
Can you maybe also add a storybook story if you have time? Not a blocker though.

mistercrunch:
Side note: there's an issue with the headers on these tables that seem to be intricately related to some bug with `reactable` and hot loader

pkdotson:
> Can you maybe also add a storybook story if you have time? Not a blocker though.

Yah i'm going to be adding that next after this is merged in!

PR #632, opened 2020-06-25
feat(style): add opacity and zIndex variables to theme

lilykuang:
🏆 Enhancements

### DIFF
--- a/packages/superset-ui-style/src/index.ts
+++ b/packages/superset-ui-style/src/index.ts
@@ -90,6 +90,12 @@ const defaultTheme = {
       light2: '#EFF8FE',
     },
   },
+  opacity: {
+    light: '10%',
+    mediumLight: '35%',
+    mediumHeavy: '60%',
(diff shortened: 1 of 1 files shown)


packages/superset-ui-style/src/index.ts
```diff
@@ -111,6 +117,12 @@ const defaultTheme = {
       xxl: 28,
     },
   },
+  zIndex: {
+    aboveDashboardCharts: 10,
+    dropdown: 11,
+    max: 3000,
+  },
+  timingNormal: 0.3,
```
suddjian, line 125: Maybe this could have a more descriptive name? I'm not sure what `timingNormal` means.

ktmud: If this is use for CSS animation, maybe use something like `transitionDurationDefault`.

lilykuang: named it `transitionTiming`

PR #634, opened 2020-06-26
feat(connection): easier API for json payload

ktmud:
🏆 Enhancements

Provide an easier API to send JSON requests. So instead of

```js
    await SupersetClient.post({
      url: '/api/v1/chart/data',
      headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify({
        query_context: ...
      }),
    });
```

users can just do

```js
    await SupersetClient.post({
      url: '/api/v1/chart/data',
      jsonPayload: {
        query_context: ...
      },
    });
```

Also make `postPayload` (and `jsonPayload`) accept strings, which will be automatically parsed as a JSON object.

### Test Plans

- Test locally with storybook
- Added an additional unit test

### DIFF
--- a/packages/superset-ui-connection/src/types.ts
+++ b/packages/superset-ui-connection/src/types.ts
@@ -12,21 +12,26 @@ export type FetchRetryOptions = {
 export type Headers = { [k: string]: string };
 export type Host = string;
 
-// More strict generic JSON types
-export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
-export type JSONPrimitive = string | number | boolean | null;
-export type JSONArray = JSONValue[];
-export type JSONObject = { [member: string]: JSONValue };
+export type JsonPrimitive = string | number | boolean | null;
+/**
+ * More strict JSON value types. If this fails to satisfy TypeScript when using
+ * as function arguments, use `JsonObject` instead. `StrictJsonObject` helps make
+ * sure all values are plain objects, but it does not accept specific types when
+ * used as function arguments.
+ * (Ref: https://github.com/microsoft/TypeScript/issues/15300).
+ */
+export type StrictJsonValue = JsonPrimitive | StrictJsonObject | StrictJsonArray;
+export type StrictJsonArray = StrictJsonValue[];
+export type StrictJsonObject = { [member: string]: StrictJsonValue };
 
 export type JsonValue = JsonPrimitive | JsonObject | JsonArray;
-export type JsonPrimitive = JSONPrimitive;
 export type JsonArray = JsonValue[];
-// `JSONObject` does not accept specific types when used as function arguments,
-// so we had to employ `any` (Ref: https://github.com/microsoft/TypeScript/issues/15300).
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type JsonObject = { [member: string]: any };
 
-// Post form or JSON payload, if string, will parse with JSON.parse
+/**
+ * Post form or JSON payload, if string, will parse with JSON.parse
+ */
 export type Payload = JsonObject | string;
 
 export type Method = RequestInit['method'];
--- a/packages/superset-ui-connection/src/types.ts
+++ b/packages/superset-ui-connection/src/types.ts
@@ -11,16 +11,20 @@ export type FetchRetryOptions = {
 };
 export type Headers = { [k: string]: string };
 export type Host = string;
+
+// More strict generic JSON types
 export type JSONPrimitive = string | number | boolean | null;
 export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
 export type JSONObject = { [member: string]: JSONValue };
 export type JSONArray = JSONValue[];
 
-// `JSONObject` could not accept more specific types, so we had to employ any.
-// see: https://github.com/microsoft/TypeScript/issues/15300
+// `JSONObject` does not accept specific types when used as function arguments,
+// so we had to employ `any` (Ref: https://github.com/microsoft/TypeScript/issues/15300).
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type JsonObject = { [member: string]: any };
-export type PostPayload = JsonObject | string;
+
+// Post form or JSON payload, if string, will parse with JSON.parse
+export type Payload = JsonObject | string;
 
 export type Method = RequestInit['method'];
 export type Mode = RequestInit['mode'];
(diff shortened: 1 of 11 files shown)


ktmud:
cc @lilykuang @nytai

packages/superset-ui-connection/src/types.ts
```diff
@@ -11,11 +11,18 @@ export type FetchRetryOptions = {
...
+export type JSONPrimitive = string | number | boolean | null;
+export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
+export type JSONObject = { [member: string]: JSONValue };
+export type JSONArray = JSONValue[];
+
+// `JSONObject` could not accept more specific types, so we had to employ any.
+// see: https://github.com/microsoft/TypeScript/issues/15300
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Json = { [k: string]: any };
+export type JsonObject = { [member: string]: any };
```
kristw, line 22: (optional) can also do `Record<string, any>`

packages/superset-ui-connection/src/types.ts
```diff
@@ -49,25 +43,21 @@ export interface RequestBase {
...
-  endpoint?: undefined;
+  cache?: Cache;
+  redirect?: Redirect;
 }
 
-export type RequestConfig = RequestWithEndpoint | RequestWithUrl;
-
 export interface JsonTextResponse {
-  json?: Json;
+  json?: JsonObject;
```
kristw, line 60: Hmmm `JsonObject` vs `JSONObject` seem easy to make a mistake.

packages/superset-ui-query/src/api/v1/postChartData.ts
```diff
@@ -7,17 +7,15 @@ export interface Params extends BaseParams {
   queryContext: QueryContext;
 }
 
-export default function postChartData({
+export default async function postChartData({
```
kristw, line 10: This is broader js question. 
What is the advantage of `async` vs returning promise? I sometimes not sure what do I get from converting to async function, which sometimes ask for the `regenerator-runtime` when transpile.

packages/superset-ui-connection/src/types.ts
ktmud:
```ts
export type JSONValue = JSONPrimitive | JSONObject | JSONArray;
export type JSONObject = Record<string, JSONValue>;
```

This will generate following error:

```
Type alias 'JSONObject' circularly references itself.  ts(2456)
```

packages/superset-ui-connection/src/types.ts
ktmud: I think it's fine. In an idea world, we shouldn't need `JsonObject` if microsoft/TypeScript#15300 was fixed so these two should be interchangeable for most cases---so it's fine for users to use either. I just added aliases for `JsonValue`, `JsonArray`, etc., to make it more complete.

We can probably remove `JSONObject` but that would also remove the option of stricter typing.

packages/superset-ui-query/src/api/v1/postChartData.ts
ktmud: `async/await` reduces parentheses/indent levels and makes the code somewhat more straightforward IMO.

The missing `regenerator-runtime` error is annoying but there is a known solution for that so I can live with it.

kristw: Let's not do it in this PR yet. 
To adopt this, better make all the api wrapper functions `/api/**/*.ts` adopt it for consistency.

packages/superset-ui-connection/src/types.ts
kristw: > I just added aliases for JsonValue, JsonArray, etc., to make it more complete.
Are they meant to be used anywhere? If not perhaps we can remove?

For `JsonObject`, we could rename that to `PlainObject`?

ktmud: It's different than `PlainObject`. The values shouldn't really be `any` if we had a choice (e.g. `new Person()` may be valid for `PlainObject`, but isn't valid for a JSON object parsed from `JSON.parse`). I think it makes sense to have a special name for this in case microsoft/TypeScript#15300 got fixed.

It also goes better with type names like `JsonResponse`, `TextResponse`, better.

packages/superset-ui-query/src/api/v1/postChartData.ts
ktmud: I changed all of those in #635 . I'd rather not have to revert these changes as there is already a mix of both in use.

packages/superset-ui-connection/src/types.ts
kristw: Here is my ask. I am OK with keeping the two types but would appreciate help reducing the confusion when to use `JSONObject` and when to use `JsonObject`. 
* The current names are not helpful. 
  * Maybe `JSONObject => StrictJsonObject`
  * or `JSONObject => JsonObject` and `JsonObject => RelaxedJsonObject` ?
* Add `/** */` comment for the types so autocomplete can help explain. such as

```
/**
 * If this fails to satisfy TypeScript when using as function arguments, use `RelaxedJsonObject` instead. 
 * `JsonObject` does not accept specific types when used as function arguments,
 * (Ref: https://github.com/microsoft/TypeScript/issues/15300).
 */
type JsonObject
/** 
 * This is less strict type of `JsonObject`. If that type fails to satisfy TypeScript when using as function arguments, use this type instead. 
 * `JsonObject` does not accept specific types when used as function arguments,
 * (Ref: https://github.com/microsoft/TypeScript/issues/15300).
 */
type RelaxedJsonObject
```

packages/superset-ui-query/src/api/v1/postChartData.ts
kristw: ok

packages/superset-ui-connection/src/types.ts
ktmud: Great suggestions! Happy to update!

packages/superset-ui-connection/src/types.ts
```diff
@@ -11,11 +11,30 @@ export type FetchRetryOptions = {
 };
 export type Headers = { [k: string]: string };
 export type Host = string;
+
+export type JsonPrimitive = string | number | boolean | null;
+/**
```
kristw, line 16: You may want to put this before the `StrictJsonObject` line. 
Currently, it will only appear in auto-complete of `StrictJsonValue`.

PR #636, opened 2020-06-29
feat: updated viz plugin generator

rusackas:
💔 Breaking Changes

🏆 Enhancements

This updates the viz plugin generator with extensive comments, upgrades to the latest architecture/patterns, and various yeoman options (react class/function components, timseries/non-timseries, with/without badges).

The point of this is to offer up an updated, and more extensible Hello World plugin, to enhance the developer experience when building new viz. More PRs will be coming to make this more "deluxe" but this provides a much better (and better documented) starting point for viz developers.

Four draft PRs show the permutations of this generator. 

| | Timeseries | Non-Timeseries | 
| --- | --- | --- |
| **Class** | #639 | #637 |
| **Function** | #640 | #638 |

📜 Documentation

🐛 Bug Fix

🏠 Internal


packages/generator-superset/generators/plugin-chart/templates/src/MyChart.erb
```diff
@@ -16,25 +17,122 @@
...
+
+    h3 {
+      /* You can use your props to control CSS! */
+      font-size: ${({ theme }) => theme.typography.sizes[props.headerFontSize]};
+      font-weight: ${({ theme }) => theme.typography.weights[props.boldText ? 'bold' : 'normal']};
+    }
+  `;
+
+  return (
+    <Wrapper ref={rootElem}>
```
ktmud, line 133: 1. Does the `Wrapper` need `rootElem`?
2. It should probably be placed outside of the functional component otherwise a new styled component is created every time the chart component renders. Not sure about the performance impact of this, but that doesn't sound optimal.

I also like a convention I saw elsewhere which names styled divs `Styles`:

```jsx
const Styles = styled.div`
  h3 {
    /* You can use your props to control CSS! */
    font-size: ${({ theme, headerFontSize }) => theme.typography.sizes[headerFontSize]};
    font-weight: ${({ theme, boldText }) => theme.typography.weights[boldText ? 'bold' : 'normal']};
  }
`;

function MyComponent(props: { headerFontSize: string; boldText: boolean; }) {

  return (
    <Styles {...props}>
      ...
    </Styles>
  );
}
```

What do you think?

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { t } from '@superset-ui/translation';
+import { validateNonEmpty } from '@superset-ui/validator';
```
ktmud, line 20: Can utilize `ControlPanelConfig` here:

https://github.com/apache-superset/superset-ui/blob/d6da661a11094a9895c76c6faff6692e08ad8b16/plugins/plugin-chart-table/src/controlPanel.tsx#L125-L130

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * It is also possible to define custom controls by importing the
+   * necessary dependencies and overriding the default parameters, which
+   * can then be placed in the `controlSetRows` section
+   * of the `Query` section instead of a predefined control.
+   *
+   * import { validateNonEmpty } from '@superset-ui/validator';
+   * import { internalSharedControls } from '@superset-ui/control-utils';
+   *
+   * const myControl = {
```
ktmud, line 57:
```suggestion
   * const myControl: ControlConfig<'SelectControl'> = {
```

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+   * `groupby`, which automatically appends the values from the
+   * `series` control to the `groupby` control when the query is generated.
+   *
+   * It is also possible to define custom controls by importing the
+   * necessary dependencies and overriding the default parameters, which
+   * can then be placed in the `controlSetRows` section
+   * of the `Query` section instead of a predefined control.
+   *
+   * import { validateNonEmpty } from '@superset-ui/validator';
+   * import { internalSharedControls } from '@superset-ui/control-utils';
```
ktmud, line 55:
```suggestion
   * import { sharedControls, ControlConfig } from '@superset-ui/chart-control'; 
```

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+   * - validateNumber: must be an intger or decimal value
+   */
+
+  // For control input types, see: superset-frontend/src/explore/components/controls/index.js
+<%if (chartType === 'timeseries') { %>  requiresTime: true,<% } %>
+  controlPanelSections: [
+    {
+      label: t('Query'),
+      expanded: true,
+      controlSetRows: [['groupby'], ['metrics'], ['adhoc_filters'], ['row_limit', null]],
```
ktmud, line 100: ESLint will probably reformat this

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+        ],
+        [
+          {
+            name: 'header_font_size',
+            config: {
+              type: 'SelectControl',
+              label: t('Font Size'),
+              default: 'xl',
+              choices: [
+                ['xxs', 'xx-small'],
```
ktmud, line 138: Can add a comment

```suggestion
                // [value, label]
                ['xxs', 'xx-small'],
```

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
```diff
@@ -0,0 +1,174 @@
...
+                ['xs', 'x-small'],
+                ['s', 'small'],
+                ['m', 'medium'],
+                ['l', 'large'],
+                ['xl', 'x-large'],
+                ['xxl', 'xx-large'],
+              ],
+              renderTrigger: true,
+              // ^ this makes it apply instantaneously, without triggering a "run query" button
```
ktmud, line 147: Can move this line to where `renderTrigger` first appears.

packages/generator-superset/generators/plugin-chart/templates/src/plugin/transformProps.erb
```diff
@@ -0,0 +1,72 @@
...
+   */
+  const { width, height, formData, queryData } = chartProps;
+  const { data } = queryData;
+
+  console.log('formData via TransformProps.ts', formData);
+
+  return {
+    width,
+    height,
+<%if (chartType === 'timeseries') { %>    // eslint-disable-next-line @typescript-eslint/no-unsafe-call,@typescript-eslint/no-unsafe-member-access
```
ktmud, line 60: Can also do this above:

```ts
import { DataRecord } from '@superset-ui/chart';

<%if (chartType === 'timeseries') { %>
interface <%= packageLabel %>Datum extends DataRecord {
  __timestamp: string | number | null | Date;
}
<% } else { %>
type <%= packageLabel %>Datum = DataRecord;
<% } %>


export default function transformProps(chartProps: ChartProps) {
  ...
  const data = queryData.data as <%= packageLabel %>Datum[];
```

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
rusackas: Interestingly, it seems not to. At least, it doesn't come through that way in [the PR](https://github.com/apache-superset/superset-ui/pull/639/files#diff-ffe3ea995e74fc17c57cd00fda38e6c6R100)(s). ¯\\\_(ツ)_/¯

ktmud: I stand corrected

packages/generator-superset/generators/plugin-chart/templates/README.erb
```diff
@@ -1,7 +1,7 @@
 ## @superset-ui/plugin-chart-<%= packageName %>
```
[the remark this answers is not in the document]
rusackas: For the time being, the resulting plugin is more or less best suited to use in `superset-ui` so I think we should leave it for the time being. We're still working on other features that will make plugins better suited for external use, and when those are ready, I think it would make sense to add another choice in the Yeoman generator process, and name the package accordingly.

packages/generator-superset/generators/plugin-chart/templates/src/MyChart.erb
rusackas: Perfectly OK with the `Styled` name. It's more explicit. `Wrapper` felt a little dirty, since in a way, _everything's_ a wrapper.

As for whether the plugin needs the `ref`, that's just part of the example scaffolding so people can more easily grab a hold of the DOM element and go crazy on it with D3 (or whatever).

Now, as for _where_ to put that ref in the JSX, I'm not quite sure... I can _certainly_ take it up a layer, as you suggest, like so:
```
return (
  <div ref={rootElem}
    <Style>
      ...more stuff
    </Style>
  </div>
)
```
This would address your styled component instantiation concern. _However_, I'm not sure that's actually a problem... When I put two of these on a dash, they share the same class, which makes me think it's safe enough. Am I mistaken?

Pasted_Image_6_30_20__9_43_PM

Pasted_Image_6_30_20__9_46_PM

packages/generator-superset/generators/plugin-chart/templates/src/plugin/transformProps.erb
rusackas: This is AWESOME. Thank you!

packages/generator-superset/generators/plugin-chart/templates/src/plugin/controlPanel.erb
rusackas: Thanks, this is great! Done :D

packages/generator-superset/generators/plugin-chart/templates/src/MyChart.erb
rusackas: ok, n/m, I think I understand where you're coming from... I'll externalize the Styled component.

ktmud: Yeah, emotion probably has proactive caching that makes this a non-issue, but externalizing it feels like a better practice because you can at least be sure you don't create a new styled component every time the chart re-renders---even though they may point to the same cache.

It will also help you 1) limit which props are available to styles 2) make it easier to extract it to another file when it becomes too large.

rusackas:
All variant PRs are passing, and comments (hopefully) addressed. Will merge in the morning if nobody rescinds approvals ;) Thanks for all your help @ktmud

PR #641, opened 2020-06-29
chore(plugin-chart-parallel-coordinates): convert para coords chart to emotion styling

pkdotson:
💔 Breaking Changes

🏆 Enhancements
Convert parallel coordinates chart to emotion styling.

📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/legacy-plugin-chart-parallel-coordinates/package.json
```diff
@@ -28,8 +28,10 @@
     "access": "public"
   },
   "dependencies": {
+    "@superset-ui/style": "^0.14.3",
     "d3": "^3.5.17",
-    "prop-types": "^15.6.2"
+    "prop-types": "^15.7.2",
+    "react": "^16.13.1"
```
ktmud, line 34: Same issue with other plugins.

plugins/legacy-plugin-chart-parallel-coordinates/src/index.js
```diff
@@ -33,7 +33,7 @@ const metadata = new ChartMetadata({
 export default class ParallelCoordinatesChartPlugin extends ChartPlugin {
   constructor() {
     super({
-      loadChart: () => import('./ReactParallelCoordinates.js'),
+      loadChart: () => import('./ReactParallelCoordinates.jsx'),
```
ktmud, line 36: Same here, remove `.jsx`

plugins/legacy-plugin-chart-parallel-coordinates/package.json
ktmud: react and prop-types should also be peerDependencies.

kristw: react is peer but prop-types is dep

kristw: Is it really resolved? The code is still the same.

PR #635, opened 2020-06-28
feat(connection): optimize typing and API for SupersetClient

ktmud:
💔 Breaking Changes

🏆 Enhancements

🏠 Internal

- <strike>Introduce a `makeApi` generator to easily add API endpoints</strike> moved to #642
- More robust typing and error handling for `SupersetClient`
- Add a `baseUrl` option for SupersetClientClass (we should probably deprecate the `host` and `protocol` option in the future)
- Optimize storybook demos

Based on and merges into #634

PR #642, opened 2020-06-29
feat(query): add makeApi API generator

ktmud:
🏆 Enhancements

Introduce a `makeApi` API generator to easily add typed query methods for new (and existing) API endpoints.

Based on and merges into #635.


ktmud:
There are some more changes coming. Still WIP.

PR #651, opened 2020-06-30
chore(chart-sankey): use emotion styling typography

pkdotson:
💔 Breaking Changes

🏆 Enhancements
before
 Superset

after
© Superset

📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/legacy-plugin-chart-sankey/package.json
```diff
@@ -30,9 +30,11 @@
   "dependencies": {
     "d3": "^3.5.17",
     "d3-sankey": "^0.4.2",
-    "prop-types": "^15.6.2"
+    "prop-types": "^15.6.2",
+    "react": "^16.13.1"
```
ktmud, line 34: both should be peerDependencies

plugins/legacy-plugin-chart-sankey/src/index.js
```diff
@@ -33,7 +33,7 @@ const metadata = new ChartMetadata({
 export default class SankeyChartPlugin extends ChartPlugin {
   constructor() {
     super({
-      loadChart: () => import('./ReactSankey.js'),
+      loadChart: () => import('./ReactSankey.jsx'),
```
ktmud, line 36: I don't think the suffix is necessary.

plugins/legacy-plugin-chart-sankey/src/ReactSankey.jsx
```diff
@@ -0,0 +1,73 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React from 'react';
+import { reactify } from '@superset-ui/chart';
+import styled from '@superset-ui/style';
+import PropTypes from 'prop-types';
+import Component from './Sankey';
+
+const ReactComponent = reactify(Component);
```
ktmud, line 25: Maybe make these names more explicit:

```
import Sankey from './Sankey';

const ReactSankey = reactify(Sankey);
```

PR #652, opened 2020-06-30
chore(chart-worldmap): use emotion style

pkdotson:
💔 Breaking Changes

🏆 Enhancements
before 
• Superset # Security

after
Screen Shot 2020-06-30 at 12 07 35 PM
📜 Documentation

🐛 Bug Fix

🏠 Internal


plugins/legacy-plugin-chart-world-map/package.json
```diff
@@ -30,12 +30,14 @@
   "dependencies": {
     "d3": "^3.5.17",
     "datamaps": "^0.5.8",
-    "prop-types": "^15.6.2"
+    "prop-types": "^15.6.2",
+    "react": "^16.13.1"
```
ktmud, line 34: `react` should be peerDependencies

plugins/legacy-plugin-chart-world-map/package.json
```diff
@@ -30,12 +30,14 @@
   "dependencies": {
     "d3": "^3.5.17",
     "datamaps": "^0.5.8",
-    "prop-types": "^15.6.2"
+    "prop-types": "^15.6.2",
```
ktmud, line 33: seems invalid `package.json`.

plugins/legacy-plugin-chart-world-map/src/ReactWorldMap.jsx
```diff
@@ -16,7 +16,31 @@
...
+  className: PropTypes.string.isRequired,
+};
+
+export default styled(WorldMapComponent)`
+  .superset-legacy-chart-world-map {
+    position: relative;
+    svg {
+      background-color: #feffff;
+    }
+  }
```
ktmud, line 45: Does styled component support less syntax already?

ktmud: never mind, got the answer myself

PR #660, opened 2020-07-01
feat: adding requiresTime type

rusackas:
💔 Breaking Changes

🏆 Enhancements

Adds typing for the requiresType property, on ControlPanelConfig, in the chart-controls pkg.

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -304,6 +304,7 @@ export interface ControlPanelSectionConfig {
 }
 
 export interface ControlPanelConfig {
+  requiresTime?: boolean;
   controlPanelSections: ControlPanelSectionConfig[];
   controlOverrides?: ControlOverrides;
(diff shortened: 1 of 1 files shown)


ktmud:
I see this property defined for a lot of charts, but couldn't find it in the actual code---bother frontend or backend. Could you point me to the code piece where it was used if there's any? (Otherwise maybe cleanup?)

rusackas:
You're right, I'm not seeing its purpose either. Maybe @villebro has some insight here.

rusackas:
I think you may be right... it _seems_ like the plugin works just fine without the `requireTime` param. I'll just leave this PR hanging out here for a little while, press on with the generator PR, and maybe we can go through and I'll open a PR deleting `requireTime` everywhere, to see how that goes.

villebro:
I've never seen it being used anywhere, so AFAIK it's probably ok to remove it.

ktmud:
Per discussion, removed it instead

PR #661, opened 2020-07-01
fix(plugin-chart-table): anchor should keep href

ktmud:
🐛 Bug Fix

Re-enable table chart to create anchor element with link.

## Test Plan

Storybook

### DIFF
--- a/plugins/plugin-chart-table/src/utils/formatValue.ts
+++ b/plugins/plugin-chart-table/src/utils/formatValue.ts
@@ -23,10 +23,10 @@ import { DataColumnMeta } from '../types';
 const xss = new FilterXSS({
   whiteList: {
     ...getDefaultWhiteList(),
-    span: ['style', 'title'],
-    div: ['style'],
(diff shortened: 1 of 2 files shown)


plugins/plugin-chart-table/src/utils/formatValue.ts
```diff
@@ -23,10 +23,10 @@ import { DataColumnMeta } from '../types';
...
   whiteList: {
     ...getDefaultWhiteList(),
-    span: ['style', 'title'],
-    div: ['style'],
-    a: ['style'],
-    img: ['style', 'src', 'alt', 'title', 'width', 'height'],
+    span: ['style', 'class', 'title'],
+    div: ['style', 'class'],
+    a: ['style', 'class', 'href', 'title', 'target'],
+    img: ['style', 'class', 'src', 'alt', 'title', 'width', 'height'],
```
ktmud, line 29: Let's also allow CSS classnames.

PR #663, opened 2020-07-02
fix(plugin-chart-table): formatting non-numeric metrics

ktmud:
🐛 Bug Fix

Non-numeric metrics shouldn't be passed to the numeric filter or "adaptive dates" time formatter.

### Test Plan

- Added storybook
- Tested in Superset
- Added Cypress tests in Superset: https://github.com/apache-superset/superset-ui/pull/663/files

#### Bug

image

#### Expected (after fix)
 
Snip20200702_17

### DIFF
--- a/plugins/plugin-chart-table/src/transformProps.ts
+++ b/plugins/plugin-chart-table/src/transformProps.ts
@@ -17,7 +17,7 @@
  * under the License.
  */
 import memoizeOne from 'memoize-one';
-import { DataRecord } from '@superset-ui/chart';
+import { DataRecord, DataRecordValue } from '@superset-ui/chart';
 import { QueryFormDataMetric } from '@superset-ui/query';
 import { getNumberFormatter, NumberFormats } from '@superset-ui/number-format';
 import {
@@ -31,6 +31,7 @@ import { TableChartProps, TableChartTransformedProps, DataType, DataColumnMeta }
 
 const { PERCENT_3_POINT } = NumberFormats;
 const TIME_COLUMN = '__timestamp';
+const toString = (x: DataRecordValue) => String(x);
 
 /**
  * Consolidate list of metrics to string, identified by its unique identifier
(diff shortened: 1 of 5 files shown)


plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -108,24 +121,34 @@ const processColumns = memoizeOne(function processColumns(props: TableChartProps
...
     const format = columnFormats?.[key];
     const isTime = isTimeType(key, records);
-    const isMetric = metricsSet.has(key);
+    // for the purpose of presentation, only numeric values are treated as metrics
+    const isMetric = metricsSet.has(key) && isNumeric(key, records);
     const isPercentMetric = percentMetricsSet.has(key);
     let dataType = DataType.Number; // TODO: get this from data source
     let formatter;
     if (isTime) {
       // Use ganularity for "Adaptive Formatting" (smart_date)
```
kristw, line 133: typo `g"r"a`

plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -47,25 +48,37 @@ function isTimeColumn(key: string) {
...
+    isTimeColumn(key) ||
+    data.some(x => {
+      const value = x[key];
+      return value instanceof Date || (typeof value === 'string' && REGEXP_DATETIME.test(value));
+    })
+  );
+}
+
+function isNumeric(key: string, data: DataRecord[] = []) {
+  return data.every(x => x[key] === null || x[key] === undefined || typeof x[key] === 'number');
```
kristw, line 64: Usually we do `typeof x[key] === 'undefined'` to be safe because `undefined` is not a reserved keyword and can be overwritten. 

This is one of the common questions in our codebase as the `typeof xxx === 'undefined'` is kinda verbose. The intention was to better be safe than sorry. 

cc @williaster who introduced me to this caveat. 

---

Good to know in ECMAScript 5.1+ `undefined` was made readonly. There are still small possibility of overwriting via local variable. Although that would be a harder accident to make. (ps. the author of this post seem to be quite mad at this syntax lol)
https://stackoverflow.com/questions/8783510/how-dangerous-is-it-in-javascript-really-to-assume-undefined-is-not-overwritte

ktmud: image

It seems it can't be overridden in Chrome at least. (Chrome just silently ignore your assignment.)

---

"if they did it would be reason enough to never use any code they wrote... right?" LOL on this line

ktmud: I'm basically fine with anything that ESLint doesn't complaint.

plugins/plugin-chart-table/src/transformProps.ts
ktmud: Fixed and pushed to master directly.

PR #667, opened 2020-07-06
feat: update timeout code to return new error format

etr2460:
🏆 Enhancements
Now that Superset's frontend has fallback logic for any parts not yet migrated to the new error format, we can update this safely, providing more details and context about the error

to: @ktmud @williaster

### DIFF
--- a/packages/superset-ui-connection/src/callApi/rejectAfterTimeout.ts
+++ b/packages/superset-ui-connection/src/callApi/rejectAfterTimeout.ts
@@ -4,7 +4,14 @@ export default function rejectAfterTimeout<T>(timeout: number) {
     setTimeout(
       () =>
         reject({
-          error: 'Request timed out',
+          errors: [
+            {
+              error_type: 'FRONTEND_TIMEOUT_ERROR',
+              extra: { timeout },
+              level: 'error',
+              message: 'Request timed out',
+            },
+          ],
           statusText: 'timeout',
         }),
       timeout,
--- a/packages/superset-ui-connection/test/callApi/callApiAndParseWithTimeout.test.ts
+++ b/packages/superset-ui-connection/test/callApi/callApiAndParseWithTimeout.test.ts
@@ -100,7 +100,14 @@ describe('callApiAndParseWithTimeout()', () => {
       } catch (error) {
         expect(fetchMock.calls(mockTimeoutUrl)).toHaveLength(1);
         expect(error).toEqual({
-          error: 'Request timed out',
+          errors: [
+            {
+              error_type: 'FRONTEND_TIMEOUT_ERROR',
+              extra: { timeout: 1 },
+              level: 'error',
+              message: 'Request timed out',
+            },
+          ],
           statusText: 'timeout',
         });
       }


packages/superset-ui-connection/src/callApi/rejectAfterTimeout.ts
```diff
@@ -4,7 +4,14 @@ export default function rejectAfterTimeout<T>(timeout: number) {
...
         reject({
-          error: 'Request timed out',
+          errors: [
+            {
+              error_type: 'FRONTEND_TIMEOUT_ERROR',
+              extra: { timeout },
+              level: 'error',
+              message: 'Request timed out',
+            },
+          ],
```
ktmud, line 14: I've made the frontend capable of [handling `errors: []`](https://github.com/ktmud/superset-ui/blob/superset-query/packages/superset-ui-query/src/api/v1/handleError.ts#L73) in  #666 . Once the error renderer is updated accordingly, maybe you can just return the single error without putting it in an array?

etr2460: I'm a bit confused why we're adding in the complexity of representing the error both as top level attributes within the object and as something nested within the `errors` attribute. Wouldn't it be easier to always rely on the existence of `errors` to know that this response has an error in the new structure? Also, thinking that an array with one element is easier to deal with than handling the different cases of an array with multiple elements vs. no array at all

ktmud: The support for both single error object and arrays has to be added because

1. We can’t refactor all backend API at the same time.
2. We also need to catch JS errors that are not an array.

For most use cases, there is only the need to display one API error even if the backend returns multiple (it’s kind of information overload if we pop up multiple alert toasts at the same time), so it makes sense for the frontend to consolidate errors into one object by default (all errors still accessible via `originalError` for those who need it).

Array checking is relatively cheap. I’m more concerned with error objects having different fields, which I think SIP-40 and SIP-41 has done a great deal improving it.

No strong feeling on whether the final error handler should return things in array or not, but two things are sure: 1. We do need an error handler at the frontend; 2. It doesn’t feel elegant to having to construct an array at a low level API when you can just `throw new Error()`. Creating an array here is akin to having a Python procedure call `json_error_response` with an array.

etr2460: We can't refactor all backend APIs at the same time, correct. So that's why we've been returning either an `error` (a string error message) or an `errors` array of SupersetError objects. Superset then pulls the first message from the `errors` array out into the `error` attribute if it doesn't exist, thus maintaining backwards compatibility until all instances of `error` are removed (see here: https://github.com/apache/incubator-superset/blob/master/superset-frontend/src/utils/getClientErrorObject.ts#L55-L59)

Is the intention to move the logic in `getClientErrorObject` over to superset-ui? At this point I think we have full backwards compatibility within Superset at least, so I still don't quite understand what #666 is adding in terms of error handling. It looks to me like it's stripping out the error_type and other fields that we need on the frontend to render the error properly

ktmud: Yes, you can say I'm moving that logic to superset-ui. The intention was to create a general API generator that handles both JS errors and different format of API errors and always returns the same error format. Other fields like `error_type` are always available in the `originalError` field. We can lift it to the top level of the `SupersetApiError` class if needed.

etr2460: Yeah, that makes a lot of sense. In that case, can we move towards all errors returning a payload of type `Payload` below? This structure should be supported everywhere by Superset, and it aligns with the structure of payloads the backend returns as well. I think the frontend should return JS errors in the same format as the backend API errors

```typescript
type SupersetError = {
  message: string;
  error_type: string; // Actually an enum in Superset's codebase, but it's always a string
  level: "info" | "warning" | "error";
  extra: Record<string, any> | null;
};

type Payload = {
  errors: SupersetError[];
}
```

ktmud: The [frontend error and payload types](https://github.com/apache-superset/superset-ui/pull/666/files#diff-711a2129f575a0c7e2a482fe7d6a9bacR48-R102) more or less look like what you described. I can move them closer by adding `error_type` and `extra` to the `SupersetApiError` class.

etr2460:
closing in favor of https://github.com/apache/incubator-superset/pull/10274

PR #669, opened 2020-07-06
fix(plugin-chart-table): sort and search time column

ktmud:
🐛 Bug Fix

1. Fix `getTime` undefined error when sorting a time column in table chart. Also refactors how datetime is handled for `react-table` search, sort and render methods.

   

2. Fix spacing issue for table pagination in dashboards
   
   **Before:**

   

   **After:**

   

## Test Plan 

Unit test in superset-ui and E2E test in superset-frontend.

cc @serenajiang @kristw

### DIFF
--- a/plugins/plugin-chart-table/src/transformProps.ts
+++ b/plugins/plugin-chart-table/src/transformProps.ts
@@ -24,6 +24,8 @@ import {
   getTimeFormatter,
   smartDateFormatter,
   getTimeFormatterForGranularity,
+  TimeFormatFunction,
+  TimeFormatter,
 } from '@superset-ui/time-format';
 
 import isEqualArray from './utils/isEqualArray';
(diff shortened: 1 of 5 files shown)


plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -24,14 +24,52 @@ import {
...
+    // `new Date()` is capable of handling null/boolean/undefined, etc;
+    super(value as string);
+
+    this.input = input;
+    this.formatter = formatter;
+  }
+
+  toString(): string {
+    if (this.formatter === String) {
+      return String(this.input);
```
etr2460, line 68: maybe this should be `return this.formatter(this.input);`?

plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -24,14 +24,52 @@ import {
...
+
+  constructor(
+    input: DataRecordValue,
+    { formatter = String, forceUTC = true }: { formatter?: TimeFormatFunction; forceUTC?: boolean },
+  ) {
+    let value = input;
+
+    // force UTC time for timestamps without a timezone
+    if (forceUTC && typeof value === 'string' && REGEXP_TIMESTAMP_NO_TIMEZONE.test(value)) {
+      value = `${value}Z`;
```
etr2460, line 55: is this correct? I thought you needed to replace the `T` with a `Z` to convert the string to UTC time

plugins/plugin-chart-table/src/transformProps.ts
ktmud: It's basically the same thing?

plugins/plugin-chart-table/src/transformProps.ts
ktmud: `T` is the separator between date and time. I don't think I've ever seen timestamps like `2020-07-07Z18:20:02.495`.

plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -24,14 +24,52 @@ import {
...
 const { PERCENT_3_POINT } = NumberFormats;
 const TIME_COLUMN = '__timestamp';
-const toString = (x: DataRecordValue) => String(x);
+const REGEXP_TIMESTAMP_NO_TIMEZONE = /T(\d{2}:){2}\d{2}$/;
+
+/**
+ * Extended Date object with custom formatter.
+ * By default `String(date)` returns the raw input.
+ */
+export class DateWithFormatter extends Date {
```
kristw, line 42: could this be separate to its own file?

ktmud:
> I am not sure about extending `Date` but seems like ok for the requirement.

Yeah, I wanted to use `Date.getTime()` value for sorting, but the formatted string for searching. This is the cleanest solution I can think of.

PR #674, opened 2020-07-08
chore: deprecate legacy table plugin

ktmud:
🏠 Internal

With the release of #556 , `legacy-plugin-chart-table` becomes obsolete. 

Move it to potentially deprecated.


rusackas:
Fantastic! Great work on this project, THANK YOU for tackling it. Must feel _great_ to get rid of the old one :)

giphy

PR #675, opened 2020-07-08
chore: Cleaning up control overrides

rusackas:
💔 Breaking Changes

🏆 Enhancements

Just cleaning up a couple of controlOverrides that looked out of place. 

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/plugins/legacy-plugin-chart-heatmap/src/controlPanel.js
+++ b/plugins/legacy-plugin-chart-heatmap/src/controlPanel.js
@@ -223,7 +223,9 @@ export default {
               type: 'CheckboxControl',
               label: t('Normalized'),
               renderTrigger: true,
-              description: t('Whether to normalize the histogram'),
+              description: t(
+                'Whether to apply a normal distribution based on rank on the color scale',
+              ),
               default: false,
             },
           },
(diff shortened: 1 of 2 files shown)


plugins/legacy-plugin-chart-heatmap/src/controlPanel.js
```diff
@@ -254,7 +256,6 @@ export default {
     },
   ],
   controlOverrides: {
-    normalized: t('Whether to apply a normal distribution based on rank on the color scale'),
```
rusackas, line 257: This one looked erroneous, so I moved it to where it looks to be intended

plugins/legacy-preset-chart-nvd3/src/LineMulti/controlPanel.ts
```diff
@@ -149,10 +149,6 @@ export default {
     y_axis_format: {
       label: t('Left Axis Format'),
     },
-    x_axis_format: {
-      choices: D3_TIME_FORMAT_OPTIONS,
-      default: 'smart_date',
-    },
```
rusackas, line 155: This one appears to have the same options as the default for that control, so it's unneeded.

PR #688, opened 2020-07-17
fix: implement extra filter logic

villebro:
🐛 Bug Fix

This implements the `merge_extra_filters` function + applicable features other legacy Python code in `utils/core.py` to fix these TODOs:

```typescript
/** Logic formerly in viz.py's process_query_filters */
export default function processFilters(formData: QueryFormData) {
  // TODO: Implement
  // utils.convert_legacy_filters_into_adhoc(self.form_data)

  // TODO: Implement
  // merge_extra_filters(self.form_data)
```

This removes all deprecated `QueryObject` fields (for instance, replacing `QueryObject.granularity_sqla` with `granularity` and `QueryObject.having` to `QueryObject.extras.having` etc) and makes extra filters work properly.


packages/superset-ui-query/src/extractExtras.ts
```diff
@@ -0,0 +1,47 @@
...
+    } else {
+      // @ts-ignore
+      partialQueryObject.filters.push(filter);
+    }
+  });
+
+  // map to undeprecated names and remove deprecated fields
+  if (isDruidFormData(formData) && !partialQueryObject.druid_time_origin) {
+    partialQueryObject.extras = {
+      druid_time_origin: formData.druid_time_origin,
```
rusackas, line 32: `druid_time_origin` is optional on this data type. Should a default be specified here, e.g. `druid_time_origin: formData.druid_time_origin || someDefault,`?

villebro: The backend expects most of these to be undefined if they haven't been given an explicit value (`time_grain`, `druid_time_origin` etc). So these should be safe to leave as is.

PR #703, opened 2020-07-25
fix: charts rerender with zero-height during tab switch

ktmud:
🐛 Bug Fix

Some charts rely on the dimension of mounted chart containers (`container.clientWidth` and `container.clientHeight`), but both values are zero when Bootstrap tabs are still animating. The unfortunate timing leads to some charts unable to re-render when switching tabs after applying filters (see apache/incubator-superset#10349 ).

Ideally we should fix the [Tabs](https://github.com/apache/incubator-superset/blob/221f38099b1a1a93cd0a3fd6b717f9fdce0b4e18/superset-frontend/src/dashboard/components/gridComponents/Tabs.jsx#L239) component and only re-render charts when animation finishes but I was not able to find a clean way to do that, so fixing the known broken charts instead.

## Test Plan

Manual testing

### DIFF
--- a/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
+++ b/plugins/legacy-plugin-chart-world-map/src/WorldMap.js
@@ -41,13 +41,10 @@ const propTypes = {
 const formatter = getNumberFormatter();
 
 function WorldMap(element, props) {
-  const { data, height, maxBubbleSize, showBubbles } = props;
+  const { data, width, height, maxBubbleSize, showBubbles } = props;
 
   const div = d3.select(element);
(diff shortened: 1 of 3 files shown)


villebro:
Nice @ktmud 👍 Do you know if this problem specific to Bootstrap, and would the corresponding AntD Tab component work correctly without a similar fix?

ktmud:
@villebro I haven't tested. It depends on how they implemented the transition animation.

villebro:
@ktmud feel free to merge and deploy at will so we can bump the package in `incubator-superset`

PR #705, opened 2020-07-27
feat: Waterfall chart

simchaNielsen:
feat: Add Waterfall Chart plugin.

Chart Images:
Initial:
thumbnail

Cross filter example:
Screen Shot 2020-07-27 at 14 06 07

Description:

This feature add new plugin of Waterfall Chart that using `Recharts` library to build this chart.

In Control panel it adds to fields:
`xAxisColumn` and `periodColumn` that will be passed as `groupBy` columns to the DB to get data.


ktmud:
Thanks for the great contribution, @amitNielsen !

Since we have plans to [move Superset to ECharts](https://github.com/apache/incubator-superset/issues/10418), do you mind checking how much work would it be to reimplement [the same visualization](https://echarts.apache.org/examples/en/editor.html?c=bar-waterfall2) in ECharts?

amitNielsen:
In one of our community meetup it was concluded that although core plugins pack will be implemented in ECharts there would be other plugin pack and that it totally ok.

 Ours going towards recharts

PR #711, opened 2020-07-28
feat: improve world map colors

mistercrunch:
Screen Shot 2020-07-27 at 10 22 33 PM

### DIFF
--- a/plugins/legacy-plugin-chart-world-map/package.json
+++ b/plugins/legacy-plugin-chart-world-map/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "d3": "^3.5.17",
     "d3-array": "^2.4.0",
+    "d3-color": "^1.4.1",
     "datamaps": "^0.5.8",
     "prop-types": "^15.6.2"
   },
--- a/plugins/legacy-plugin-chart-world-map/package.json
+++ b/plugins/legacy-plugin-chart-world-map/package.json
@@ -28,7 +28,9 @@
     "access": "public"
   },
   "dependencies": {
+    "@superset-ui/color": "^0.14.9",
     "d3": "^3.5.17",
+    "d3-array": "^2.4.0",
     "datamaps": "^0.5.8",
     "prop-types": "^15.6.2"
   },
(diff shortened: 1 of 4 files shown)


plugins/legacy-plugin-chart-world-map/package.json
```diff
@@ -28,7 +28,9 @@
     "access": "public"
   },
   "dependencies": {
+    "@superset-ui/color": "^0.14.9",
```
ktmud, line 31: `@superset-ui/*` packages are normally `peerDependencies`

mistercrunch: ooops

ktmud:
Close and reopen to trigger CI

mistercrunch:
@ktmud unclear why merging is blocked despite Will's approval. Looking to merge this.

Side-thought: wondering if we want to proxy some sort of "utils" to make sure core visualizations are aligned on the version of `d3` and `lodash` that they bring in, or using peerdependecies for this as well?

I'm just afraid that many visualization would depend on misaligned pinned versions of commonly used libs and bloat the bundles sizes.

ktmud:
@mistercrunch Agreed we should probably have a proxy for the common libraries. `d3` could be included, not sure about `lodash` though (since many lodash functions have easy alternatives in `esnext`).

mistercrunch:
I hate to introduce yet another package `superset-ui-utils` or the like, I wonder if there's a way to generate a common `package-lock.json` from many `package.json`. 

Probably premature optimization... I'd be curious to see how many versions of `d3-*` we end up with...

ktmud:
> generate a common `package-lock.json` from many `package.json`.

That's what monorepo is used for. We do have only one `package-lock.json` (or rather `yarn.lock`) in `superset-ui`, but it's still possible to have many version of `d3` packages if not all plugins are updated at the same time. Currently it's relatively easy to update them all since all core plugins live in the same repo---you just search `"d3.*":` in `package.json`, sync the versions, and fix any potential backward compatibility issues.

In the async plugin world, we won't be able to control which libraries users bundle into their plugin, but we can provide this proxy to common libraries as part of the plugin API so plugins can reuse the libraries shipped with Superset (ECharts, D3, AntD, etc) and load faster.

mistercrunch:
Nice! I didn't know the monorepo had all this magic happening.

Issue #712, opened 2020-07-28
Dist bar bar value will render multiple times


Mikedu1988:
**Describe the bug**
We have a bar chart, and if we click the breakdowns quickly, the value of the dist bar will render multiple times

**To Reproduce**
1. quickly click breakdowns for many times

**Expected behavior**
Only render the last result

**Screenshots**
![image](https://user-images.githubusercontent.com/24424690/88648007-24bcab80-d0f9-11ea-99ae-cb9b20bb2e34.png)

**Additional context**
I think the problem is here in legacy-preset-chart-nvd3/lib/utils.js where bar-chart-label-group removed and added, but if user click more than once during ANIMATION_TIME, there will be more values rendered.
![image](https://user-images.githubusercontent.com/24424690/88648146-47e75b00-d0f9-11ea-8685-f220bd24d709.png)

Mikedu1988:
issue fixed and merged, close the ticket

PR #713, opened 2020-07-28
fix: dist bar will render value multi times

Mikedu1988:
🐛 Bug Fix
related issue: https://github.com/apache-superset/superset-ui/issues/712

Now irrelevantt records will be removed, only the last applied breakpoints will be rendered.


Mikedu1988:
@ktmud

package.json
```diff
@@ -78,7 +78,8 @@
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-test-renderer": "^16.13.1",
-    "rimraf": "^3.0.2"
+    "rimraf": "^3.0.2",
+    "uuid": "^8.3.0"
```
ktmud, line 82: why is this needed?

Mikedu1988: I think this is added during debug, it's not needed. I'll push one commit again.

Mikedu1988:
juse one quick question, we forked and built our own superset docker image, how can we pick up this bug fix? should we just try a rebuild? or we need to bump the version for the plugin here and then we build our image based on the newer version. @ktmud

villebro:
@Mikedu1988 this is now released as version `0.14.21` of `@superset-ui/legacy-preset-chart-nvd3` (will be bumped in main Superset repo shortly)

PR #714, opened 2020-07-29
feat(legacy-plugin-chart-sunburst): add linear color scheme

villebro:
🏆 Enhancements

Currently sunburst chart uses a linear color scheme when using a secondary metric that can't be customized. This adds a linear color control and uses that when the secondary metric is defined.

### Screenshot
captured (18)

### DIFF
--- a/plugins/legacy-plugin-chart-sunburst/src/controlPanel.ts
+++ b/plugins/legacy-plugin-chart-sunburst/src/controlPanel.ts
@@ -34,7 +34,7 @@ export default {
     {
       label: t('Chart Options'),
       expanded: true,
-      controlSetRows: [['color_scheme', 'label_colors']],
+      controlSetRows: [['color_scheme', 'linear_color_scheme']],
     },
   ],
   controlOverrides: {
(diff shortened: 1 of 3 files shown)


plugins/legacy-plugin-chart-sunburst/src/controlPanel.ts
```diff
@@ -34,7 +34,7 @@ export default {
     {
       label: t('Chart Options'),
       expanded: true,
-      controlSetRows: [['color_scheme', 'label_colors']],
+      controlSetRows: [['color_scheme', 'linear_color_scheme']],
```
villebro, line 37: the `label_colors` control isn't used anywhere in this chart, and actually doesn't work (shows up blank). Interestingly enough it's used in several charts. Deferring looking into this to a later PR.

plugins/legacy-plugin-chart-sunburst/src/controlPanel.ts
```diff
@@ -51,6 +51,12 @@ export default {
           'When omitted, the color is categorical and based on labels',
       ),
     },
+    color_scheme: {
+      description: t('When only a primary metric is provided, a categorical color scale is used.'),
+    },
+    linear_color_scheme: {
+      description: t('When a secondary metric is provided, a linear color scale is used.'),
+    },
```
rusackas, line 59: We _could_ swap this based on the control state when the secondary metric is provided, but maybe that'd be too jarring to the user to just have the color scheme automatically change.

PR #717, opened 2020-07-30
feat(plugin-chart-table): add translate to chart elements

stuarthu:
Enable translate for some elements on table chart.
Will also update superset translate package.

🏆 Enhancements

# BEFORE
截屏2020-07-30 17 28 18

# AFTER
截屏2020-07-30 17 28 08

### DIFF
--- a/plugins/plugin-chart-table/src/DataTable/components/SelectPageSize.tsx
+++ b/plugins/plugin-chart-table/src/DataTable/components/SelectPageSize.tsx
@@ -17,6 +17,7 @@
  * under the License.
  */
 import React from 'react';
+import { t } from '@superset-ui/translation';
 
 export type SizeOption = number | [number, string];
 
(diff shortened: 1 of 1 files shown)


stuarthu:
@ktmud @mistercrunch

stuarthu:
@villebro

ktmud:
This is does not work when 'Show' or 'entries' are used in other context when they may have different meanings. You also shouldn't import `superset-ui` in `DataTable` as there are plans to make it an independent package.

I'll make another PR to address this.

stuarthu:
> This is does not work when 'Show' or 'entries' are used in other context when they may have different meanings. You also shouldn't import `superset-ui` in `DataTable` as there are plans to make it an independent package.
> 
> I'll make another PR to address this.

Hi @ktmud , if it needs more than 1 week to address this, can we submit this PR first to solve the translation issue and refactor later?

PR #728, opened 2020-08-09
refactor: more @superset-ui/core to superset-ui

ktmud:
🏠 Internal

Move `@superset-ui/core` to a new package called `superset-ui`. This is the first step of merging all `@superset-ui/` core packages into one package.

Merges into a feature branch `monopackage`.


packages/superset-ui-chart-composition/src/ChartFrame.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { PureComponent } from 'react';
-import { isDefined } from '@superset-ui/core';
+import { isDefined } from 'superset-ui/lib/core';
```
ktmud, line 2: Because of restrictions from `nimbus`, currently we still have to import from the `superset-ui/lib` folder. I really want to get rid of the "/lib" folder for published packages, but it would require more advanced customization with the build process that is not easy to do with `nimbus`.

Will try to revisit this after all packages are merged.

suddjian: I don't really understand enough about nimbus to grok these restrictions. Does this also mean we'll have to import from `/lib` in Superset?

kristw: > Because of restrictions from nimbus

What restriction? I don't remember having that issue before.

kristw:
> Move @superset-ui/core to a new package called superset-ui. 

Better use `@superset-ui/xxx` name to leverage existing npm org for publishing scoped package. Otherwise we will need two permission granting for every new maintainer instead of just giving access to `@superset-ui`
You can also repurpose `@superset-ui/core` to be the monopackage.

packages/superset-ui-chart-composition/src/ChartFrame.tsx
kristw: > I really want to get rid of the /lib folder for published packages

Could you elaborate more on why the removal of `lib` is necessary?

ktmud: > What restriction? I don't remember having that issue before.

nimbus expects the src and outputs to be certain layout and it's difficult to override the defaults.

ktmud: > I don't really understand enough about nimbus to grok these restrictions. Does this also mean we'll have to import from `/lib` in Superset?

For the time-being or before the packages are all merged, yes. We can always merge the feature branch only when we find a solution to get rid of `/lib`.

It's also possible to import all of them from the root index file `import { SupersetChart } from "superset-ui"` though.

ktmud:
> > Move @superset-ui/core to a new package called superset-ui.
> 
> Better use `@superset-ui/xxx` name to leverage existing `npm` org for publishing scoped package. Otherwise we will need two permission granting for every new maintainer instead of just giving access to `@superset-ui`
> You can also repurpose `@superset-ui/core` to be the monopackage.

Sure. I'll use `@superset-ui/core`.

packages/superset-ui-chart-composition/src/ChartFrame.tsx
ktmud: > > I really want to get rid of the /lib folder for published packages
> 
> Could you elaborate more on the motivation to remove `lib`?

It just looks better. 😄   Plus it'd also be easier (and less confusing) when we redirect it to the `/src` folder via `resolve` aliases, for IDE intellisense or `npm link`.

But I guess we can live with it if it's too much work or adds too much complexity to the build process.

ktmud:
We'll just repurpose `@superset-ui/core` to contain all core packages. Since there is no point to move "@superset-ui/core" to "@superset-ui/core/core", closing this PR and starting to move other packages to `core`.

packages/superset-ui-chart-composition/src/ChartFrame.tsx
suddjian: Makes sense. This looks pretty good to me:

> ```import { SupersetChart } from "superset-ui"```

Though it may be too many things exported under one namespace.

kristw: > Makes sense. This looks pretty good to me:
> 
> > `import { SupersetChart } from "superset-ui"`
> 
> Though it may be too many things exported under one namespace.

I think we should export all under one. Seems a lot simpler. 
There is also no naming collision at the moment. 
(I previously did re-export all subpackages in `@superset-ui/superset-ui` that re-exports all from the micro packages and there was no name conflicts)

PR #729, opened 2020-08-10
refactor: move translation, connection, query to core

ktmud:
🏠 Internal

Follow up of #728 .

Started with merging `connection` and `query` into core, because they are closely related. Ended up with moving `translation` as well because it was used in `test/setup.js`.

PR #730, opened 2020-08-11
build: move number-format and time-format to core

ktmud:
🏠 Internal

Follow up for #729 , merging more packages to `@superset-ui/core`. This time the d3 formatters.


ktmud:
@kristw Tests are failing because `encodable` depends on an older version of `@superset-ui/number-format`. What do you think is a good way to reduce such inter-dependencies?

kristw:
> @kristw Tests are failing because `encodable` depends on an older version of `@superset-ui/number-format`. What do you think is a good way to reduce such inter-dependencies?

I am working to decouple `encodable` from `superset-ui`. Should take a few more days.

kristw:
Wonder if there is an easy way to disable certain plugin building / testing temporarily to unblock you.

ktmud:
> Wonder if there is an easy way to disable certain plugin building / testing / publishing temporarily to unblock you.

We can move the packge to `temporary-plugins/`? The only block package is `presets-chart-xy`, which we only use for Boxplot in Superset.

kristw:
> > Wonder if there is an easy way to disable certain plugin building / testing / publishing temporarily to unblock you.
> 
> We can move the packge to `temporary-plugins/`? The only block package is `presets-chart-xy`, which we only use for Boxplot in Superset.

sgtm

ktmud:
There are a couple of other plugins also use `encodable`. It's too much work to move them (and their storybook examples) back and forth. I'll just leave this PR here and work on merging other packages instead.

PR #734, opened 2020-08-12
feat(legacy-plugin-chart-pivot-table): add support for timestamp format

villebro:
🏆 Enhancements

This adds support for formatted dates on `pivot-table`. This is done by prefixing date values with `__timestamp:`, after which they can be identified and parsed into formattable date objects. In addition sorting is done based on the epoch value, meaning that the format won't affect the sorting order. Other changes:
- Add date format
- Move visual options to own tab (currently on `Query` pane)
- use time grain for formatting if smart formatting is selected
- make sorting of numeric values based on the parsed float value, not the text value

### AFTER:
image

### BEFORE:
image

### DIFF
--- a/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
+++ b/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
@@ -106,16 +106,16 @@ function PivotTable(element, props) {
         const format = columnFormats[metric] || numberFormat || '.3s';
         const tdText = $(this)[0].textContent;
         const parsedValue = parseFloat(tdText);
-        if (!Number.isNaN(parsedValue)) {
-          $(this)[0].textContent = formatNumber(format, parsedValue);
-          $(this).attr('data-sort', parsedValue);
-        } else {
+        if (Number.isNaN(parsedValue)) {
           const regexMatch = dateRegex.exec(tdText);
           if (regexMatch) {
             const date = new Date(parseFloat(regexMatch[1]));
             $(this)[0].textContent = dateFormatter(date);
             $(this).attr('data-sort', date);
           }
+        } else {
+          $(this)[0].textContent = formatNumber(format, parsedValue);
+          $(this).attr('data-sort', parsedValue);
         }
       });
   });
--- a/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
+++ b/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
@@ -20,7 +20,11 @@
 import dt from 'datatables.net-bs';
 import PropTypes from 'prop-types';
 import { formatNumber } from '@superset-ui/number-format';
-import { smartDateFormatter, getTimeFormatter } from '@superset-ui/time-format';
+import {
+  getTimeFormatter,
+  getTimeFormatterForGranularity,
+  smartDateFormatter,
+} from '@superset-ui/time-format';
 import fixTableHeight from './utils/fixTableHeight';
 import 'datatables.net-bs/css/dataTables.bootstrap.css';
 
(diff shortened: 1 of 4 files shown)


plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
```diff
@@ -44,25 +45,37 @@ const propTypes = {
 };
 
 function PivotTable(element, props) {
-  const { data, height, columnFormats, numberFormat, numGroups, verboseMap } = props;
+  const { data, height, columnFormats, dateFormat, numberFormat, numGroups, verboseMap } = props;
 
   const { html, columns } = data;
   const container = element;
   const $container = $(element);
```
rusackas, line 52: jQuery!? 😭

ktmud: Unfortunately pivot table still uses `jQuery.dataTables`.

plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
```diff
@@ -44,25 +45,37 @@ const propTypes = {
...
+  const dateFormatter =
+    dateFormat === 'smart_date' ? smartDateFormatter : getTimeFormatter(dateFormat);
```
ktmud, line 54: `getTimeFormatter('smart_date')` should work just fine. But since we are here, can you update it to follow the same logic as the table chart? 

https://github.com/apache-superset/superset-ui/blob/d323ce86dd7000e9b65a2a7d6b861d713616701b/plugins/plugin-chart-table/src/transformProps.ts#L138-L151

It has fallbacks to d3-format in pre-defined columns and uses Time Grain for `smart_date`.

ktmud: If the code was hard to follow for you, I can also update this later and refactor these logics into some kind of common utilities.

plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
villebro: Hey, if it works it works! 😄

plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
villebro: Thanks for this reference! I didn't have strong opinions here aka. I was just out to make this work. I'll digest the table chart logic and see if it makes sense for me to streamline here 👍

villebro:
Thanks for your reviews @ktmud and @rusackas ! <3

villebro:
Oh, one more thing @ktmud and @rusackas . Currently the sorting carets are broken (displays a black box, see the screenshots). I wasn't sure how to fix this, do you guys know how to make the icon work properly?

villebro:
Thanks @kristw , I'll update accordingly 👍

ktmud:
> Oh, one more thing @ktmud and @rusackas . Currently the sorting carets are broken (displays a black box, see the screenshots). I wasn't sure how to fix this, do you guys know how to make the icon work properly?

I used `react-icons` in the table chart. https://github.com/apache-superset/superset-ui/blob/d323ce86dd7000e9b65a2a7d6b861d713616701b/plugins/plugin-chart-table/src/TableChart.tsx#L91-L98

In the future we should probably move the SVG icons from `superset-frontend` to `superset-ui` so they can be reused by visualizations.

plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
villebro: @ktmud I implemented a slightly modified version of your formatter that felt most suitable for this slightly simpler case. Does it make sense to you?

villebro:
> I used `react-icons` in the table chart.
> 
> https://github.com/apache-superset/superset-ui/blob/d323ce86dd7000e9b65a2a7d6b861d713616701b/plugins/plugin-chart-table/src/TableChart.tsx#L91-L98
> 
> In the future we should probably move the SVG icons from `superset-frontend` to `superset-ui` so they can be reused by visualizations.

@ktmud I was unsure how to apply this snippet to this code, as it doesn't seem to be natively React. Perhaps it's vest if we merge this as-is and leave this fix for later?

plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
ktmud: Looks good! Thanks for the update. Do you mind adding a datetime column to [Storybook](https://github.com/apache-superset/superset-ui/blob/d631b17597483f981f5de67d2c482e8fae94dc55/packages/superset-ui-demo/storybook/stories/plugins/legacy-plugin-chart-pivot-table/Stories.tsx#L27) as well?

ktmud:
> @ktmud I was unsure how to apply this snippet to this code, as it doesn't seem to be natively React. Perhaps it's best if we merge this as-is and leave this fix for later?

Sounds good

PR #738, opened 2020-08-15
feat: logger supports console.table, with console.log fallback

rusackas:
💔 Breaking Changes

🏆 Enhancements
Supports console.table safely :)

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-core/src/utils/logging.ts
+++ b/packages/superset-ui-core/src/utils/logging.ts
@@ -16,6 +16,7 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+
 const logger = window.console || {
   debug() {},
   log() {},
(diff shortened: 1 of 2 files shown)


rusackas:
Not quite sure how to unit test this, since it depends on the browser.

ktmud:
> Not quite sure how to unit test this, since it depends on the browser.

Here's how I tested it: https://github.com/apache-superset/superset-ui/blob/9a011e47c3edb34d0aafed4856414b571b49645f/packages/superset-ui-core/test/utils/logging.test.ts#L42

If you have time, might as well do the fallbacks for all other console methods as well.

```ts
const console = window.console || {};
const log = console.log || () => {};
const logger = {
  log,
  warn: console.warn || log,
  debug: console.debug || log,
  ...
}
```

PR #744, opened 2020-08-19
build: update encodable dependency

kristw:
🏠 Internal

* Bump `encodable` to latest version which no longer has nested dependency to `@superset-ui/color`, `@superset-ui/number-format` and `@superset-ui/time-format`.
* By default `encodable` ships with its own color and formatter registries. This PR include shims to point to the same registries that `@superset-ui` is using, so the specified formats and colorScheme from control panel can be correctly used. Have to call this shim function in the setup.
  * There are multiple copies of the same shim in each package at the moment. It could be extracted to a new package, but seems like an overkill for this first pass.
  * I may also refactor `@superset-ui/colors,number-format,time-format` later to use the same underlying registries so the shim won't be necessary.

scatter plot working
image

word cloud working
image

map working
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "esprint": "^0.7.0",
     "fast-glob": "^3.2.4",
     "fs-extra": "^9.0.0",
+    "global-box": "^1.2.0",
     "husky": "^4.2.5",
     "identity-obj-proxy": "^3.0.0",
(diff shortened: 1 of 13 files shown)


kristw:
There seems to be some issue with box plot. Investigating....

kristw:
The issue is fixed and good to go.

PR #748, opened 2020-08-20
feat: add certification icon to metrics

etr2460:
🏆 Enhancements
Builds from https://github.com/apache/incubator-superset/pull/10630 to add the certification icon into the MetricOption component. Unfortunately there's some code duplication here as the main Icon component is in incubator-superset. In the future we'll want to move icons into superset-ui I think.

Test Plan:
`npm link` to incubator-superset, test with a certified metric and an uncertified metric

Screen Shot 2020-08-20 at 1 42 10 PM
Screen Shot 2020-08-20 at 1 42 15 PM

to: @ktmud @graceguo-supercat

### DIFF
--- a/packages/superset-ui-chart-controls/src/components/MetricOption.tsx
+++ b/packages/superset-ui-chart-controls/src/components/MetricOption.tsx
@@ -16,21 +16,33 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+/* eslint-disable camelcase */
 import React from 'react';
+import styled from '@superset-ui/style';
 import InfoTooltipWithTrigger from './InfoTooltipWithTrigger';
 import { ColumnTypeLabel } from './ColumnTypeLabel';
+import CertifiedIconWithTooltip from './CertifiedIconWithTooltip';
+
+const FlexRowContainer = styled.div`
+  align-items: center;
+  display: flex;
+
+  > svg {
+    margin-right: ${({ theme }) => theme.gridUnit}px;
+  }
+`;
 
 export interface MetricOptionProps {
   metric: {
-    // eslint-disable-next-line camelcase
     verbose_name: string;
-    // eslint-disable-next-line camelcase
     metric_name: string;
     label: string;
     description: string;
-    // eslint-disable-next-line camelcase
     warning_text: string;
     expression: string;
+    is_certified?: boolean;
+    certified_by?: string | null;
+    certification_details?: string | null;
   };
   openInNewWindow: boolean;
   showFormula: boolean;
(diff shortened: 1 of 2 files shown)


ktmud:
Does it bother you that the selected pills has different heights?

etr2460:
> Does it bother you that the selected pills have different heights?

I don't know... I noticed that too, but wasn't sure how best to fix it. I don't want to make the icon any smaller as that makes it less noticeable. Perhaps increasing the size of the pill for everything is better? But then it'd be inconsistent with all the other pills in the sidepane controls. In short, I wasn't sure what to do here :P

packages/superset-ui-chart-controls/src/components/MetricOption.tsx
```diff
@@ -16,21 +16,33 @@
...
     verbose_name: string;
-    // eslint-disable-next-line camelcase
     metric_name: string;
     label: string;
     description: string;
-    // eslint-disable-next-line camelcase
     warning_text: string;
     expression: string;
+    is_certified?: boolean;
+    certified_by?: string | null;
```
john-bodley, line 44: @etr2460 sorry for not picking up on this earlier, but shouldn't `certified_by` be an `integer` and refer to the user ID?

etr2460: So regardless of the backend structure here, we'd want to pass down all the text necessary to render it in the frontend without an additional api call.

However, we made the choice of a free text `certified_by` field on the backend for a few reasons:
- It makes iteration easier at this early phase
- It allows entities other than users (like groups of users, team names, certification protocols) to certify metrics
- If a user is deleted (perhaps because of leaving a team) we still keep the record of who certified the metric (although this could be fixed in the future with paranoid deletes)

john-bodley: @etr2460 I have some concerns with denormalized data: 

- There's no consistency/accuracy with free-form entries as opposed to a typeahead, i.e., the user could be named "John Doe" however it could be incorrectly labelled as certified by "Jon Doe".
- Superset (for right or wrong) only knows about users. It seems somewhat strange to add flexibility for certification (granted this is embedded in a JSON blob and thus there's no type coupling from a foreign key perspective).
- Paranoid deletes should never delete users. The `ab_user` table has an `active` column which is how we should make users inactive as opposed to deleting records.

john-bodley: @etr2460 as long as we're committed to the fact that there may be additional work required for said feature.

etr2460: Don't worry, I'm committed to a decent amount more work here. Definitely don't want to get this to "working" and then ignore the tech debt left behind

etr2460:
@ktmud the pills are now the same height!

williaster:
thanks for making them the same height! ✨

PR #751, opened 2020-08-25
feat(color): add official ECharts color schemes

villebro:
🏆 Enhancements

Add ECharts 4.x and 5.x official color schemes.
4.x: https://github.com/apache/incubator-echarts/blob/eb396f27137af1d74b42b1b28752f204a34dbecf/src/model/globalDefault.js#L36-L41
5.x: https://github.com/apache/incubator-echarts/blob/4c916f527f211a34c8dd5555c23870db25a78b05/src/model/globalDefault.ts#L42-L55

Categorical:
image
Sequential:
image

Ping @Ovilia


kristw:
I am confused. Why are they called "ECharts 4 colors" and "ECharts 5 colors" while there are more than 4 and 5?

villebro:
@kristw it's for version 4.x vs version 5.x. I'll update to make it more clear.

PR #752, opened 2020-08-25
chore: silencing console errors by making MetricOption props optional

rusackas:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
Explore had a bunch of console warnings going on since various implementations of <MetricOption> did not provide all the props that the TS definition had listed as mandatory. This loosens up the typing so the console is less painful to look at :)

🏠 Internal


mistercrunch:
cleanup

PR #768, opened 2020-09-03
refactor: merge core superset-ui packages

ktmud:
A colossal of moving files around to merge all core superset-ui packages into one monopackage `@superset-ui/core`. This move removes the needs to maintain correct inter-package dependencies and will enable us to iterate on Superset's charting foundation faster.

Everything from the previous packages are exported at root. For most cases, the implication for end users is to simply replace their `import .. from '@superset-ui/xxx'` to  `import .. from '@superset-ui/core'`. More refactors may be planned to make some of the exports less ambiguous. 

The work was done in a feature branch. This PR merges the feature branch to `master`. Here is a list of all related PRs:

1. move translation, connection, query to core (#729)
2. update encodable to remove formatter inter-dependency (#744)
3. move number-format and time-format to core (#730)
4. move superset-ui/dimension to core (#732)
5. move superset-ui/color to core (#755)
6. move superset-ui/style to core (#756)
7. move superset-ui/validator to core (#757)
8. move superset-ui/chart-composition to core (#759)
9. move superset-ui/chart to core (#760)

All packages are merged except `@superset-ui/chart-controls` because if we move it to core and export it from there, it would imports the i18n function `t(...)` before translations are initialized (currently users have to `import { configure } from '@superset-ui/core'` to setup the language pack. 

Migration plan:

- [ ] Deprecate old packages in NPM
- [ ] Publish a major version for `@superset-ui/core` and `@superset-ui/chart-controls`.
- [ ] Update deck.gl and other other plugins not in this monorepo.
- [ ] Replace all packages in `incubator-superset`

💔 Breaking Changes

🏆 Enhancements

🏠 Internal


kristw (approved):
so long!

PR #769, opened 2020-09-03
build: enable skipLibCheck for tsc

ktmud:
🏆 Enhancements
🏠 Internal

From TypeScript [handbook](https://www.typescriptlang.org/docs/handbook/compiler-options.html):

```
--skipLibCheck       Skip type checking of all declaration files (*.d.ts).
```

This cuts build-time for `tsc` in half. Most TS errors can still be captured by ESLint.

Also limit `typeRoots` to `./node_modules` in the root folder so to reduce chances of type conflicts from `packages/*` and `plugins/*`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,7 +174,10 @@
         "declaration": true,
         "emitDeclarationOnly": true,
         "composite": true,
-        "resolveJsonModule": true
+        "resolveJsonModule": true,
+        "typeRoots": [
+          "./node_modules/@types"
+        ]
       }
     }
   },
--- a/package.json
+++ b/package.json
@@ -170,6 +170,7 @@
     "typescript": {
       "compilerOptions": {
         "rootDir": "./src",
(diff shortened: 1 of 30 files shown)


rusackas (approved):
LGTM, but wanted to ask:
Swapping the boolean for the cache cleaning sounds like a good way to speed things up, but are you sure swapping that default doesn't have any potentially negative impacts?

Also:
> This cuts build-time for tsc in half.

OMG! 🎉 ❤️

kristw (approved):

ktmud:
> Swapping the boolean for the cache cleaning sounds like a good way to speed things up, but are you sure swapping that default doesn't have any potentially negative impacts?

It won't break CI and published packages at least. For local developers, things may break if they cleaned up `d.ts` files without removing `.tsbuildinfo`. But they can also run `yarn clean` manually and start fresh again.

PR #770, opened 2020-09-03
chore: compress thumbnail images with pngquant

ktmud:
Optimized the thumbnail images with pngquant to reduce bundlesize:

```
pngquant --quality=65-85 ./{packages,plugins}/**/*.png --ext .png --force   
```


kristw (approved):

rusackas (approved):
LGTM... seems like it would be cool to have this be part of an asset pipeline step in the build process, but that's another story entirely!

PR #771, opened 2020-09-04
refactor: convert legacy-plugin-chart-event-flow to typescript

ktmud:
Found some bad imports that were not captured by CI, neither ESLint nor tsc, decided to convert the whole package to TypeScript so we will have less such errors in the future.

### DIFF
--- a/plugins/legacy-plugin-chart-event-flow/src/controlPanel.tsx
+++ b/plugins/legacy-plugin-chart-event-flow/src/controlPanel.tsx
@@ -17,10 +17,17 @@
  * under the License.
  */
 import React from 'react';
-import { t, validateNonEmpty, ColumnOption, columnChoices } from '@superset-ui/core';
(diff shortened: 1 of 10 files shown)


plugins/legacy-plugin-chart-event-flow/src/controlPanel.tsx
```diff
@@ -17,10 +17,17 @@
  * under the License.
  */
 import React from 'react';
-import { t, validateNonEmpty, ColumnOption, columnChoices } from '@superset-ui/core';
```
ktmud, line 20: Here is the bad import: `ColumnOption` and `columnChoices` should have been imported from `@superset-ui/chart-controls` instead of `@superset-ui/core`.

kristw (approved):

rusackas (approved):

PR #783, opened 2020-09-11
feat(plugin-chart-echarts): add more robust formatting for series

villebro:
🏆 Enhancements

Adds proper support for non-string labels to ECharts Pie chart, especially for numeric, temporal and `NULL` types.

### DIFF
--- a/plugins/plugin-chart-echarts/src/utils/series.ts
+++ b/plugins/plugin-chart-echarts/src/utils/series.ts
@@ -19,6 +19,7 @@
  */
 import {
   DataRecord,
+  DataRecordValue,
   NumberFormatter,
   TimeFormatter,
   TimeseriesDataRecord,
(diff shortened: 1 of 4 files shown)


lilykuang (approved):

riahk (approved):

plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -39,7 +44,43 @@ export function extractTimeseriesSeries(
     }));
 }
 
-export function extractGroupbyLabel(datum: DataRecord, groupby: string[]): string {
+export function formatSeriesName({
```
ktmud, line 47: Can we change the signature to `formatSeriesName(name, { numberFormatter, timeFormatter })` and make `name` an required argument?

villebro: Thanks, I wasn't aware of this being a common pattern, I'll update accordingly 👍

ktmud (approved):

PR #788, opened 2020-09-17
fix(plugin-chart-word-cloud): make colors schemes work

villebro:
🐛 Bug Fix
Currently the selected color scheme isn't being respected. This is probably a regression from #768. 

### BEFORE
image

### AFTER
image

### DIFF
--- a/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
+++ b/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
@@ -2,9 +2,6 @@ import React from 'react';
 import cloudLayout, { Word } from 'd3-cloud';
 import { PlainObject, createEncoderFactory, DeriveEncoding } from 'encodable';
 import { SupersetThemeProps, withTheme } from '@superset-ui/core';
-import configureEncodable from '../configureEncodable';
-
-configureEncodable();
 
 export const ROTATION = {
   flat: () => 0,
--- a/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
+++ b/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
@@ -2,6 +2,9 @@ import React from 'react';
 import cloudLayout, { Word } from 'd3-cloud';
 import { PlainObject, createEncoderFactory, DeriveEncoding } from 'encodable';
 import { SupersetThemeProps, withTheme } from '@superset-ui/core';
+import configureEncodable from '../configureEncodable';
+
+configureEncodable();
 
 export const ROTATION = {
   flat: () => 0,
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
```diff
@@ -2,6 +2,9 @@ import React from 'react';
 import cloudLayout, { Word } from 'd3-cloud';
 import { PlainObject, createEncoderFactory, DeriveEncoding } from 'encodable';
 import { SupersetThemeProps, withTheme } from '@superset-ui/core';
+import configureEncodable from '../configureEncodable';
+
+configureEncodable();
```
villebro, line 7: @ktmud I wasn't sure where/how to call this "the right way", any pointers would be appreciated!

villebro:
Side note: the non-legacy plugin is using `transformProps.ts` from the legacy plugin. We should probably remove `transformProps.ts` from the new word cloud plugin if it isn't used or going to be worked on any time soon. Or better yet, just remove the legacy plugin and move any code that the new plugin is using to the appropriate place.

plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
ktmud: Thanks for catching and fixing this!

I feel it's probably better to add these in the plugin definition file (`src/plugin/index.ts`).  @kristw what do you think?

ktmud:
Boxplot seems to have the same problem.

plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
kristw: umm, the regression is in `incubator-superset`, right?
Better call this in the setup stage of `incubator-superset`, not in the plugin. 

I configured it here for the storybook and haven't done that to the `incubator-superset` because we were still consolidating core packages at that time.
https://github.com/apache-superset/superset-ui/blob/0f11940d8201feea3c5f5adee82aef68a9d201f3/packages/superset-ui-demo/.storybook/preview.js#L70
See preview.
https://superset-ui.now.sh/?path=/story/chart-plugins-plugin-chart-word-cloud--legacy-shim

kristw: or `src/plugin/index.ts` as Jesse suggested, but then have to do that for the other ones that use encodable too.

kristw: This should only be required temporarily so either solution is fine. I am about to refactor the core package to use same underlying resolvers with encodable so this configuration should eventually be removed.

villebro: Makes sense 👍  Updated.

villebro:
@ktmud boxplot also fixed, should be good to go.

ktmud (approved):

PR #797, opened 2020-09-28
fix: copying controls from Sankey to Sankey Loop

rusackas:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix
Some folks pointed out the Sankey Loop plugin wasn't working... welp, it has no controls! This copies the control panel over from the regular Sankey plugin. Calling this a Draft PR until I (or anyone) can confirm this actually works.

🏠 Internal


villebro (approved):
LGTM. I have a follow-up PR which enables this in the backend (required minor adjustments).
image

PR #798, opened 2020-09-28
fix: echarts formdata series -> groupby

rusackas:
🐛 Bug Fix

#772 introduced a regression where the missing property `formData.series` was mapped to `groupby`, instead of `formData.groupby`.


villebro (approved):

villebro:
I'll fix this tomorrow + address unit tests (they seem wrong right now).

PR #800, opened 2020-09-29
fix(plugin-chart-echarts): fix incorrect groupby in buildQuery

villebro:
🐛 Bug Fix
Fix a regression caused by party contradictory naming in controls/buildQuery.


rusackas (approved):
LGTM!

PR #801, opened 2020-10-05
feat(plugin-chart-echarts): add BoxPlot chart

villebro:
🏆 Enhancements
Add BoxPlot to ECharts plugin package. In addition to having full feature parity with `preset-chart-xy/BoxPlot`, this plugin adds the following features:
- adds support for using non-temporal columns as the basis for the boxplot.
- adds timegrain to the control panel.

Explore View:
image

New control panel. Notice that groupby has been renamed `series`, and `Distribute across` has been added for non-temporal marginals:
image
image

# World Bank's Data example dashboard after:
image

# World Bank's Data example dashboard before:
image


packages/superset-ui-chart-controls/src/shared-controls/index.tsx
```diff
@@ -111,7 +111,7 @@ const groupByControl: SharedControlConfig<'SelectControl', ColumnMeta> = {
   clearable: true,
   default: [],
   includeTime: false,
-  description: t('One or many controls to group by'),
+  description: t('One or many columns to group by'),
```
villebro, line 114: Bycatch

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-echarts/Pie/Stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+        showLegend: boolean('Show legend', false),
+        pieLabelType: select(
+          'Pie label type',
+          ['key', 'value', 'percent', 'key_value', 'key_percent', 'key_value_percent'],
+          'key_value_percent',
+        ),
+      }}
+    />
+  );
+};
```
villebro, line 41: The ECharts Pie story was missing, so I added it here.

plugins/plugin-chart-echarts/src/BoxPlot/buildQuery.ts
```diff
@@ -0,0 +1,63 @@
...
+      type = 'percentile';
+      percentiles = [parseInt(percentileMatch[1], 10), parseInt(percentileMatch[2], 10)];
+    } else {
+      throw new Error(`Unsupported whisker type: ${whiskerOptions}`);
+    }
+    console.log(formData);
+    return [
+      {
+        ...baseQueryObject,
+        is_timeseries: distributionColumns.length === 0,
```
villebro, line 47: For backward compatibility, the chart uses the temporal column to construct the marginal distribution for the boxplot chart.

plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
```diff
@@ -0,0 +1,114 @@
...
+    },
+  ],
+  sectionOverrides: {
+    druidTimeSeries: {
+      controlSetRows: [['granularity', 'druid_time_origin'], ['time_range']],
+    },
+    sqlaTimeSeries: {
+      controlSetRows: [['granularity_sqla', 'time_grain_sqla'], ['time_range']],
+    },
+  },
```
villebro, line 98: Timegrain has been added (not sure why it was missing)

plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
```diff
@@ -0,0 +1,114 @@
...
+      clearable: false,
+      description: t('Categories to group by on the x-axis.'),
+    },
+    columns: {
+      label: t('Distribute across'),
+      multi: true,
+      description: t(
+        'Columns to calculate distribution across. Defaults to temporal column if left empty.',
+      ),
+    },
```
villebro, line 112: Additional control to construct marginal distribution based on other than temporal dimension.

plugins/plugin-chart-echarts/src/BoxPlot/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  xTickLayout?: BoxPlotFormXTickLayout;
+};
+
+export type BoxPlotFormDataWhiskerOptions =
+  | 'Tukey'
+  | 'Min/max (no outliers)'
+  | '2/98 percentiles'
+  | '9/91 percentiles';
+
+export type BoxPlotFormXTickLayout = '45°' | '90°' | 'auto' | 'flat' | 'staggered';
```
villebro, line 33: 90° rotation added. Staggered isn't really supported in ECharts BoxPlot (neither was it in the previous plugin), but leaving this here for backward compatiblity + if it gets implemented in ECharts sometime in the future.

plugins/plugin-chart-echarts/src/Pie/controlPanel.ts
```diff
@@ -119,7 +119,7 @@ const config: ControlPanelConfig = {
               min: 10,
               max: 100,
               step: 1,
-              default: 80,
+              default: 50,
```
villebro, line 122: Bycatch: making the default pie slightly smaller, as labels outside will otherwise easily overflow the canvas if a chart is made very small.

plugins/plugin-chart-echarts/src/Timeseries/controlPanel.ts
```diff
@@ -291,7 +291,7 @@ const config: ControlPanelConfig = {
   },
   controlOverrides: {
     row_limit: {
-      default: 100,
+      default: 10000,
```
villebro, line 294: This default value was too small - should default to at least 10000 for timeseries charts.

villebro:
@kristw @ktmud @rusackas this is ready for review, please take a look if you have the time!

eugeniamz:
Can you add the screenshot of the controls? I know that you added functionality, it would be nice to see it here.

villebro:
@eugeniamz I added the control panel screenshots to the description.

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-echarts/BoxPlot/data.ts
```diff
@@ -0,0 +1,122 @@
+export default [
```
rusackas, line 1: Just an aside/thought, but I know the E2E tests for most plugins are in Superset, not in the Plugins themselves. 

I'm hoping we can find a way to make the E2E tests live alongside the plugins themselves. On top of that, this data file makes me wonder if Jest/Enzyme tests that `mount` the viz from this Storybook entry, with all this test data, might be sufficient enough.

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-echarts/Pie/Stories.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { SuperChart, getChartTransformPropsRegistry } from '@superset-ui/core';
+import { boolean, select, withKnobs } from '@storybook/addon-knobs';
```
rusackas, line 3: This is fine, but just thought it might be worth noting that Storybook now supports "controls" so we might want to sweep up knobs in favor of controls some rainy day.

villebro: Right, I forgot about those! I'll leave that for an even rainier day than today. 🌧️

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-echarts/BoxPlot/data.ts
villebro: True, I think whatever Storybook is doing should constitute an equivalent of an e2e test. I can look into how the two could somehow be combined.

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -0,0 +1,178 @@
...
+        const name = metrics.length === 1 ? groupbyLabel : `${groupbyLabel}, ${metric}`;
+        // Outlier data is a nested array of numbers (uncommon, therefore no need to add to DataRecordValue)
+        const outlierDatum = (datum[`${metric}__outliers`] || []) as number[];
+        return {
+          name: 'outlier',
+          type: 'scatter',
+          data: outlierDatum.map(val => [name, val]),
+          tooltip: {
+            formatter: (param: { data: [string, number] }) => {
+              const headline = `<p><strong>${param.data[0]}</strong></p>`;
```
villebro, line 89: I'm not sure what the correct way of doing this is (I tried `<b>xxx</b>` but that went all crazy in Explore view).

plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
```diff
@@ -0,0 +1,114 @@
...
+        [
+          {
+            name: 'whiskerOptions',
+            config: {
+              type: 'SelectControl',
+              freeForm: true,
+              label: t('Whisker/outlier options'),
+              default: 'Tukey',
+              description: t('Determines how whiskers and outliers are calculated.'),
+              choices: formatSelectOptions([
```
rusackas, line 42: `renderTrigger: true`, maybe?

rusackas:
FYI, the thumbnail for BoxPlot is a pie/ring/donut chart

plugins/plugin-chart-echarts/src/BoxPlot/index.ts
```diff
@@ -0,0 +1,49 @@
...
+   * advanced visualizations that require either post processing operations
+   * (pivoting, rolling aggregations, sorting etc) or submitting multiple queries.
+   */
+  constructor() {
+    super({
+      buildQuery,
+      controlPanel,
+      loadChart: () => import('./EchartsBoxPlot'),
+      metadata: new ChartMetadata({
+        description: 'Box Plot using ECharts',
```
rusackas, line 42: Pedantic, but maybe worth saying "Apache ECharts" or "Apache ECharts (incubating)" in the description?

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -0,0 +1,178 @@
...
+    colorScheme,
+    groupby = [],
+    metrics: formDataMetrics,
+    numberFormat,
+    xTicksLayout,
+  } = formData as BoxPlotQueryFormData;
+  const colorFn = CategoricalColorNamespace.getScale(colorScheme as string);
+  const numberFormatter = getNumberFormatter(numberFormat);
+
+  // TODO: remove cast once metrics is cleaned up in QueryFormData
```
rusackas, line 44: Just making sure this isn't something you'd intended to tackle in this PR ;)

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -0,0 +1,178 @@
...
+          itemStyle: {
+            color: colorFn(groupbyLabel),
+          },
+        };
+      });
+    })
+    .flat(2);
+
+  let axisLabel;
+  if (xTicksLayout === '45°') axisLabel = { rotate: -45 };
```
rusackas, line 103: I always feel like I _should_ suggest `switch` in cases like this instead of `if else` but this seems plenty easy to read, and efficient enough.

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -0,0 +1,178 @@
...
+  else if (xTicksLayout === '90°') axisLabel = { rotate: -90 };
+  else if (xTicksLayout === 'flat') axisLabel = { rotate: 0 };
+  else if (xTicksLayout === 'staggered') axisLabel = { rotate: -45 };
+  else axisLabel = { show: true };
+
+  // @ts-ignore
+  const echartOptions: echarts.EChartOption<echarts.EChartOption.SeriesBoxplot> = {
+    grid: {
+      top: 30,
```
rusackas, line 112: Sometime, it might be nice to use `withTheme` here and base these values on `gridUnit`, for the record... but not in this PR!

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -0,0 +1,178 @@
...
+              `observations: ${numberFormatter(value[7])}`,
+            ];
+            if (value[8].length > 0) {
+              stats.push(`outliers: ${numberFormatter(value[8].length)}`);
+            }
+            return headline + stats.join('<br/>');
+          },
+        },
+      },
+      // @ts-ignore
```
rusackas, line 168: I won't be losing any sleep over these three `ts-ignore`s, but just thought I should confirm that they're not a "to do" that you wanted to pursue in this PR.

rusackas (approved):
Noted a few nits that I trust you'll address before merging, but I don't see anything blocking functionality, so I'll give the stamp of approval! Thanks for tackling this, this is awesome!!!!

plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
villebro: The whiskers actually affect what's returned in the data payload, so unfortunately `renderTrigger: true` won't work here.

plugins/plugin-chart-echarts/src/BoxPlot/index.ts
villebro: Very good point 👍

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
villebro: This requires a separate PR, as it will touch on many unit tests in `superset-ui/core`.

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
villebro: My python background forbids me from writing switches 🐍

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
villebro: The need for this is actually a shortcoming in ECharts that I'm hoping to PR there in the near future.

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
villebro: I'm hoping to tackle these when upgrading to ECharts 5.0, as the typing system should become slightly more stable in that release.

villebro:
Thanks for all the comments @rusackas , really appreciate it! ❤️

PR #803, opened 2020-10-07
fix(generator-superset): add fixes to viz plugin generator

villebro:
🐛 Bug Fix
Fix viz plugin generator for breaking changes introduced by upgrade to `0.15`. Also do minor cleanup to types.

Verified to work by creating Hello World plugin via `npm link` on `incubator-superset` (no TypeScript errors), running `yarn build` and `yarn jest`.


packages/generator-superset/generators/plugin-chart/templates/src/types.erb
```diff
@@ -0,0 +1,40 @@
+/**
```
rusackas, line 1: Might just want to add this to the file tree in the README... or remove the file tree from the README (not sure how much value it really provides)

packages/generator-superset/generators/plugin-chart/templates/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  "references": [
+    {
+      "path": "../../packages/superset-ui-chart-controls"
+    },
+    {
+      "path": "../../packages/superset-ui-core"
+    }
+  ]
+}
```
rusackas, line 25: Hmm... surprised it's not barking about a newline here...

rusackas (approved):
Couple o' nits, but not seeing anything blocking! Thanks for cleaning this up!

packages/generator-superset/generators/plugin-chart/templates/src/types.erb
villebro: Good catch. `types.ts` was actually already in the README, but for some reason file was missing before this (probably fell out during the long development/review process). However, `tsconfig.json` was missing, so I added that.

packages/generator-superset/generators/plugin-chart/templates/tsconfig.json
villebro: 👍  Added newline.

PR #804, opened 2020-10-08
chore(plugin-chart-echarts): rename boxplot type to whisker_type

villebro:
🏠 Internal
Rename `type` to `whisker_type` on BoxPlot post processing op.


rusackas (approved):
LGTM!

PR #806, opened 2020-10-13
fix(plugin-chart-echarts): improve ECharts visuals

villebro:
🏆 Enhancements

Improve ECharts plugin visuals. Changes:

## Pie
- Disable outside label line by default and add a control to enable it. Feature parity with NVD3 Pie.
- Change label color to black by default (looks less messy when labels inside). Feature parity with NVD3 Pie.
- Added customizable minimum percentage threshold (default: 5 %) to be able to remove labels for thin slices. Feature parity with NVD3 Pie.
- hover now doesn't change font size, only bolds the text (less invasive). Feature parity with NVD3 Pie.

## Timeseries
- Add y-axis truncate option (default: enabled).  Feature parity with NVD3 Line.
- Add y-axis bounds (min, max). Feature parity with NVD3 Line.

## Storybook
- Add Pie Chart case (`population`) with multiple thin slices to demonstrate minimum percentage threshold.
- Add missing controls to Pie and Timeseries.

Screenshot from Storybook demonstrating lower threshold with lots of thin slices:
image

Same with labels outside:
image

Screenshot of chart where a slice below the threshold is hovered above.
image

Truncate and bounds on y-axis:
truncate


plugins/plugin-chart-echarts/src/Pie/controlPanel.ts
```diff
@@ -107,8 +111,32 @@ const config: ControlPanelConfig = {
...
+              renderTrigger: true,
+              description: t('Draw line from Pie to label when labels outside?'),
+            },
+          },
+        ],
+        [
+          {
+            name: 'show_labels_threshold',
+            config: {
+              type: 'TextControl',
```
mistercrunch, line 129: NIT(optional/out-of-scope-for-this-pr): slider could be cool here eventually. I just discovered how cool the antd sliders are! https://ant.design/components/slider/

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -109,26 +112,32 @@ export default function transformProps(chartProps: ChartProps): EchartsTimeserie
             : 0,
       });
   });
+  let [yAxisMin, yAxisMax] = yAxisBounds || [];
+  if (yAxisMin === undefined && contributionMode === 'row' && stack) yAxisMin = 0;
+  if (yAxisMax === undefined && contributionMode === 'row' && stack) yAxisMax = 1;
```
mistercrunch, line 117: This section is fairly opaque and hard to read, wondering if/how much `ts` can help prevent issues around `undef` with more types hints here. Nothing outrageous though.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -109,26 +112,32 @@ export default function transformProps(chartProps: ChartProps): EchartsTimeserie
...
-      containLabel: true,
     },
     xAxis: { type: 'time' },
     yAxis: {
+      ...defaultYAxis,
       type: logAxis ? 'log' : 'value',
-      min: contributionMode === 'row' && stack ? 0 : undefined,
-      max: contributionMode === 'row' && stack ? 1 : undefined,
+      // these might be NaN which break the y axis
+      min: yAxisMin || undefined,
```
mistercrunch, line 131: related to the comment above, but what else than int/undef could it be? Also I think `0` would be switched to undef here as `0 || undefined === undefined` which seems like it's not desirable if someone wanted `0` to be the bound

plugins/plugin-chart-echarts/src/Pie/controlPanel.ts
villebro: Oh yeah, that Slider with input number looks perfect. I tinkered with the existing slider control, but it felt clunky as it doesn't allow for setting floats, which in this case is often relevant. Let's refactor the `SliderControl` to use AntD in a follow-up PR and change this when that's in place.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
villebro: True, will refactor.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
villebro: Whoops!

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -109,26 +112,41 @@ export default function transformProps(chartProps: ChartProps): EchartsTimeserie
             : 0,
       });
   });
+
+  // yAxisBounds sometimes starts returning NaNs, which messes up the u-axis
+  let [min, max] = (yAxisBounds || []).map((val?: number) =>
+    val !== undefined && Number.isNaN(val) ? undefined : val,
```
villebro, line 118: Apparently TS doesn't allow for `Number.isNaN(undefined)`, hence the seemingly unnecessary check for `undefined` here..

ktmud: You can probably just do `yAxisBounds.map(Number)` and use `Number.isNaN` check later.

ktmud (approved):
LGTM!

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
villebro: Thanks, looks much better!

plugins/plugin-chart-echarts/src/Pie/buildQuery.ts
```diff
@@ -19,5 +19,11 @@
 import { buildQueryContext, QueryFormData } from '@superset-ui/core';
 
 export default function buildQuery(formData: QueryFormData) {
-  return buildQueryContext(formData);
+  const { metric } = formData;
+  return buildQueryContext(formData, baseQueryObject => [
+    {
+      ...baseQueryObject,
+      orderby: [[metric, false]],
```
villebro, line 26: This is now needed to do descending ordering by the chosen metric: https://github.com/apache/incubator-superset/pull/11153

PR #808, opened 2020-10-14
fix(legacy-preset-chart-nvd3): custom yAxisFormat for contribution

ktmud:
🐛 Bug Fix

Currently it's not possible to change the number formatting for when "contribution" is checked or using percentage comparison in line charts.

Let's allow that.

### Test plan

1. Go to a line chart in Superset
2. Check contribution or select a percentage comparison
    
3. Change y axis format in "Customize"

The format should apply. If there is not format selected or is using the default "Adaptive Formatting", the format should be 1 decimal percentages.


kristw (approved):

kristw:
Thank you for taking care of this.

ktmud:
cc @villebro : we might want to bring feature parity for this bugfix to ECharts line chart, too.

villebro:
Thanks @ktmud , I'll add it to the backlog

PR #809, opened 2020-10-16
feat(core): add applied_time_extras to QueryObject

villebro:
🏆 Enhancements
Add `applied_time_extras` to `QueryObject` to be able to identify which filters have been set/overridden by `extra_filters`. Please see https://github.com/apache/incubator-superset/pull/11325 for additional context.

### DIFF
--- a/packages/superset-ui-core/src/query/extractExtras.ts
+++ b/packages/superset-ui-core/src/query/extractExtras.ts
@@ -6,6 +6,7 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
   const partialQueryObject: Partial<QueryObject> = {
     filters: formData.filters || [],
     extras: formData.extras || {},
+    applied_time_extras: {},
   };
 
   const reservedColumnsToQueryField: Record<string, keyof QueryObject> = {
@@ -20,6 +21,8 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
     if (filter.col in reservedColumnsToQueryField) {
       const queryField = reservedColumnsToQueryField[filter.col];
       partialQueryObject[queryField] = filter.val;
+      // @ts-ignore
+      partialQueryObject.applied_time_extras[filter.col] = filter.val;
     } else {
       // @ts-ignore
       partialQueryObject.filters.push(filter);
--- a/packages/superset-ui-core/src/query/extractExtras.ts
+++ b/packages/superset-ui-core/src/query/extractExtras.ts
@@ -1,15 +1,18 @@
 /* eslint-disable camelcase */
 import { isDruidFormData, QueryFormData } from './types/QueryFormData';
 import { QueryObject } from './types/Query';
+import { AppliedTimeExtras, TimeColumnConfigKey } from './types/Time';
 
 export default function extractExtras(formData: QueryFormData): Partial<QueryObject> {
+  const applied_time_extras: AppliedTimeExtras = {};
+  const { extras = {}, filters = [] } = formData;
   const partialQueryObject: Partial<QueryObject> = {
-    filters: formData.filters || [],
-    extras: formData.extras || {},
-    applied_time_extras: {},
+    filters,
+    extras,
+    applied_time_extras,
   };
 
-  const reservedColumnsToQueryField: Record<string, keyof QueryObject> = {
+  const reservedColumnsToQueryField: Record<TimeColumnConfigKey, keyof QueryObject> = {
     __time_range: 'time_range',
     __time_col: 'granularity_sqla',
     __time_grain: 'time_grain_sqla',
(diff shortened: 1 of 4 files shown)


pkdotson (approved):
LGTM!

villebro:
Thanks @pkdotson !

packages/superset-ui-core/src/query/extractExtras.ts
```diff
@@ -20,6 +21,8 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
     if (filter.col in reservedColumnsToQueryField) {
       const queryField = reservedColumnsToQueryField[filter.col];
       partialQueryObject[queryField] = filter.val;
+      // @ts-ignore
+      partialQueryObject.applied_time_extras[filter.col] = filter.val;
```
ktmud, line 25: Instead of `ts-ignore`, can use type assertion somewhere to make the typings compatible?

```typescript
      partialQueryObject.applied_time_extras[filter.col as QueryObjectAppliedTimeExtraKey] = filter.val;
```

packages/superset-ui-core/src/query/types/Query.ts
```diff
@@ -41,11 +41,24 @@ export type QueryObjectExtras = Partial<{
   where?: string;
 }>;
 
+export type QueryObjectAppliedTimeExtraKey =
```
ktmud, line 44: Can we rename this to something more generic such as `TimeColumnConfigKey`? Also, maybe put this into `Time.ts` ?

packages/superset-ui-core/src/query/extractExtras.ts
```diff
@@ -6,6 +6,7 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
   const partialQueryObject: Partial<QueryObject> = {
     filters: formData.filters || [],
     extras: formData.extras || {},
+    applied_time_extras: {},
   };
 
   const reservedColumnsToQueryField: Record<string, keyof QueryObject> = {
```
ktmud, line 12: Looks like this could be `Record<QueryObjectAppliedTimeExtraKey, keyof QueryObject>` or `Record<TimeColumnConfigKey, keyof QueryObject>`.

packages/superset-ui-core/src/query/types/Query.ts
```diff
@@ -41,11 +41,24 @@ export type QueryObjectExtras = Partial<{
...
+export type QueryObjectAppliedTimeExtras = {
+  QueryObjectAppliedTimeExtraKey?: string;
+};
+
 export type ResidualQueryObjectData = {
   [key: string]: unknown;
 };
 
 export type QueryObject = {
+  /** Extra filters that have been applied to the query object */
```
ktmud, line 60: This comment doesn't seem to indicate how is this is different from `extra`.

ktmud (commented):
Added a couple of naming suggestions. I'm also not fully getting what is this used for. Can you also add more context on how is this going to be used in `superset-frontend` code (maybe a linked PR?) and what are we constrained with if we don't do this?

villebro:
Thanks for comments @ktmud , I'll add some more context to make this clearer and do the suggested code improvements. This is something that is needed to be able to complete the p0 filter indicator feature (filter indicators will otherwise not work with the chart data endpoint). I envision this and other related structures, namely how ad-hoc filters are transformed into simple filters + static `where`s and `having`s will need to change soon when we start working on native filter components. Therefore I was hoping to make this slightly simpler so that we don't have to do a much more comprehensive refactor of this code up front.

packages/superset-ui-core/src/query/types/Query.ts
```diff
@@ -41,11 +41,24 @@ export type QueryObjectExtras = Partial<{
...
 
+export type QueryObjectAppliedTimeExtraKey =
+  | '__time_col'
+  | '__time_grain'
+  | '__time_range'
+  | '__time_origin'
+  | '__granularity';
+
+export type QueryObjectAppliedTimeExtras = {
+  QueryObjectAppliedTimeExtraKey?: string;
```
suddjian, line 52: I don't think this type is doing what you think it's doing

suddjian:
I'm not sure what `applied_extra_filters` is meant to be used for. Is it a mapping to column names?

packages/superset-ui-core/src/query/extractExtras.ts
```diff
@@ -1,14 +1,18 @@
...
 
 export default function extractExtras(formData: QueryFormData): Partial<QueryObject> {
+  const applied_time_extras: Partial<Record<TimeColumnConfigKey, keyof QueryObject>> = {};
+  const { extras = {}, filters = [] } = formData;
   const partialQueryObject: Partial<QueryObject> = {
-    filters: formData.filters || [],
-    extras: formData.extras || {},
+    filters,
+    extras,
+    applied_time_extras,
```
villebro, line 12: Bycatch: change to use destructuring here.

packages/superset-ui-core/src/query/extractExtras.ts
```diff
@@ -18,35 +22,30 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
...
+      filters.push(filter);
     }
   });
 
   // map to undeprecated names and remove deprecated fields
   if (isDruidFormData(formData) && !partialQueryObject.druid_time_origin) {
-    partialQueryObject.extras = {
-      druid_time_origin: formData.druid_time_origin,
-    };
+    extras.druid_time_origin = formData.druid_time_origin;
```
villebro, line 36: bycatch: better to add to the object than discard the original one.

packages/superset-ui-core/src/query/extractExtras.ts
```diff
@@ -18,35 +22,30 @@ export default function extractExtras(formData: QueryFormData): Partial<QueryObj
...
     delete partialQueryObject.druid_time_origin;
   } else {
     // SQL
-    partialQueryObject.extras = {
-      ...partialQueryObject.extras,
-      time_grain_sqla: partialQueryObject.time_grain_sqla || formData.time_grain_sqla,
-    };
+    extras.time_grain_sqla = partialQueryObject.time_grain_sqla || formData.time_grain_sqla;
```
villebro, line 40: same here

villebro:
@suddjian and @ktmud thanks for your comments, I did my best to address them. I added a link to the `incubator-superset` PR, which implements these changes for the new chart data endpoint (also implements the same for the legacy endpoint).

suddjian (approved):
lgtm

ktmud (approved):
LGTM

PR #810, opened 2020-10-19
fix(plugin-chart-echarts): fix default y axis bounds

villebro:
🐛 Bug Fix
The default `yAxisBounds` are currently set to `null`, which evaluate to zero when passed to the `Number` function: (`Number(null) === 0`). The defaults were copied from the NVD3 control panel, but apparently there evaluating to zero isn't a problem.

### AFTER
image

### BEFORE
image


rusackas (approved):
LGTM

PR #812, opened 2020-10-27
feat(core): add support for annotations and url params

villebro:
Add support for `url_params` and `annotation_layers` to `QueryFormData` and `QueryObject`. The `AnnotationLayer` schema corresponds to that generated by `AnnotationLayerControl` in `superset-ui`.


suddjian (approved):
nice!

ktmud:
@villebro can you shed light on how `url_params` is used by which component in Superset?

PR #815, opened 2020-10-28
fix(plugin-chart-echarts): improve yAxisBounds parsing

villebro:
🐛 Bug Fix
Currently the `YAxisBoundsControl` will initialize with `null` values despite being passed `undefined`, causing trouble when setting the y axis bounds. This improves the parsing of said values prior to using them so Y-axis bounds are always either undefined or valid values.

### Before
image

### After
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/utils/controls.ts
+++ b/plugins/plugin-chart-echarts/src/utils/controls.ts
@@ -1,0 +1,25 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+// eslint-disable-next-line import/prefer-default-export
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-echarts/test/utils/controls.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    expect(parseYAxisBound('abc')).toBeUndefined();
+  });
+
+  it('should return numeric value for valid values', () => {
+    expect(parseYAxisBound(0)).toEqual(0);
+    expect(parseYAxisBound('0')).toEqual(0);
+    expect(parseYAxisBound(1)).toEqual(1);
+    expect(parseYAxisBound('1')).toEqual(1);
+    expect(parseYAxisBound(10.1)).toEqual(10.1);
+    expect(parseYAxisBound('10.1')).toEqual(10.1);
```
villebro, line 35: AFAIK the `YAxisBoundsControl` doesn't actually return strings, but I decided to parse string values, just in case.

plugins/plugin-chart-echarts/src/utils/controls.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+// eslint-disable-next-line import/prefer-default-export
+export function parseYAxisBound(bound?: string | number | null): number | undefined {
+  if (bound === undefined || bound === null || Number.isNaN(Number(bound))) {
```
rusackas, line 21: undefined returns NaN, so you don't technically need that, but if it helps for code-reading clarity, so be it :)

rusackas (approved):
LGTM!

plugins/plugin-chart-echarts/src/utils/controls.ts
villebro: Thanks for the review! Just to make sure I understood correctly: That last bit is to catch unparseable strings and `NaN`s (`undefined was already checked at the beginning). If this can be written more compactly I'm happy to learn+change!

PR #816, opened 2020-10-28
fix(plugin-chart-echarts): disable pie chart animation

villebro:
🏆 Enhancements
Disable animation in Pie Chart.

### Before
piebefore

### After
pieafter


ktmud (approved):
👏

PR #817, opened 2020-10-28
feat(plugin-chart-echarts): add support for formula annotations

villebro:
🏆 Enhancements

Add initial support for annotation layers to ECharts time-series chart by implementing the formula annotation layer. This has feature parity with the current NVD3 formula annotation layer. Event, interval and time-series annotations to follow later.

image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
@@ -61,8 +61,27 @@ const config: ControlPanelConfig = {
       ],
     },
     {
+      label: t('Annotations and Layers'),
+      expanded: false,
+      controlSetRows: [
+        [
+          {
+            name: 'annotation_layers',
+            config: {
+              type: 'AnnotationLayerControl',
+              label: '',
+              default: [],
+              description: 'Annotation Layers',
+              renderTrigger: true,
+              tabOverride: 'data',
+            },
+          },
+        ],
+      ],
+    },
+    {
       label: t('Predictive Analytics'),
-      expanded: true,
+      expanded: false,
       controlSetRows: [
         [
           {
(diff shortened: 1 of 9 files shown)


plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
```diff
@@ -60,9 +60,28 @@ const config: ControlPanelConfig = {
...
+              tabOverride: 'data',
+            },
+          },
+        ],
+      ],
+    },
     {
       label: t('Predictive Analytics'),
-      expanded: true,
+      expanded: false,
```
villebro, line 84: This should have never been expanded by default.

plugins/plugin-chart-echarts/src/utils/annotation.ts
```diff
@@ -0,0 +1,46 @@
...
+): [Date, number][] {
+  const { value } = formula;
+  const node = mathjsParse(value);
+  const func = node.compile();
+  return data.map(row => {
+    return [new Date(Number(row.__timestamp)), func.evaluate({ x: row.__timestamp }) as number];
+  });
+}
+
+export function parseAnnotationOpacity(opacity?: AnnotationOpacity): number | undefined {
```
mistercrunch, line 35: I think you'll know this will always return a number, you can remove `undefined` as something it might return

villebro: Oh yeah, not sure why that slipped in.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -113,6 +117,36 @@ export default function transformProps(chartProps: ChartProps): EchartsTimeserie
...
+  annotationLayers.forEach((layer: AnnotationLayer) => {
+    const {
+      name,
+      color,
+      opacity: annotationOpacity,
+      width: annotationWidth,
+      show: annotationShow,
+      style,
+    } = layer;
+    if (isFormulaAnnotationLayer(layer) && annotationShow) {
```
mistercrunch, line 129: NIT: would flip the two as `annotationShow` is certainly cheaper to evaluate and js won't even call `isFormulaAnnotationLayer` if it's `false`

mistercrunch (approved):
A few minor NITs, otherwise LGTM

PR #818, opened 2020-10-28
feat: update time_compare description and choices

etr2460:
🏆 Enhancements

Seems like there are a bunch of files with the same controls as https://github.com/apache/incubator-superset/pull/11294, so i'm making the changes there throughout superset-ui

to: @pkdotson @ktmud @kristw


rusackas (approved):
LGTM!

PR #819, opened 2020-10-30
fix(chart-controls): move @superset-ui/core to a peerDependency

suddjian:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

I _think_ this fixes the warnings we're seeing in the superset test suite, but I'm not totally sure because `npm link` hasn't been working for me

Screen Shot 2020-10-29 at 11 44 03 PM

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart-controls/package.json
+++ b/packages/superset-ui-chart-controls/package.json
@@ -26,11 +26,11 @@
     "access": "public"
   },
   "dependencies": {
-    "@superset-ui/core": "0.15.10",
     "lodash": "^4.17.15",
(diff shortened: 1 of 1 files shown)


ktmud:
Where were you seeing the warning? And how did `npm link` not work for you? Did you `npm link` only one of superset-ui/core and superset-ui/chart-controls? If you edit one of them, you'd need to link both.

suddjian:
I get these warnings when running `npm test` on incubator-superset. You can see them in the frontend CI output.

The linking itself worked, but I got a different set of errors when I ran the tests.

PR #820, opened 2020-10-31
feat(plugin-chart-echarts): make boxplot groupby optional

villebro:
🏆 Enhancements

Make Boxplot groupby optional. Also replace high and low with max and min respectively as is convention in Boxplots and make tooltip labels more verbose.

### Screenshot
When groupby is unspecified, leave x axis empty and remove headline from tooltip. Also notice updated tooltip.
image


villebro:
Ping @bkyryliuk

bkyryliuk (approved):
thanks!

mistercrunch (approved):

PR #823, opened 2020-11-02
fix(plugin-chart-echarts): fix broken boxplot storybook example

villebro:
🐛 Bug Fix
The recent PR #820 introduced naming changes that were not updated to the Storybook data, breaking the Box Plot example.

### Before
image

### After
image


suddjian (approved):
yup

PR #825, opened 2020-11-03
chore(chart-controls): clean up typing

ktmud:
After [recent changes](https://github.com/apache/incubator-superset/pull/11435) to chart controls in Superset, some typing
has become obsolete. Let's clean it up.

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -149,9 +149,6 @@ export type InternalControlType =
   | 'MetricsControl'
   | 'AdhocFilterControl'
   | 'FilterBoxItemControl'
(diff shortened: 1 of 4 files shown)


packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -200,11 +197,7 @@ export interface BaseControlConfig<
   warning?: ReactNode;
   error?: ReactNode;
   // override control panel state props
-  mapStateToProps?: (
-    state: ControlPanelState,
-    control: this,
-    actions?: ControlPanelActionDispatchers,
```
ktmud, line 206: The `actions` argument was once provided, but has since removed after some refactoring.

villebro (approved):

PR #828, opened 2020-11-05
feat(plugin-chart-echarts): implement event interval and timeseries annotations

villebro:
Add support for event, interval and time series annotations to ECharts Timeseries chart.

image


packages/superset-ui-core/src/query/types/AnnotationLayer.ts
```diff
@@ -54,6 +56,11 @@ export type IntervalAnnotationLayer = BaseAnnotationLayer &
     annotationType: 'INTERVAL';
   };
 
+export type TableAnnotationLayer = BaseAnnotationLayer &
+  TableSourceAnnotationLayer & {
+    annotationType: 'EVENT' | 'INTERVAL';
```
ktmud, line 61: Can we give `annotationType` its own type? Either a string union literal type or `Enum` would be fine.

```ts
type AnnotationType = 'Event' | 'Interval' | ...

interface BaseAnnotationLayer {
  annotationType: AnnotationType
}
```

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -127,24 +144,140 @@ export default function transformProps(chartProps: ChartProps): EchartsTimeserie
...
-          type: style,
-          width: annotationWidth,
-        },
-        type: 'line',
-        smooth: true,
-        // @ts-ignore
-        data: evalFormula(layer, data),
-        symbolSize: 0,
-      });
+    if (annotationShow) {
```
ktmud, line 147: These if branches (and the transform function itself) are growing a little bit too big, do you think if it makes sense to split some pieces into smaller functions? For example, maybe each annotation transformer can have its own function?

ktmud (commented):
Super useful feature! Glad to see the ECharts-powered visualization becomes more and more powerful. Added a couple of comments on typing and code structure.

packages/superset-ui-core/src/query/types/AnnotationLayer.ts
villebro: Good idea

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
villebro: You're right, this was probably overdue, will split it up.

villebro:
@ktmud I've split up the series and annotation transformers out of the main loop and replaced the enum-like fields with enums. There's still potential for more refactoring, but this feels like an ok first step that can be improved upon in subsequent PRs (there will undoubtedly be follow up PRs to this).

villebro:
@ktmud it would be great to get a re-review of this PR so we can introduce this functionality.

ktmud (approved):
LGTM! Sorry for the delay. Totally missed the notification on this one.

PR #831, opened 2020-11-09
feat(legacy-plugin-chart-sankey): allow sort by metric

ktmud:
🏆 Enhancements

Adding a "Sort by metric" checkbox control for Sankey chart. If selected, the query results will be sorted by metric in descending order. This is useful when there are many rows of data but it's not easy to regroup source and targets via custom SQL queries (e.g. when working with a datasource that does not support arbitrary SQL queries).

Also added a caveat warning for row limit:

### DIFF
--- a/plugins/legacy-plugin-chart-sankey/src/controlPanel.ts
+++ b/plugins/legacy-plugin-chart-sankey/src/controlPanel.ts
@@ -17,13 +17,46 @@
  * under the License.
  */
 import { t } from '@superset-ui/core';
+import { ControlPanelConfig } from '@superset-ui/chart-controls';
 
-export default {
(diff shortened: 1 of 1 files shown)


villebro:
A by-thought: if there are many orphaned paths (=both source and target only used in a single path), I wonder if these should be excluded from the chart (not sure what the behaviour is currently).

villebro (approved):

ktmud:
> A by-thought: if there are many orphaned paths (=both source and target only used in a single path), I wonder if these should be excluded from the chart (not sure what the behaviour is currently).

I think they would just be considered part of the flow at the first level. Even thought the source might actually be at the second or later level if there's no row limit... That's why a warning about row limit is important.

PR #832, opened 2020-11-11
chore: bump mathjs to latest version on all plugins

villebro:
Bump `mathjs` to the latest version (`8.0.1`) and make required code changes. Verified to work locally.

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/package.json
+++ b/plugins/legacy-preset-chart-nvd3/package.json
@@ -36,7 +36,7 @@
     "dompurify": "^2.0.6",
     "fast-safe-stringify": "^2.0.6",
     "lodash": "^4.17.11",
-    "mathjs": "^3.20.2",
+    "mathjs": "^8.0.1",
     "moment": "^2.20.1",
     "nvd3-fork": "2.0.3",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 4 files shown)


rusackas (approved):

ktmud:
do we know if the new version breaks any of the existing formulas? (Happy to pull some sample data from Airbnb users if needed)

villebro:
@ktmud I don't believe the evaluation logic itself has changed, only the function name. I tested with some simple ones `sin(x)`, constant values etc, all worked fine.

villebro:
..having said that, I'd be thankful for some real test cases so I can add them to the test suite. Btw, I'll be moving all annotation logic to `superset-ui-core` from the ECharts plugin once this development stabilizes.

ktmud:
Just pulled example formulas from our production data and didn't see any complex formulas we should be worry about:

```
SELECT
  id,
  slice_name,
  viz_type,
  json_extract_scalar(params, '$.annotation_layers[0].value'),
  params 
FROM "db_exports"."superset_production_slices_v01"
WHERE "ds" = '2020-11-11'
  AND params like '%FORMULA%'
  AND not regexp_like(json_extract_scalar(params, '$.annotation_layers[0].value'), '^[\d\.]*$')
LIMIT 1000
```

Of tens of thousands of slices, there are only ~100 charts used real computational formulas (that is not a static number) in a FORMULA annotation layer.

PR #833, opened 2020-11-11
feat(plugin-chart-echarts): add legend customization options

villebro:
🏆 Enhancements
Currently ECharts has limited functionality for automatically resizing the chart area based on the legend, causing overflowing. This PR adds the possibility to
- Enable/disable legend on Timeseries chart (default: disabled)
- Change legend type (scroll, plain) (default: scroll)
- Change orientation of legend (top, bottom, left, right) on Pie and Timeseries (default: top)
- Define custom legend padding (default: 20 for top/bottom, 170 for left/right)

The custom padding option will be removed once automatic component resizing is implemented in ECharts. Also cleans up default props for ECharts Pie (should really be in a separate PR, apologies for that).

### Before
Currently there's no way of disabling the legend
image

### After
Now the legend is disabled by default
image

When enabled, it defaults to scroll mode
image

The legend can also be moved to top/right/bottom/left:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
@@ -16,8 +16,27 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+import React from 'react';
 import { t, validateNonEmpty } from '@superset-ui/core';
 import { ControlPanelConfig, D3_FORMAT_OPTIONS, sections } from '@superset-ui/chart-controls';
+import { DEFAULT_FORM_DATA } from './types';
+import {
+  legendMarginControl,
+  legendOrientationControl,
+  legendTypeControl,
+  showLegendControl,
+} from '../controls';
+
+const {
+  donut,
+  innerRadius,
+  labelsOutside,
+  labelType,
+  labelLine,
+  outerRadius,
+  numberFormat,
+  showLabels,
+} = DEFAULT_FORM_DATA;
 
 const noopControl = { name: 'noop', config: { type: '', renderTrigger: true } };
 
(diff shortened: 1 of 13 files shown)


plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
```diff
@@ -147,10 +154,23 @@ const config: ControlPanelConfig = {
...
+        ],
+        [
+          {
+            name: 'donut',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Donut'),
+              default: donut,
+              renderTrigger: true,
+              description: t('Do you want a donut or a pie?'),
```
rusackas, line 171: Jeez... asking the hard questions here, man!

rusackas: At some point, it might be good to make a custom "switch" component for when checkboxes just won't do:

donutpie

plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
```diff
@@ -16,8 +16,28 @@
...
+  labelsOutside,
+  labelType,
+  labelLine,
+  outerRadius,
+  numberFormat,
+  showLabels,
+  showLabelsThreshold,
+} = DEFAULT_FORM_DATA;
 
 const noopControl = { name: 'noop', config: { type: '', renderTrigger: true } };
```
rusackas, line 42: I haven't used this before... looks like it's just a place holder to adjust layout? A little weird, but... ¯\\\_(ツ)_/¯

villebro: Yes, with the new explore layout we always want to make the control fill the entire row (on the old one some controls looked funny due to being too wide). I'll update accordingly.

junlincc:
let's get this in for v.1.0.1. @villebro thank you Ville! 
https://github.com/apache-superset/superset-roadmap/issues/92 @maloun96 Victor, I'm moving this roadmap item to you, please apply all the changes in this PR to all charts(nvd3+echarts)  Thank you!!

plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -86,3 +88,63 @@ export function extractGroupbyLabel({
...
 }
+
+export function getLegendProps(
+  type: LegendType,
+  orientation: LegendOrientation,
+  show: boolean,
+): echarts.EChartOption.Legend {
+  const legend: echarts.EChartOption.Legend = {
+    orient:
+      orientation === LegendOrientation.Top || orientation === LegendOrientation.Bottom
```
maloun96, line 99: I would prefer using includes

```suggestion
      [LegendOrientation.Top, LegendOrientation.Bottom].includes(orientation)
```

junlincc (approved):
lgtm! this is great!

PR #834, opened 2020-11-12
fix(plugin-chart-echarts): default to 45 degree rotated labels

villebro:
🐛 Bug Fix
Default to 45 degree rotated x-axis labels

### BEFORE
image

### AFTER
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
@@ -105,7 +105,7 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
   else if (xTicksLayout === '90°') axisLabel = { rotate: -90 };
   else if (xTicksLayout === 'flat') axisLabel = { rotate: 0 };
   else if (xTicksLayout === 'staggered') axisLabel = { rotate: -45 };
-  else axisLabel = { show: true };
+  else axisLabel = { rotate: -45 };
 
   // @ts-ignore
   const echartOptions: echarts.EChartOption<echarts.EChartOption.SeriesBoxplot> = {


plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -105,7 +105,7 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
   else if (xTicksLayout === '90°') axisLabel = { rotate: -90 };
   else if (xTicksLayout === 'flat') axisLabel = { rotate: 0 };
   else if (xTicksLayout === 'staggered') axisLabel = { rotate: -45 };
-  else axisLabel = { show: true };
+  else axisLabel = { rotate: -45 };
```
ktmud, line 108: Can we be smart and only do rotation if there are more than x number of x ticks?

suddjian: To really be smart it would have to be something like ticks / width

junlincc: can we just be consistent.? bar chart has it rotate 45 by default.

suddjian (approved):
US politics is so fucked up that just the number 45 triggers me now.

junlincc:
> US politics is so fucked up that just the number 45 triggers me now.

it's rotating...

plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
```diff
@@ -105,7 +105,7 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
   else if (xTicksLayout === '90°') axisLabel = { rotate: -90 };
   else if (xTicksLayout === 'flat') axisLabel = { rotate: 0 };
   else if (xTicksLayout === 'staggered') axisLabel = { rotate: -45 };
```
ktmud, line 107: Since we have a `xTicksLayout` control already? Why not just set the default of `xTicksLayout` to `rotate: -45`?

PR #836, opened 2020-11-16
refactor: Relocating/renaming various chart controls

rusackas:
💔 Breaking Changes

🏆 Enhancements
Takes advantage of tweaks to `tabOverride` in control layouts from https://github.com/apache/incubator-superset/pull/11708

I'll post screenshots inline with code comments below, so it's not quite so overwhelming in the description.

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart-controls/src/sections.tsx
+++ b/packages/superset-ui-chart-controls/src/sections.tsx
@@ -101,7 +101,6 @@ export const annotations = {
           default: [],
           description: 'Annotation Layers',
           renderTrigger: true,
-          tabOverride: 'data',
         },
       },
     ],
--- a/packages/superset-ui-core/src/color/CategoricalColorScale.ts
+++ b/packages/superset-ui-core/src/color/CategoricalColorScale.ts
@@ -68,7 +68,7 @@ class CategoricalColorScale extends ExtensibleFunction {
    * @returns an object where the key is the data value and the value is the hex color code
    */
   getColorMap() {
-    const colorMap: { [key: string]: string } = {};
+    const colorMap: { [key: string]: string | undefined } = {};
     this.scale.domain().forEach(value => {
       colorMap[value.toString()] = this.scale(value);
     });
--- a/packages/superset-ui-core/src/color/SequentialScheme.ts
+++ b/packages/superset-ui-core/src/color/SequentialScheme.ts
@@ -52,6 +52,7 @@ export default class SequentialScheme extends ColorScheme {
     const piecewiseScale: (t: number) => string = piecewise(interpolateHcl, this.colors);
     const adjustExtent = scaleLinear().range(extent).clamp(true);
 
+    // @ts-ignore
     return quantize<string>(t => piecewiseScale(adjustExtent(t)), numColors);
   }
 }
--- a/plugins/legacy-plugin-chart-calendar/src/controlPanel.js
+++ b/plugins/legacy-plugin-chart-calendar/src/controlPanel.js
@@ -62,6 +62,7 @@ export default {
     {
       label: t('Chart Options'),
       expanded: true,
+      tabOverride: 'customize',
       controlSetRows: [
         ['linear_color_scheme'],
         [
--- a/plugins/legacy-plugin-chart-country-map/src/controlPanel.ts
+++ b/plugins/legacy-plugin-chart-country-map/src/controlPanel.ts
@@ -27,8 +27,9 @@ export default {
       controlSetRows: [['entity'], ['metric'], ['adhoc_filters']],
     },
     {
-      label: t('Options'),
+      label: t('Chart Options'),
       expanded: true,
+      tabOverride: 'customize',
       controlSetRows: [
         [
           {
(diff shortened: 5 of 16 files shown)


packages/superset-ui-chart-controls/src/sections.tsx
```diff
@@ -101,7 +101,6 @@ export const annotations = {
           default: [],
           description: 'Annotation Layers',
           renderTrigger: true,
-          tabOverride: 'data',
```
rusackas, line 104: Removing this here as it's support has been stripped in the PR on `incubator-superset` - this section will remain on the Data tab anyway.

packages/superset-ui-core/src/color/SequentialScheme.ts
```diff
@@ -52,6 +52,7 @@ export default class SequentialScheme extends ColorScheme {
     const piecewiseScale: (t: number) => string = piecewise(interpolateHcl, this.colors);
     const adjustExtent = scaleLinear().range(extent).clamp(true);
 
+    // @ts-ignore
```
rusackas, line 55: Was seeing typescript errors here, but wasn't feeling inclined to tackle it in this PR. Suggestions welcome :)

packages/superset-ui-core/src/color/CategoricalColorScale.ts
```diff
@@ -68,7 +68,7 @@ class CategoricalColorScale extends ExtensibleFunction {
    * @returns an object where the key is the data value and the value is the hex color code
    */
   getColorMap() {
-    const colorMap: { [key: string]: string } = {};
+    const colorMap: { [key: string]: string | undefined } = {};
```
rusackas, line 71: Just trying to get TS to quiet down here.

plugins/legacy-plugin-chart-country-map/src/controlPanel.ts
```diff
@@ -88,9 +89,9 @@ export default {
   ],
   controlOverrides: {
     entity: {
-      label: t('ISO 3166-2 codes of region/province/department'),
+      label: t('ISO 3166-2 Codes'),
```
rusackas, line 92: image

plugins/legacy-plugin-chart-country-map/src/controlPanel.ts
```diff
@@ -27,8 +27,9 @@ export default {
       controlSetRows: [['entity'], ['metric'], ['adhoc_filters']],
     },
     {
-      label: t('Options'),
+      label: t('Chart Options'),
```
rusackas, line 30: image

plugins/legacy-plugin-chart-calendar/src/controlPanel.js
```diff
@@ -62,6 +62,7 @@ export default {
     {
       label: t('Chart Options'),
       expanded: true,
+      tabOverride: 'customize',
```
rusackas, line 65: Screen Shot 2020-11-13 at 4 41 59 PM

plugins/legacy-plugin-chart-event-flow/src/controlPanel.tsx
```diff
@@ -38,7 +38,7 @@ const config: ControlPanelConfig = {
             name: 'all_columns_x',
             config: {
               type: 'SelectControl',
-              label: t('Column containing event names'),
+              label: t('Event Names'),
```
rusackas, line 41: image

plugins/legacy-plugin-chart-event-flow/src/controlPanel.tsx
```diff
@@ -89,7 +89,7 @@ const config: ControlPanelConfig = {
       controlSetRows: [['adhoc_filters']],
     },
     {
-      label: t('Additional metadata'),
+      label: t('Metadata'),
```
rusackas, line 92: image
 (this is admittedly a bit redundant with the matching headers)

ktmud: what is the metadata used for? Maybe we can make the title even more explanatory?

junlincc: hey Evan, we shouldn't remove "Additional" here, adding an addition meta data is not gonna affect the chart/query itself, it's just adding an additional column to the table generated by the query... so, let's not remove 'additional'. but we should have a better tooltip for sure.. also, please keep the Event definition section open by default, right now, it's closed by default. 
Chris W made this chart, it can be very useful for monitoring event status change but not many people are using it or understand how to use it.... 
Screen Shot 2020-11-18 at 5 55 48 PM

@ktmud any thoughts?

ktmud: Let's keep the "Additional metadata" section title for now. In the future, things like this should probably be in a "Tooltip" section in chart customization.

rusackas, line 92:
```suggestion
      label: t('Additional metadata'),
```

rusackas:
Opening this one for review, as it's been lingering for a while! Maybe we can merge these changes in, bump the versions, and start another PR with additional tweaks on other plugins.

rusackas:
@ktmud would you mind giving this round of tweaks a review? More to come on the heels of this one :)

ktmud (approved):

ktmud:
LGTM!

PR #839, opened 2020-11-25
fix(legacy-plugin-chart-pivot-table): remove nulls from table

villebro:
🐛 Bug Fix
Currently null values are displayed as `null` in the table. This changes the label to empty, and treats the null value as `Number.NEGATIVE_INFINITY` on sort.

### AFTER
image

### BEFORE
image


rusackas (approved):
Well, you don't get to use `NEGATIVE_INFINITY` every day.

If only you got to display `-∞` for fun.

LGTM.

PR #841, opened 2020-11-26
fix(plugin-chart-word-cloud): ensure top results are always displayed

agatapst:
While resizing chart sometimes top results were filtered out because their sizes were too big. This
solution makes sure that top 10% of results will always be displayed by gradually scaling down the
chart if needed.

fix apache/incubator-superset#11784

🐛 Bug Fix
Before:
before_resizing

After:
resizing_after

Issues associated with this bug were reported in D3 library and not completely solved yet.
[- most important words (highest counts) missing if they don't fit](https://github.com/jasondavies/d3-cloud/issues/36)
[- missing words in final layout](https://github.com/jasondavies/d3-cloud/issues/159)
The cost of my workaround solution may be slower pace of displaying the chart while resizing.

cc @villebro

### DIFF
--- a/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
+++ b/plugins/plugin-chart-word-cloud/src/chart/WordCloud.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import cloudLayout, { Word } from 'd3-cloud';
-import { PlainObject, createEncoderFactory, DeriveEncoding } from 'encodable';
+import { PlainObject, createEncoderFactory, DeriveEncoding, Encoder } from 'encodable';
 import { SupersetThemeProps, withTheme, seedRandom } from '@superset-ui/core';
 
 export const ROTATION = {
@@ -38,6 +38,7 @@ export interface WordCloudProps extends WordCloudVisualProps {
 
 export interface WordCloudState {
   words: Word[];
+  scaleFactor: number;
 }
 
 const defaultProps: Required<WordCloudVisualProps> = {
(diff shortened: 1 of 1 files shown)


junlincc:
Thanks for the quick fix @agatapst! this is awesome! 
a bit context for you, we are in the process of migrating from D3 charts to [Echarts](https://echarts.apache.org/examples/en/index.html), and have Echarts be Superset's main charting library. luckily, Echarts does offer wordcloud as well, see link https://github.com/ecomfe/echarts-wordcloud. 
Due to D3's limitations in wordcloud, instead of further invest in improving it, next step will be testing Echarts wordcloud to see if it reaches full feature parity and if so, then to migrate. this is not a priority yet, but I would like to get very familiar with Data Viz in Superset if you are interested. meet @villebro here, he will be your go to person for any questions. :)

villebro (commented):
Thanks @agatapst , looks really good! While this will have a performance impact, I think it's worth it given how misleading the results are without this fix. 

A few suggestions:
- I've noticed that Word Cloud is especially poor at rendering long "words". For instance, The value "Latin America and Caribbean Islands" will tend to not render at all if it's the only value unless the chart is really big. So having a story in the story book with a few really long words would be great.
- The logic for ensuring that at minimum the first 10 % values are present is slightly difficult to follow. Do you think we could break out the relevant utility function(s) and add a few unit tests to make sure it works with e.g. 1, 5, 20 and 100 values, just to make sure the rounding works correctly?

agatapst:
@villebro Thanks for your reply, I will think that through and come back with some solution soon. :)

villebro:
@agatapst this would be great to get merged, but needs a rebase.

agatapst:
@villebro sorry for the delay! it's rebased now

villebro:
Thanks @agatapst!

agatapst:
@villebro all checks have passed, if PR is ok without any changes, it's ready to merge
I can come back to your suggestions later, because tbh there always were tasks with higher priority to work on - but if you think this PR should be improved asap, I can work on it today and tomorrow

ktmud (approved):
Seems like a reasonable solution.

PR #846, opened 2020-12-01
feat: support multi queries request

simchaNielsen:
🏆 Enhancements

Gives opportunity to get multi queries response results from incubator:
See PR: https://github.com/apache/incubator-superset/pull/11880/files

### DIFF
--- a/packages/superset-ui-core/src/chart/models/ChartProps.ts
+++ b/packages/superset-ui-core/src/chart/models/ChartProps.ts
@@ -47,6 +47,8 @@ export interface ChartPropsConfig {
   hooks?: Hooks;
   /** Formerly called "payload" */
   queryData?: QueryData;
+  /** Formerly called "payload" */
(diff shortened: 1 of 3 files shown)


amitmiran137 (approved):

packages/superset-ui-core/src/chart/models/ChartProps.ts
```diff
@@ -47,6 +47,8 @@ export interface ChartPropsConfig {
   hooks?: Hooks;
   /** Formerly called "payload" */
   queryData?: QueryData;
+  /** Formerly called "payload" */
+  queriesData?: QueryData[];
```
villebro, line 51: Can we add a comment to `queryData` stating that it only contains the first result and is here to preserve backwards compatibility and will be removed in a future version?

villebro (approved):
Minor comment about adding a note about deprecation of `queryData`, other than that LGTM.

PR #847, opened 2020-12-02
build: remove nimbus

ktmud:
🏠 Internal

- Remove dependency on `nimbus`, a tool that nobody maintains anymore.
- Sync build configs with `superset-frontend` so to have more consistency between `superset-ui` and `superset-frontend`. This paves the path for monorepo and could potentially reduce the likelihood of broken `npm link` builds.
- Run prettier and fixes lint errors


suddjian (approved):
Awesome, thank you! Does contributing documentation need to change as a result of this?

ktmud:
> Awesome, thank you! Does contributing documentation need to change as a result of this?

All yarn commands should still work, but the generator may need to add `tsconfig.json` to the template as it is now tracked manually and there is no way to automatically generate it (previous you can do `yarn type:nimbus`, which has been removed).

plugins/legacy-plugin-chart-heatmap/src/Heatmap.js
```diff
@@ -340,18 +340,19 @@ function Heatmap(element, props) {
...
       }
-      image.data[++p] = c.r;
-      image.data[++p] = c.g;
-      image.data[++p] = c.b;
-      image.data[++p] = alpha;
+      image.data[p + 0] = c.r;
+      image.data[p + 1] = c.g;
+      image.data[p + 2] = c.b;
+      image.data[p + 3] = alpha;
+      p += 4;
```
ktmud, line 355: Updated to clean up `++` (disabled by superset-fronend, too).

Issue #799, opened 2020-09-28
Area chart has inconsistent "stacked_style" behavior when using control panel vs. radio buttons


kristw:
#### A. via the control panel
![Pasted_Image_9_28_20__3_57_PM](https://user-images.githubusercontent.com/1659771/94494161-59a1b900-01a3-11eb-998b-9f9966484b98.png)

#### B. if "extra controls" is checked in the control panel, a set of radio buttons will appear on the chart to toggle between the 3 modes:

![Pasted_Image_9_28_20__3_57_PM](https://user-images.githubusercontent.com/1659771/94494175-62928a80-01a3-11eb-955a-bc494601eba7.png)
![Pasted_Image_9_28_20__3_58_PM](https://user-images.githubusercontent.com/1659771/94494221-7d64ff00-01a3-11eb-9007-8362f144e2f3.png)

## Problem

The `stacked_style` in the control panel and the radio buttons are not synced both ways. 
* Changing the dropdown in the control panel updates the radio buttons.
* But changing the radio buttons does not change the control panel. It only changes the area chart appearance and causes the tooltip to show incorrect values. 

### **To Reproduce**
1. Create an area chart with at least two series.
2. Go to "customize". Check the "extra controls" in the control panel
3. Change "stacked style" in the control panel to "stream". 
4. Click on the radio buttons to select "Expanded"

Notice that the dropdown in control panel and the radio buttons are showing different "stacked style". 
Also, the tooltip shows an extra column with incorrect information.

![Pasted_Image_9_28_20__4_08_PM](https://user-images.githubusercontent.com/1659771/94494834-ffa1f300-01a4-11eb-9016-328b3eac1e44.png)

5. Now change "stacked style" in the control panel to "expand". This is the **expected behavior**

![Pasted_Image_9_28_20__4_10_PM](https://user-images.githubusercontent.com/1659771/94494908-324beb80-01a5-11eb-9f51-6a64d9ff738c.png)

## **Environment (please complete the following information):**
 - superset-ui version: master

junlincc:
@kristw yup this is still an issue and our new Echarts time series area chart is not reaching feature parity yet.
quick fix i can think about is to remove the radio button in 
`EXTRA CONTROLS`, and only use `STACKED STYLE` to change chart style. 

[...]

PR #850, opened 2020-12-03
fix(legacy-preset-chart-nvd3): time compare and stacked area tooltips

### LINKED ISSUE #795: Timeseries percent change chart does not show correct date in tooltip
**Describe the bug**

> On the "timeseries percent change" superset visualization, the date in the tooltip shows up as `0Nan`

**To Reproduce**

image

**Expected behavior**
`0Nan` should be replaced with date. 

**Screenshots**
Pasted_Image_9_22_20__3_57_PM

kristw:
🐛 Bug Fix

Fix #795 and #799 

# Fix 795

For the `compare` chart in Superset or the `cumulativeLineChart` (from nvd3), the time is passed to the tooltip as a formatted time string. There is no way to customize the format unlike other chart types. Trying to apply a time format function on a string resulting in `0NaN`. 

This fix will at least make the tooltip has same format with the x-axis, better than `0NaN`.

Pasted_Image_12_2_20__6_01_PM

Also add storybook example for verification

# Fix 799

The stacked area chart has 3 stacked style, the `expand` style pass data to tooltip slightly differently from the other two. It does not include a total row. This PR fix the total number calculation for `expand` case.

Pasted_Image_12_2_20__6_00_PM

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/src/utils.js
+++ b/plugins/legacy-preset-chart-nvd3/src/utils.js
@@ -106,12 +106,37 @@ function getFormattedKey(seriesKey, shouldDompurify) {
 // Custom sorted tooltip
 // use a verbose formatter for times
 export function generateRichLineTooltipContent(d, timeFormatter, valueFormatter) {
-  const time = timeFormatter(d.value);
-
   let tooltip = '';
   tooltip +=
     "<table><thead><tr><td colspan='3'>" +
-    `<strong class='x-value'>${time === '0NaN' ? d.value : time}</strong>` +
+    `<strong class='x-value'>${timeFormatter(d.value)}</strong>` +
+    '</td></tr></thead><tbody>';
+  d.series.sort((a, b) => (a.value >= b.value ? -1 : 1));
+  d.series.forEach(series => {
+    const key = getFormattedKey(series.key, true);
+    tooltip +=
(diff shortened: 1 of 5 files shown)


plugins/legacy-preset-chart-nvd3/src/utils.js
```diff
@@ -132,8 +132,40 @@ export function generateRichLineTooltipContent(d, timeFormatter, valueFormatter)
   return dompurify.sanitize(tooltip);
 }
 
-export function generateAreaChartTooltipContent(d, timeFormatter, valueFormatter) {
-  const total = d.series[d.series.length - 1].value;
+export function generateCompareTooltipContent(d, valueFormatter) {
+  let tooltip = '';
+  tooltip +=
+    "<table><thead><tr><td colspan='3'>" +
+    `<strong class='x-value'>${d.value}</strong>` +
```
ktmud, line 139: Template literals support multiline strings directly so the manual concatenation might not be needed.

ktmud (approved):
lgtm with one style nit

plugins/legacy-preset-chart-nvd3/src/utils.js
kristw: The other tooltip generator use this style so I am gonna leave it same for now. 
Don't want to change all of them yet. 
Also to avoid potential unintentional spacing from space characters between tags.

PR #855, opened 2020-12-09
fix(legacy-preset-chart-nvd3): make deep copy of queryData

kgabryje:
🐛 Bug Fix
Fixes https://github.com/apache/incubator-superset/issues/11889. State of `LineMulti` was passed through props to another component and modified there, which caused an infinite loop if re-renders. Now we pass a deep copy of state instead of the original object.

Before: see gif in https://github.com/apache/incubator-superset/issues/11889
After: the chart is stable after rendering
image

@rusackas Can you please take a look?


junlincc:
Screen Shot 2020-12-09 at 3 29 09 PM

this is finally fixed! tears are coming down 😭 thank you so much Kamil!!

rusackas (approved):
LGTM!

junlincc (approved):

PR #856, opened 2020-12-09
chore: add extra field for Datasource control config

graceguo-supercat:
🏆 Enhancements

We want to display health check message for a datasource: [pending PR](https://github.com/apache/incubator-superset/pull/11970)
This PR is to add extra field in datasource control config.

@ktmud

### DIFF
--- a/packages/superset-ui-chart-controls/src/shared-controls/index.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/index.tsx
@@ -168,9 +168,10 @@ const datasourceControl: SharedControlConfig<'DatasourceControl'> = {
   label: t('Datasource'),
   default: null,
   description: null,
-  mapStateToProps: ({ datasource }) => {
+  mapStateToProps: ({ datasource, dataset_health_check_message }) => {
     return {
       datasource,
+      healthCheckMessage: dataset_health_check_message,
     };
   },
 };
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -78,6 +78,7 @@ export interface DatasourceMeta {
 export interface ControlPanelState {
   form_data: QueryFormData;
   datasource: DatasourceMeta | null;
+  dataset_health_check_message?: string | null;
   controls: ControlStateMapping;
 }
 


ktmud:
This feels more like a state tied to a specific `datasource` other than a global prop. I'd imagine it being set in `mapStateToProps` other than component initialization, so it can be updated when the datasource changes.

graceguo-supercat:
> This feels more like a state tied to a specific `datasource` other than a global prop. I'd imagine it being set in `mapStateToProps` other than component initialization, so it can be updated when the datasource changes.

Good idea!! i will follow this direction and add fix.

packages/superset-ui-chart-controls/src/shared-controls/index.tsx
```diff
@@ -168,9 +168,10 @@ const datasourceControl: SharedControlConfig<'DatasourceControl'> = {
   label: t('Datasource'),
   default: null,
   description: null,
-  mapStateToProps: ({ datasource }) => {
+  mapStateToProps: ({ datasource, dataset_health_check_message }) => {
     return {
       datasource,
+      healthCheckMessage: dataset_health_check_message,
```
ktmud, line 174: Should this be an attribute of `datasource`? I'd imagine the same check may become available to the dataset API endpoint, too.

graceguo-supercat: No. i feel `datasource` contains data that stored in database, while health check message is sanity check on-the-fly, and won't this message won't persist in database.

ktmud: It's a state associated with a specific datasource. It could be considered an extended read-only prop from the API's standpoint---similar to the various transformation [we already did](https://github.com/apache/incubator-superset/blob/7ae8cd07cc2942ae99fcf8ae95aba9ffd65682b9/superset%2Fconnectors%2Fbase%2Fmodels.py#L221-L236) for `order_by` and `verbose_map` in `datasource.data`.

The check may even become an instance method for the `BaseDatasource` model so it's easier to access.

Imagine if we need to access the same info for multi datasources (e.g. display the warning icon in dataset CRUD list view), having the message attached to the datasource would make rendering much easier.

graceguo-supercat: so health_check_message could be part of datasource.data but is not stored in database? 
Yes that will be good because we can access it whenever we call datasource.data. But I do have a little  concern about if this health check become heavy, so I think we should keep the sanity check away from data model, and only trigger it when needed (instead of keep it in the core data bundle).

ktmud: Yes, which is why it doesn't have to be inside the [datasource.data](https://github.com/apache/incubator-superset/blob/7ae8cd07cc2942ae99fcf8ae95aba9ffd65682b9/superset%2Fconnectors%2Fbase%2Fmodels.py#L217) property but could be an instance method and added to the API output in the view handlers. We can also change `datasource.data` to be a method instead of property so each call can choose whether to run these checks or not.

graceguo-supercat: I do not think dataset health sanity check should be instance property function. It works more like SecurityManager, or complicated math utils, which run against some fields of dataset, but nothing related to the Dataset object (and its instance).

I think we should to keep sanity check logic outside Dataset class.

ktmud: Sure. Those are valid arguments. Following the DAO model as laid out in [SIP-35](https://github.com/apache/incubator-superset/issues/9077), maybe the best place for the check runs is in `superset/datasets/commands/heath_check.py`.  However, considering currently the check is still very lightweight and bulk of the code is in the config file, I'm not too concerned about mixing the one LOC that calls a function in the global config with the Datasource class as an instance method.

Either way, at the API payload level though, I still think it doesn't hurt to attach this message to a datasource object.

ktmud: FWIW, metrics has a `warning_text` field built in, too: https://github.com/apache/incubator-superset/pull/3499

This warning text seems serve a similar purpose to the datasource warning message, at least from what users can see. The only difference is one has to be added manually, the other can run automatically.

Since the datasource doesn't change that much, maybe instead of running the health checks every time on the fly, you can just run the checks using db insert/update hooks and write the warning text to db.

If we go with this route, you won't even need to create a new column and run db migrations, because you can just use the existing `json_metadata` field.

graceguo-supercat: I am not sure which table has `json_metadata` field that could be used to store datasource health check results. But I feel this approach has some many disadvantage:

- Unnecessary db migrations.
If you store datasource sanity check result into database, whenever you add/remove/update a new rule, you need to run db migration for the existed datasource right? For virtual datasource, many datasources are not being re-used by charts anymore, these db migration cause too much overhead. 
- Each company implements their own rules.
The most important goal for sanity check is to allow each company implements their own rules. Because of the versioning of db migration,  every db migration will affect everyone in the whole open source community. You can ask Airbnb add a feature flag so that Dropbox won't be affected, then ask Dropbox and others add another flag so that Airbnb won't affected...If things go this troublesome way probably means this solution is not good enough.
- Flexible and quick feedback
If we can add datasource health check on-the-fly, it's easier deploy, and don't have to do cleanup if we disabled the rule later. 

For those important datasources, airbnb already had data pipeline to label them, like "verified" or "deprecated" etc (and updates labels a few times a day). Health check API is a complementary solution for user created datasources. Because these datasources are used quite differently, we should offer some flexible way to provide guideline and regulations.

ktmud: > I am not sure which table has json_metadata field that could be used to store datasource health check results.

You can use the [`extra`](https://github.com/apache/incubator-superset/blob/0396c705d4b3d886618c664aa339178bfb188066/superset%2Fconnectors%2Fsqla%2Fmodels.py#L464) field we already use for table certification. Previous I confused it with `json_metadata` that is only used in annotation datasource.

---

Thanks for the thoughts! I'll admit I didn't put too much thought into this when I proposed the db metadata approach, but here is a potential solution to the problems you raised:

> If you store datasource sanity check result into database, whenever you add/remove/update a new rule, you need to run db migration for the existed datasource right? 

You don't have to if you version the check function. The stored construct can be something like this:

```json
{
  "health_check": { "version": "1", "error": "SELECT_STAR" }
}
```

The check function config can be updated as such:

```python
def check_datasource_health():
   return {
      "error": "...",
      "extra": {
         "num_select": 2
      },
    }

check_datasource_health.version = "1"

DATASOURCE_HEALTH_CHECK = check_datasource_health
```

Then in the health check command, you check the stored version against the check function config. Run or rerun the checks when there is no stored check results or the version mismatch.

This approach would also run the checks on the fly (if needed), so no migrations or offline jobs or db hooks are required anymore. Comparing to your original proposal, the only difference is that the check results are cached. Maybe this is an unnecessary optimization, but at least the concerns about company specific rules and fast iterations shall have been mitigated.

We don't have to do this if it's too much work, but the point I wanted to make is it shall be OK to link the health check to the datasource model and consider it part of the core functionality, but can be configured to run optionally.

graceguo-supercat: I think `check_datasource_health` should be a function that takes Datasource as input, then it runs some simple sql parse, then return message. In your example, i don't understand what is the point of version and how to use `check_datasource_health` to run sql parse?

ktmud: Yes, it can be a simple feature, but with [SIP-35](https://github.com/apache/incubator-superset/issues/9077) in progress, it's important to think about how this fits into the bigger picture.

Version is used to check whether the stored health check results is outdated therefore needs a rerun. It will be used in a method of `SqlaTable` (or a new health check command under `superset/datasets/commands`):

E.g.

```python
class SqlaTable:

    def check_health(self, commit: bool = False) -> Optional[HealthCheckResult]:
        check = config["DATASOURCE_HEALTH_CHECK"]
        if check is None:
            return

        extra = self.get_extra_dict()
        if extra.get("health_check", {}).get("version") == check.version:
            return extra["health_check"]

        result = check(self)
        extra["health_check"] = {"version": check.version, "result": result}
        self.extra = json.dumps(self.get)

        db.session.merge(self)
        if commit:
            db.session.commit()

        return result

```

ktmud: Sorry for the misleading code, but `check_datasource_health` does accept datasource as an input. I've updated my previous comment accordingly.

graceguo-supercat: Thanks for the new example, i understand your idea better. Is persisting to db really necessary? If so, then we need to update:
- when rule changed, or
- when datasource is updated. How to make sure this check_health get called whenever datasource is updated?

graceguo-supercat: I just checked airbnb's airflow job: it will update `extra` periodically, which means if you want to re-use the same field,  airflow jobs and sanity check may step on each other's toe.

ktmud: It's not super necessary, but as you mentioned earlier, the health check _could_ get expensive. In case we want to cache the results somewhere, this would be the way to do it.

When datasource is updated, we can simply add a one liner `datasource.check_health(commit=True)` to the [update](https://github.com/apache/incubator-superset/blob/a0b81357cdaefd8f26f6ee98b8330f526dd3d85a/superset%2Fdatasets%2Fcommands%2Fupdate.py#L64) command.

I thins the Airflow job can be updated pretty easily. In fact, since you don't know what other future features may put things into `extra`, any updates on the `extra` field should be a `merge` operation instead of a simple override.

graceguo-supercat: Though `the health check could get expensive.` but it's not the concern right now. If most of the check is expensive, i will choose to send health check _async_ in the explore view, instead of adding check _during_ the explore request.

graceguo-supercat: We will go direction that _save health_check results_ into datasource `extra` field. So there is no need to change datasource control config. close this PR.

PR #862, opened 2020-12-14
refactor: remove queryFields from formData and improve typing

ktmud:
🏠 Internal

- Reorganize and rename some typing for QueryForm and QueryObject.
- Fix some npm link and build issues.
- Get rid of `convertMetric` and `is_timeseries`, move relevant query config consolidation to backend.
   Ideally, backend should be more robust in handling input than the frontend, because a) it's better for security, b) the API payload can be simpler so we don't have to duplicate the same logic across different clients (imagine a Superset iOS app in the future).
- Remove `queryFields` from `formData` as it should not be stored in database. See https://github.com/apache/incubator-superset/pull/9962#issuecomment-655791527

This requires an accompanying PR in incubator-superset.

The typing optimization is mostly for bringing clarity to metrics and columns.

1. **Metric**: metric definition found in datasource, represent a Metric entity stored in database.
2. **AdhocMetric**: adhoc metric definition used in form data
3. **PredefinedMetric**: a reference to `Metric` used in form data with possible label overrides.
4. **QueryFormMetric**: either an `AdhocMetric` or `PredefinedMetric` or a string representing the metric name of a predefined `Metric`.


packages/superset-ui-chart-controls/src/shared-controls/index.tsx
```diff
@@ -126,7 +126,7 @@ const groupByControl: SharedControlConfig<'SelectControl', ColumnMeta> = {
     if (state.datasource) {
       const options = state.datasource.columns.filter(c => c.groupby);
       if (includeTime) {
-        options.push(timeColumnOption);
+        options.unshift(timeColumnOption);
```
ktmud, line 129: Move time column to the first option for `groupby`. It's hard to notice this is an auto-generated column when it's at the end of the list.

packages/superset-ui-core/src/types/index.ts
```diff
@@ -1,7 +1,21 @@
...
+ * under the License.
+ */
+export * from '../query/types';
 
-export enum OverwritePolicy {
-  ALLOW = 'ALLOW',
-  PROHIBIT = 'PROHIBIT',
-  WARN = 'WARN',
-}
+export type Maybe<T> = T | null;
```
villebro, line 21: 😄  Ok, I can live with this, but it does feel slightly extreme

packages/superset-ui-core/test/query/buildQueryObject.test.ts
```diff
@@ -47,30 +55,34 @@ describe('buildQueryObject', () => {
...
+      {
+        datasource: '5__table',
+        granularity_sqla: 'ds',
+        viz_type: 'table',
+        metrics: ['sum__num'],
+        my_custom_metric_control: 'avg__num',
+      },
+      { my_custom_metric_control: 'metrics' },
+    );
+    expect(query.metrics).toEqual(['sum__num', 'avg__num']);
```
villebro, line 85: I wish GH could show a proper diff on this

villebro (approved):
A few non-blocking comments

packages/superset-ui-core/src/types/index.ts
ktmud: I learned this from a codegen library for GraphQL.. 😄

packages/superset-ui-core/test/query/buildQueryObject.test.ts
ktmud: image

You mean like this?

villebro: Live and learn! 👍

package.json
```diff
@@ -14,7 +14,7 @@
     "build-storybook": "cd packages/superset-ui-demo && yarn build-storybook",
     "chromatic": "cd packages/superset-ui-demo && yarn chromatic",
     "sb": "yarn storybook",
-    "clean": "rm -rf ./{packages,plugins}/**/{lib,esm,tsconfig.tsbuildinfo} build/",
+    "clean": "yarn build --clean --no-babel --no-type",
```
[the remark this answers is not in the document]
ktmud: It's because `clean` doesn't clean up `node_modules` under each package installed by `npm link`. I tried to do some optimization in this PR: https://github.com/apache-superset/superset-ui/pull/862/files#diff-a992dd64a8c626fedc95d613f43602b91f6dc6339893afaf405dc05fa13c9cea

But it may not fully resolve this issue still.

Issue #866, opened 2020-12-15
Is it ok that SupersetClientClass is ignoring DEFAULT_FETCH_RETRY_OPTIONS?


scriminaci:
I was struggling configuring retry options and I found out that SupersetClientClass import the DEFAULT_FETCH_RETRY_OPTIONS but it does not use them in the request method.

If the behavior is not the designed one, I could create a pull request.

Thanks
Mario

ktmud:
https://github.com/apache-superset/superset-ui/blob/a4ecdb3d5a7791e864703318d6da9154bb842762/packages%2Fsuperset-ui-core%2Fsrc%2Fconnection%2FSupersetClientClass.ts#L82

It does use it here, right?

scriminaci:
It sets the fetchRetryOptions attribute but to actually use it.

It should pass it in the request here
https://github.com/apache-superset/superset-ui/blob/a4ecdb3d5a7791e864703318d6da9154bb842762/packages/superset-ui-core/src/connection/SupersetClientClass.ts#L134

The result is that it calls callApiAndParseWithTimeout without passing anything. For example, credentials are passed correctly.

ktmud:
Ah, I see what you mean. Feel free to make a PR.

Thanks for spotting this!

PR #867, opened 2020-12-16
fix: add fetchRetryOptions in SupersetClientClass.request method

scriminaci:
🐛 Bug Fix

Following the thread 
https://github.com/apache-superset/superset-ui/issues/866#issuecomment-745503482
I propose this PR to use the default fetchRetryOptions in the SupersetClientClass.

Thanks
Mario


ktmud (approved):
LGTM. Thank you for the fix!

PR #869, opened 2020-12-16
chore(chart-controls): remove dedicated time section

villebro:
💔 Breaking Changes
This PR paves the way for breaking up the dedicated Time section as defined by SIP-34. As a first step we are removing the `druidTimeSection` and `sqlaTimeSection`, and expect control panels to explicitly specify where any SQL or Druid temporal controls are placed. It is then the responsibility of the frontend to not render any controls that are not applicable. For instance, when editing a chart with a SQL dataset, Druid controls should not be visible and vice versa.

To ensure existing charts don't break, we're introducing two new predefined sections: `legacyRegularTime` (only contains time range and time column) and `legacyTimeseriesTime` which also includes the time grain and Druid time origin. These are implemented in all predefined viz plugins featured in the `plugins` directory.

In addition, some missing types are added that were picked up during this refactoring process and bugs fixed.

### Screenshots
Regular (=non timeseries) chart. No visible changes (see the "Time" section):
image

Timeseries chart. Notice that "Time Column" and "Time Grain" are now on separate rows, otherwise no changes:
image

The only visible change is that time controls are now one per row (traditionally)

### DIFF
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -69,8 +69,8 @@ export interface DatasourceMeta {
   main_dttm_col: string;
   // eg. ['["ds", true]', 'ds [asc]']
   order_by_choices?: [string, string][] | null;
-  time_grain_sqla: [string, string][];
-  granularity_sqla: [string, string][];
+  time_grain_sqla?: string;
+  granularity_sqla?: string;
   datasource_name: string | null;
   description: string | null;
 }
--- a/packages/superset-ui-chart-controls/src/utils/columnChoices.ts
+++ b/packages/superset-ui-chart-controls/src/utils/columnChoices.ts
@@ -21,10 +21,10 @@ import { DatasourceMeta } from '../types';
 /**
  * Convert Dataource columns to column choices
  */
-export default function columnChoices(datasource?: DatasourceMeta | null) {
+export default function columnChoices(datasource?: DatasourceMeta | null): [string, string][] {
   return (
     datasource?.columns
-      .map(col => [col.column_name, col.verbose_name || col.column_name])
+      .map((col): [string, string] => [col.column_name, col.verbose_name || col.column_name])
       .sort((opt1, opt2) => (opt1[1].toLowerCase() > opt2[1].toLowerCase() ? 1 : -1)) || []
   );
 }
--- a/packages/superset-ui-chart-controls/test/utils/columnChoices.test.tsx
+++ b/packages/superset-ui-chart-controls/test/utils/columnChoices.test.tsx
@@ -23,6 +23,7 @@ describe('columnChoices()', () => {
   it('should convert columns to choices', () => {
     expect(
       columnChoices({
+        id: 1,
         metrics: [],
         type: DatasourceType.Table,
         main_dttm_col: 'test',
@@ -40,6 +41,10 @@ describe('columnChoices()', () => {
             verbose_name: 'bar',
           },
         ],
+        verbose_map: {},
+        column_format: { fiz: 'NUMERIC', about: 'STRING', foo: 'DATE' },
+        datasource_name: 'my_datasource',
+        description: 'this is my datasource',
       }),
     ).toEqual([
       ['foo', 'bar'],
--- a/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
+++ b/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
@@ -17,12 +17,13 @@
  * under the License.
  */
 import { t } from '@superset-ui/core';
-import { formatSelectOptions, ControlPanelConfig } from '@superset-ui/chart-controls';
+import { ControlPanelConfig, formatSelectOptions, sections } from '@superset-ui/chart-controls';
 import React from 'react';
 import { headerFontSize, subheaderFontSize } from '../sharedControls';
 
 const config: ControlPanelConfig = {
   controlPanelSections: [
+    sections.legacyTimeseriesTime,
     {
       label: t('Query'),
       expanded: true,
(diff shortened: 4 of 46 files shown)


packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -69,8 +69,8 @@ export interface DatasourceMeta {
   main_dttm_col: string;
   // eg. ['["ds", true]', 'ds [asc]']
   order_by_choices?: [string, string][] | null;
-  time_grain_sqla: [string, string][];
-  granularity_sqla: [string, string][];
+  time_grain_sqla?: string;
+  granularity_sqla?: string;
```
villebro, line 73: Bycatch, these types were just plain wrong (prob copied from the line above).

packages/superset-ui-chart-controls/src/utils/columnChoices.ts
```diff
@@ -21,10 +21,10 @@ import { DatasourceMeta } from '../types';
 /**
  * Convert Dataource columns to column choices
  */
-export default function columnChoices(datasource?: DatasourceMeta | null) {
+export default function columnChoices(datasource?: DatasourceMeta | null): [string, string][] {
   return (
     datasource?.columns
-      .map(col => [col.column_name, col.verbose_name || col.column_name])
+      .map((col): [string, string] => [col.column_name, col.verbose_name || col.column_name])
```
villebro, line 27: Bycatch: missing return type

packages/superset-ui-chart-controls/test/utils/columnChoices.test.tsx
```diff
@@ -40,6 +41,10 @@ describe('columnChoices()', () => {
             verbose_name: 'bar',
           },
         ],
+        verbose_map: {},
+        column_format: { fiz: 'NUMERIC', about: 'STRING', foo: 'DATE' },
+        datasource_name: 'my_datasource',
+        description: 'this is my datasource',
```
villebro, line 47: Bycatch: not sure why linter wasn't picking this up, but the object was missing required props

ktmud:
image

Do you think we should align the run buttons with the rest of the content in the second column?

plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
```diff
@@ -41,18 +48,23 @@ const config: ControlPanelConfig = {
               label: t('Comparison Period Lag'),
               isInt: true,
               description: t('Based on granularity, number of time periods to compare against'),
+              renderTrigger: true,
```
villebro, line 51: no reason why this can't trigger instantly

plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
```diff
@@ -41,18 +48,23 @@ const config: ControlPanelConfig = {
...
           },
+        ],
+        [
           {
             name: 'compare_suffix',
             config: {
               type: 'TextControl',
               label: t('Comparison suffix'),
               description: t('Suffix to apply after the percentage display'),
+              renderTrigger: true,
```
villebro, line 62: same here

plugins/legacy-preset-chart-big-number/src/BigNumberTotal/controlPanel.ts
```diff
@@ -37,6 +44,7 @@ export default {
               type: 'TextControl',
               label: t('Subheader'),
               description: t('Description text that shows up below your Big Number'),
+              renderTrigger: true,
```
villebro, line 47: same here

villebro:
> image
> 
> Do you think we should align the run buttons with the rest of the content in the second column?

I might be wrong, but I believe the Run/Save button are defined in `superset-frontend`. Check this PR which is touching up the aesthetics on Explore view: https://github.com/apache/incubator-superset/pull/12008 (oh, I just noticed that you'd posted there too, but will leave this link for reference)

FYI I'm going to break this PR in two parts to make it easier to review - first part makes the breaking changes in `chart-controls` necessary for completing the migration, and updates all plugins to use the new constructs without making any visual changes. The next PR (and subsequent PRs) will then fix the layouts.

plugins/legacy-preset-chart-nvd3/src/Bubble/controlPanel.ts
```diff
@@ -31,13 +36,14 @@ import {
   leftMargin,
 } from '../NVD3Controls';
 
-export default {
-  label: t('Bubble Chart'),
```
villebro, line 35: Ha! Extra typing identified an incorrect property!

ktmud:
Yeah, I pasted the image in the wrong PR while juggling through different browser windows. You guys are so productive nowadays, I could not keep up!

ktmud (approved):
Not sure how does this look like in the real world, but I trust your judgement, so giving my stamp!

plugins/legacy-preset-chart-nvd3/src/Compare/controlPanel.ts
```diff
@@ -61,12 +61,6 @@ export default {
     timeSeriesSection[1],
     sections.annotations,
   ],
-  sectionOverrides: {
```
rusackas, line 64: SO stoked to see these overrides going away!!!!

rusackas (approved):
LGTM! Thanks for all the cleanup!

pkdotson (approved):
LGTM!

plugins/legacy-preset-chart-nvd3/src/Bubble/controlPanel.ts
ktmud: Nice

suddjian (approved):
gonna pile on here

villebro:
> Not sure how does this look like in the real world, but I trust your judgement, so giving my stamp!

Sorry, forgot to add screenshots; description updated!

PR #871, opened 2020-12-17
feat: introduce shared modules logic

suddjian:
💔 Breaking Changes

🏆 Enhancements

This introduces "shared modules" which are used for [dynamic plugins](https://github.com/apache/incubator-superset/pull/10288). It's exported from here so that it can be called in the future by third party plugins.

📜 Documentation

🐛 Bug Fix

🏠 Internal


suddjian:
gonna write tests for this before opening it

rusackas (approved):
This LGTM! Always worth a little extra scrutiny when `window` comes out to play, but I think this is sensible and practical and (as far as I can tell) safe enough. The worst case is that there's some crazy malicious module being registered... but that has to be _loaded_ by the user, and _called_ by the user. No more dangerous than loading a package from NPM, or using your own JS console, in my opinion. In a future PR/effort, we can load plugins into iframes, _just in case_. The tests are awesome, and give me confidence in the feature!

PR #873, opened 2020-12-20
feat: remove single queryData from api

simchaNielsen:
💔 Breaking Changes
Removes from Charts api deprecated `queryData` param and replace it with `queriesData`

🏆 Enhancements

📜 Documentation
`queriesData` - was implemented in previous PRs and supported by `incubator-superset`, here we just remove deprecated property from api

🐛 Bug Fix

🏠 Internal


villebro:
@simchaNielsen I assume you will need to update all plugins to reference the new `queriesData[0]` instead of the current `queryData` in this PR

PR #874, opened 2020-12-20
chore(core): remove queryData and replace with queriesData

simchaNielsen:
💔 Breaking Changes

Migrate old chart plugins to new property `queriesData` from `queryData`
It's addition to: https://github.com/apache-superset/superset-ui/pull/846

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal


amitmiran137:
@simchaNielsen I think you need to change only non-legacy plugins.
only they actually have access to multiple queries support
while legacy remains the same
@villebro  WDYT?

packages/superset-ui-core/src/chart/models/ChartProps.ts
```diff
@@ -47,9 +47,6 @@ export interface ChartPropsConfig {
   height?: number;
   /** Programmatic overrides such as event handlers, renderers */
   hooks?: Hooks;
-  /** Formerly called "payload". This property going to be deprecated because
-   * contains only first item in response data array (use `queriesData` instead) */
-  queryData?: QueryData;
```
villebro, line 52: let's remove the comment below or say something like "the data returned for all `queries` objects in the request" or similar (feel free to reword).

packages/superset-ui-demo/storybook/shared/components/createQueryStory.tsx
```diff
@@ -48,11 +48,11 @@ export default function createQueryStory({
...
                         // @TODO fix typing
                         // all vis's now expect objects but api/v1/ returns an array
-                        queryData={
+                        queriesData={[
                           Array.isArray(payload.queryData)
                             ? payload.queryData[0]
-                            : payload.queryData
-                        }
+                            : payload.queryData,
+                        ]}
```
villebro, line 55: Should we instead do `queriesData={Array.isArray(payload.queryData) ? payload.queryData : [payload.queryData]}`? Or just change `payload.queryData` to `payload.queriesData` and make sure it's always an array.

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-table/TableStories.tsx
```diff
@@ -40,18 +40,20 @@ function loadData(
   props: TableChartProps,
   { pageLength = 50, rows = 1042, cols = 8, alignPn = false, showCellBars = true },
 ): TableChartProps {
-  if (!props.queryData) return props;
-  const records = props.queryData?.data?.records || [];
-  const columns = props.queryData?.data?.columns || [];
+  if (!props.queriesData || !props.queriesData[0]) return props;
+  const records = props.queriesData?.[0].data?.records || [];
+  const columns = props.queriesData?.[0].data?.columns || [];
```
villebro, line 45: I wonder if we need to check length here?
```
> const emptyArray = []
> emptyArray ? 'truthy' : 'falsy'
'truthy'
```

villebro (commented):
A few first pass comments. Also, did you check `superset-ui-generator`? We might need to check that it still works after all the breaking changes, but it'd be nice to get these changes in there, too.

packages/superset-ui-demo/storybook/shared/components/createQueryStory.tsx
simchaNielsen: This comes from `ChartDataProvider`, I didn't change there `queryData` to `queriesData` because didn't find who use it, so there it still should be queryData

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-table/TableStories.tsx
simchaNielsen: amm... can you clarify please?

packages/superset-ui-core/src/chart/models/ChartProps.ts
simchaNielsen: updated

packages/superset-ui-demo/storybook/shared/components/createQueryStory.tsx
villebro: We should probably change this one, too: https://github.com/apache-superset/superset-ui/blob/master/packages/superset-ui-core/src/chart/components/ChartDataProvider.tsx

packages/superset-ui-demo/storybook/stories/plugins/plugin-chart-table/TableStories.tsx
villebro: Sorry, disregard, I read this too quickly and didn't notice the added `||` here.

packages/superset-ui-demo/storybook/shared/components/createQueryStory.tsx
simchaNielsen: done

villebro (approved):

PR #879, opened 2020-12-22
fix(legacy-preset-chart-nvd3): add Time section to Pie control panel

villebro:
🐛 Bug Fix
Fix yet another regression from #869 . This should be the last one, and I did my best to check all control panel implementations to make sure they are correctly referencing either `legacyRegularTime` or `legacyTimeseriesTime`.


villebro:
Closing as this only affects a deprecated plugin (NVD3 Pie)

PR #889, opened 2021-01-07
refactor(plugin-chart-table): migrate to API v1

ktmud:
🏆 Enhancements

Migrate the table viz to use API v1, along with some refactoring related to QueryContext.

1. Add typing for `post_processing`
2. Allow `ChartPlugin` to have custom `ChartProps`
3. Remove `groupby` from `queryFields` as it is [deprecated](https://github.com/apache/superset/pull/9366).
4. Add `GenericDataType`, the chart data type hints for rendering

This PR is a prerequisite of https://github.com/apache/superset/pull/10270


ktmud:
Blocked by https://github.com/apache-superset/superset-ui/pull/892 because ESLint needs an upgrade in order to fix frivolous lint errors.

packages/superset-ui-core/src/query/extractQueryFields.ts
```diff
@@ -46,26 +45,45 @@ export default function extractQueryFields(
...
+
+    let normalizedKey: string = queryFieldAliases[key] || key;
+
+    // ignore groupby when in raw records mode
+    if (queryMode === QueryMode.raw && normalizedKey === 'groupby') {
+      return;
+    }
+    if (queryMode === QueryMode.aggregate) {
+      // ignore columns when (specifically) in aggregate mode.
+      if (normalizedKey === 'columns') {
```
simchaNielsen, line 66: can we unit these 2 conditions? like:
```
if (queryMode === QueryMode.aggregate && normalizedKey === 'columns) {
```

packages/superset-ui-core/src/query/types/PostProcessing.ts
```diff
@@ -0,0 +1,118 @@
+import { JsonObject } from '../../connection';
+import { TimeGranularity } from '../../time-format';
+
+/**
```
simchaNielsen, line 4: it should be before all imports

packages/superset-ui-core/src/query/types/PostProcessing.ts
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * the name of the column to generate aggrates from.
+     */
+    column?: string;
+    options?: JsonObject;
+  };
+}
+
+export interface PostProcessingAggregation {
+  operation: 'aggregation';
```
simchaNielsen, line 62: should it be some TS type?

packages/superset-ui-core/src/utils/removeDuplicates.ts
```diff
@@ -0,0 +1,35 @@
...
+  const result: T[] = [];
+  items.forEach(x => {
+    const itemHash = hash ? hash(x) : x;
+    if (!seen.has(itemHash)) {
+      seen.add(itemHash);
+      result.push(x);
+    }
+  });
+  return result;
+}
```
simchaNielsen, line 33: may be only `Set` enougph? if I got correctly logic do it like (but not sure):
```
items.forEach(x => {
    if (!hash || !hash(x));
      seen.add(x);
    }
  });
return [...seen]
```

packages/superset-ui-core/src/query/extractQueryFields.ts
ktmud: Sure. There was more conditions in this branch before but I must have cleaned them up.

packages/superset-ui-core/src/query/types/PostProcessing.ts
ktmud: Right. Thanks for catching this!  The imported were added by VS Code autocomplete and I forgot to check.

packages/superset-ui-core/src/query/types/PostProcessing.ts
ktmud: I think it's find to use string literals here, otherwise we'd need to create a type for each operation, which doesn't seem to add much additional value.

packages/superset-ui-core/src/utils/removeDuplicates.ts
ktmud: Does set spread preserve orders?

ktmud: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#iteration_methods) says yes, but this still wouldn't work for my needs because I need to check uniqueness by hash, not the item itself.

kristw: ideas: can short circuit when `hash` is not used and only check `hash` once.

```ts
function removeDuplicates<T>(items: T[], hash?: (item: T) => unknown): T[] {
  if (hash) {
    const seen = new Set();
    return items.filter(x => {
      const itemHash = hash(x);
      if (seen.has(itemHash)) return false; 
      seen.add(itemHash);
      return true;
    });
  }

  return [...new Set(items)];
}
```

plugins/plugin-chart-table/test/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "composite": false,
     "emitDeclarationOnly": false,
     "noEmit": true,
-    "rootDir": "."
+    "rootDir": "../../../"
```
kristw, line 6: Is this intentional?

packages/superset-ui-core/src/chart/models/ChartPlugin.ts
```diff
@@ -95,26 +102,24 @@ export default class ChartPlugin<T extends QueryFormData = QueryFormData> extend
   }
 
   register() {
-    const { key = isRequired('config.key') } = this.config;
-    getChartMetadataRegistry().registerValue(key as string, this.metadata);
-    getChartComponentRegistry().registerLoader(key as string, this.loadChart);
-    getChartControlPanelRegistry().registerValue(key as string, this.controlPanel);
-    getChartTransformPropsRegistry().registerLoader(key as string, this.loadTransformProps);
+    const key: string = this.config.key || isRequired('config.key');
```
kristw, line 105: Given we are all in with typescript now. Could also remove the legacy `isRequired` trick that was before the age of typescript. (No need to do in this pr).

kristw (approved):
Just a few things here and there. Overall lgtm

packages/superset-ui-core/src/query/extractQueryFields.ts
```diff
@@ -46,26 +45,46 @@ export default function extractQueryFields(
...
+    }
+    // ignore columns when (specifically) in aggregate mode.
+    if (queryMode === QueryMode.aggregate && normalizedKey === 'columns') {
+      return;
+    }
+    // groupby has been deprecated: https://github.com/apache/superset/pull/9366
+    if (normalizedKey === 'groupby') {
+      normalizedKey = 'columns';
+    }
+    if (normalizedKey === 'metrics') {
```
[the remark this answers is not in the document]
ktmud: I try to avoid `switch` nowadays since Python has taught me [`switch` is just a syntactic sugar for if else's](https://stackoverflow.com/questions/46701063/why-doesnt-python-have-switch-case). Plus I hate it when I forgot to call `break`.

plugins/plugin-chart-table/test/tsconfig.json
ktmud: Yes as I need to import a contextualized `enzyme` mount from root: https://github.com/ktmud/superset-ui/blob/8e744a77751fc8863522c8938974262e7e3463bd/plugins/plugin-chart-table/test/TableChart.test.tsx#L21

packages/superset-ui-core/src/chart/models/ChartPlugin.ts
ktmud: Noted. Let's do this when all major plugins (at their index.js file) are migrated to typescript.

villebro (approved):
This is amazing stuff, thanks so much for all the improvements here. Let's get this in and deal with any potential issues with higi priority, as I feel the improvements here far outweigh any potential regressions caused by it.

packages/superset-ui-core/src/query/types/QueryFormData.ts
```diff
@@ -149,7 +154,7 @@ export interface DruidFormData extends BaseFormData {
   druid_time_origin?: string;
 }
 
-export type QueryFormData = SqlaFormData | DruidFormData;
+export type QueryFormData = DruidFormData | SqlaFormData;
```
ktmud, line 157: Switching order so if something doesn't add up, it will fallback to `SqlaFormData`

packages/superset-ui-core/src/utils/removeDuplicates.ts
ktmud: Good idea! Updated.

packages/superset-ui-core/test/utils/removeDuplicates.test.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('removeDuplicates([...])', () => {
+  it('should remove duplciages from a simple list', () => {
+    expect(removeDuplicates([1, 2, 4, 1, 1, 5, 2])).toEqual([1, 2, 4, 5]);
+  });
+  it('should remove duplciages by key getter', () => {
+    expect(removeDuplicates([{ a: 1 }, { a: 1 }, { b: 2 }], x => x.a)).toEqual([
+      { a: 1 },
+      { b: 2 },
+    ]);
+  });
```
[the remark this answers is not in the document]
ktmud: Updated!

plugins/plugin-chart-table/src/transformProps.ts
```diff
@@ -88,86 +67,73 @@ const processDataRecords = memoizeOne(function processDataRecords(
...
     datasource: { columnFormats, verboseMap },
-    formData: {
-      tableTimestampFormat,
-      timeGrainSqla: granularity,
+    rawFormData: {
+      table_timestamp_format: tableTimestampFormat,
+      time_grain_sqla: granularity,
       metrics: metrics_,
-      percentMetrics: percentMetrics_,
+      percent_metrics: percentMetrics_,
```
ktmud, line 77: I'm just sick and tire of worrying about casing so I choose to use raw form data here. Should probably gradually deprecating this auto-case-transformation for the final charting API as well...

plugins/plugin-chart-table/src/types.ts
```diff
@@ -25,52 +25,49 @@ import {
...
+  records: DataRecord[];
   columns: string[];
 }
 
-export interface TableChartFormData {
-  alignPn?: boolean;
-  colorPn?: boolean;
-  includeSearch?: boolean;
-  pageLength?: string | number | null; // null means auto-paginate
+export type TableChartFormData = QueryFormData & {
```
ktmud, line 50: The benefit of not doing case transformation is that we can directly reuse this typing for both `formData` sent to the API and received by the chart renderer (e.g. in both `buildQuery` and `transformProps`.)

ktmud (commented):
Adding more comments and replies

ktmud:
> This is amazing stuff, thanks so much for all the improvements here. Let's get this in and deal with any potential issues with higi priority, as I feel the improvements here far outweigh any potential regressions caused by it.

Thank you, @villebro ! I've tested ECharts, NVD3, Box plot, Word Cloud and tried to be as comprehensive as I can. Also wanted to add more Cypress tests for the missing chart types (boxplot, time series table, for example) before merging, but didn't have enough time since a lot of my time has been spent on [updating Cypress itself](https://github.com/apache/superset/pull/12605)..

It would be super nice if someone from Preset could prioritize adding E2E tests for viz types or completing existing ones with more use cases so we can have more confidence in doing bigger viz plugin improvements like this. 

cc @junlincc

ktmud:
I'm going to a new version and update https://github.com/apache/superset/pull/10270

PR #896, opened 2021-01-12
chore: Update big number control options

pkdotson:
💔 Breaking Changes

🏆 Enhancements
update big number control options.
Screen Shot 2021-01-12 at 12 05 50 AM

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
+++ b/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
@@ -44,6 +44,8 @@ const config: ControlPanelConfig = {
               description: t('Based on granularity, number of time periods to compare against'),
             },
           },
+        ],
+        [
           {
             name: 'compare_suffix',
             config: {


ktmud:
Should we put the two checkboxes in separate lines as well?

ktmud:
I'm also wondering instead of putting every control in its separate row, how many of these line break issues can be fixed by setting a smaller font-size for the control label.

cc @junlincc

rusackas:
> I'm also wondering instead of putting every control in its separate row, how many of these line break issues can be fixed by setting a smaller font-size for the control label.
> 
> cc @junlincc

I think that would get us out of _some_ trouble, but it s a slippery slope. It won't likely solve all of them unless we get into teensy font territory, which would cause accessibility issues.

I think that putting each control on its own row is a good practice, and is _generally_ in the target direction of the new design.

If it does go down that road, and controls each get their own row, I think we should reconsider the `controlSetRows` layout array-of-arrays model, and consider heading toward a more responsive layout solution with CSS. But that's a bigger project!

rusackas (approved):
LGTM, fixes the acute issue, looks much better!

rusackas:
@junlincc I think we can merge as is, but lemme know if you wanna hold off and handle those checkboxes. There are probably a LOT of these checkbox issues, if we want to make a sweep of them in another PR.

ktmud:
@rusackas @junlincc do we want to update the FilterBox controls as well?

ktmud:
May I also suggest changing the checkbox label text to sentence casing....  All caps long-text is very hard to read..

### Before



### After




cc @mihir174

PR #901, opened 2021-01-15
fix(plugin-chart-echarts): remove columns from formData

villebro:
🐛 Bug Fix
Make sure `columns` property isn't defined in `formData`, as it is a reserved parameter with a specific meaning in `QueryObject`.


rusackas (approved):
Oh man... LGTM! Let me know if you need help carrying it across to Superset.

ktmud:
How about this change: https://github.com/apache-superset/superset-ui/pull/889/files#diff-68ae7687936cc7b2de6790b6c6abb9b48a150dc9977b665b3717a7e3f4a90632 ?

ktmud (approved):

ktmud:
This change is good, but it may need to be updated soon after I deprecate `groupby` in #889

PR #909, opened 2021-01-22
fix: remove max-width from tooltip div

bryanck:
🐛 Bug Fix

This fixes an issue where the rich tooltip text will overflow outside of the tooltip box in Superset.

Before:
105507084-a09a4980-5c7f-11eb-95fe-9eda8b2d56b3

After:
Screen Shot 2021-01-22 at 1 05 29 PM

This fixes Superset issue 12688 https://github.com/apache/superset/issues/12688

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/plugins/legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -1153,7 +1153,7 @@ function nvd3Vis(element, props) {
       }
     }
 
-    wrapTooltip(chart, maxWidth);
+    wrapTooltip(chart);
 
     return chart;
   };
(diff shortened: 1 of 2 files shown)


junlincc:
Thank you @bryanck for the PR and recent contribution!  feel free to reach out to me on slack if you are looking for making bigger impact on Superset. 🙏

suddjian (commented):
Thanks for the fix. I gotta ask: Do we know why this maxWidth parameter was there in the first place?

villebro:
> Thanks for the fix. I gotta ask: Do we know why this maxWidth parameter was there in the first place?

I wonder if it's there to control for ridiculously long labels breaking the tooltip?

bryanck:
Well setting the `max-width` doesn't truncate the label, it just limits the size of the box (as you can see in the screenshot), so a very long label will still be displayed.

ktmud:
Or is it for preventing wider tooltips being cutoff in Dashboard's slice containers?

ktmud (approved):
Either way, seems like a safe enough change to me.

PR #913, opened 2021-01-25
chore: move all controls for NVD3 and Big Number charts to separate rows

zhaoyongjie:
🏆 Enhancements:
Make control panel to a single column

The associated issue: https://github.com/apache/superset/issues/12719

image


ktmud (approved):

plugins/legacy-preset-chart-nvd3/src/TimePivot/controlPanel.ts
```diff
@@ -76,18 +76,16 @@ const config: ControlPanelConfig = {
     {
       label: t('Chart Options'),
       expanded: true,
-      controlSetRows: [
-        [showLegend, lineInterpolation],
-        ['color_picker', null],
-      ],
+      controlSetRows: [[showLegend], [lineInterpolation], ['color_picker', null]],
```
villebro, line 79: nit: `null` is redundant here (will remove next time I edit these files).

ktmud: I think it's for making sure the control is only of half-width? Not sure if it's still necessary with the new design though.

PR #914, opened 2021-01-25
fix(chart-controls): add a tooltip to span in ColumnOption, refactor RB Tooltips to Antd

kgabryje:
Replace RB Tooltips with Antd in `InfoTooltipWithTrigger` to unify UI, wrap metric name in a tooltip.
Change in `InfoTooltipWithTrigger` may affect multiple components in Explore
CC: @villebro

image

### DIFF
--- a/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
+++ b/packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
@@ -16,19 +16,29 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import React, { CSSProperties } from 'react';
+import React from 'react';
 import { kebabCase } from 'lodash';
-import { Tooltip, OverlayTrigger } from 'react-bootstrap';
 import { t } from '@superset-ui/core';
-
-const tooltipStyle: CSSProperties = { wordWrap: 'break-word' };
+import { Tooltip } from './Tooltip';
 
 export interface InfoTooltipWithTriggerProps {
   label?: string;
   tooltip?: string;
   icon?: string;
   onClick?: () => void;
-  placement?: string;
+  placement?:
+    | 'right'
+    | 'top'
+    | 'left'
+    | 'bottom'
+    | 'topLeft'
+    | 'topRight'
+    | 'bottomLeft'
+    | 'bottomRight'
+    | 'leftTop'
+    | 'leftBottom'
+    | 'rightTop'
+    | 'rightBottom';
   bsStyle?: string;
   className?: string;
 }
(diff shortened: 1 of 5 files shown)


villebro (approved):

packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
```diff
@@ -16,19 +16,29 @@
...
+    | 'left'
+    | 'bottom'
+    | 'topLeft'
+    | 'topRight'
+    | 'bottomLeft'
+    | 'bottomRight'
+    | 'leftTop'
+    | 'leftBottom'
+    | 'rightTop'
+    | 'rightBottom';
```
ktmud, line 41: Didn't test AntD's typing, but maybe you can replace this with `TooltipProps['placement']`?

packages/superset-ui-chart-controls/src/components/Tooltip.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { useTheme } from '@superset-ui/core';
+import { Tooltip as BaseTooltip } from 'antd';
+import { TooltipProps } from 'antd/lib/tooltip';
+
+export const Tooltip = (props: TooltipProps) => {
+  const theme = useTheme();
+  return (
+    <BaseTooltip
+      overlayStyle={{ fontSize: theme.typography.sizes.s, lineHeight: '1.6' }}
+      color={`${theme.colors.grayscale.dark2}e6`}
```
ktmud, line 11: Is the added semi-transparency required by design? Adding alpha channel this ways seems a little bit hacky. I could see this break if someone changed `theme.colors` to any non-hexcode format. But I don't know a good fix for this, so feel free to keep it (both adding a new color in `theme` or using `d3-color` for color calculation seem like an overkill).

Another suggestion, though, ideally these overrides (Superset default styles) should also allow run time overrides, which is not difficult to do:

```ts
export function Tooltip({
  overlayStyle,
  color,
  ...props
}: TooltipProps) {
  const theme = useTheme();
  const defaultColor = `${theme.colors.grayscale.dark2}e6`;
  return (
    <BaseTooltip
      overlayStyle={{ fontSize: theme.typography.sizes.s, lineHeight: '1.6', ...overlayStyle }}
      color={defaultColor || color}
      {...props}
    />
  );
};
```

ktmud:
Both comments are non-blocking nits btw.

packages/superset-ui-chart-controls/src/components/InfoTooltipWithTrigger.tsx
kgabryje: Good idea! We can event import `TooltipPlacement` directly

packages/superset-ui-chart-controls/src/components/Tooltip.tsx
kgabryje: I added the overrides.
Alpha channel has already been used like that in some other components in main superset repo, including Tooltip (this Tooltip is copied from superset). I think it wasn't required by design, but I wanted to make it look similar to the old R-B tooltip.

PR #919, opened 2021-01-28
feat(core): add orderby to extractQueryFields

ktmud:
🏆 Enhancements
🐛 Bug Fix

Add `orderby` to `extractQueryFields`, which consolidates query options related to generating actual SQL queries. 

This should also fix sorting options in table chart. Previously each plugin has to do the `order_by_choices` parsing themselves (e.g. in [the now deprecated old new table chart](https://github.com/apache-superset/superset-ui/blob/master/temporary-plugins/hold-potentially-deprecate/superset-ui-plugin-chart-table/src/buildQuery.ts#L12-L22)), I figure it'd be useful to make this global, in case we want to support query mode switch for other viz types.

Related: #889 https://github.com/apache/superset/pull/10270

### DIFF
--- a/plugins/plugin-chart-table/src/controlPanel.tsx
+++ b/plugins/plugin-chart-table/src/controlPanel.tsx
@@ -181,7 +181,7 @@ const config: ControlPanelConfig = {
             config: {
               type: 'SelectControl',
               label: t('Ordering'),
-              description: t('One or many metrics to display'),
+              description: t('Order results by selected columns'),
               multi: true,
               default: [],
               mapStateToProps: ({ datasource }) => ({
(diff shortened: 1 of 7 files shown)


plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -181,7 +181,7 @@ const config: ControlPanelConfig = {
             config: {
               type: 'SelectControl',
               label: t('Ordering'),
-              description: t('One or many metrics to display'),
+              description: t('Order results by selected columns'),
```
ktmud, line 184: Bycatch: wrong tooltip for the "Ordering".

villebro (approved):
LGTM

ktmud:
I'm going to admin merge since the blocking check "Semantic Pull Request" seems to be having a glitch.

PR #925, opened 2021-02-01
fix(plugin-chart-table): unnecessary post_processing in raw records mode

ktmud:
🐛 Bug Fix

Table chart should not send `post_processing` when in raw records mode.

Closes https://github.com/apache/superset/issues/12853

### Test Plan

Added new unit tests

To test in Superset, go to the "Games" sample chart:

global_sales

Which has a Raw Records query, the `global_sales` should display the same value as in the data samples.


zhaoyongjie (approved):
LGTM

PR #929, opened 2021-02-03
feat(core): add custom_params to form_data

amitmiran137:
🏆 Enhancements

📜 Documentation
Extra parameters that any plugins can choose to set for use in jinja templated queries

🐛 Bug Fix

🏠 Internal


ktmud:
How do you plan to expose this in Jinja and what would be a use case for these params?

I'm wondering whether this approach has any security risks.

Would exposing form data and slice metadata to Jinja template also suffice?

amitmiran137:
> How do you plan to expose this in Jinja and what would be a use case for these params?
> 
> I'm wondering whether this approach has any security risks.
> 
> Would exposing form data and slice metadata to Jinja template also suffice?

First I think we can assume that using jinja alone exposes a security risk that one should decide to live with.

Jinja can access the request object so I guess getting those params from the body shouldn't be hard

the use-case here is the following:
in a multi-tenant org, there are objects that entitled to different tenants and sometimes rules are complex and cannot be validated against the dataset with a simple RLS rule
instead, we intend to perform a custom external BE access validation with custom_params coming from a **custom-viz-plugin** in order to enforce the entitlement

we plan to use RLS with a custom jinja addon function that will do an API call with those custom_params to the relevant API's and will return only the entitled objects into the where clause within the RLS condition

villebro (commented):
Is this being picked up and available to use in the jinja context? If not, can we add that logic in this PR, as it shouldn't be too many additional LOC. It would also be nice to add a unit test (backend test would probably be enough).

villebro (approved):
Thought about this more, and figured this is simpler than I originally thought. LGTM. Would be nice to add documentation at some point about these hidden controls, as they're essentially reserved names in the control panel.

ktmud (commented):
Could you be more specific with what do those custom params look like? Are they tied to a specific user or a viz type? Does it have to be generated by JavaScript? If not, since you are going to create a custom Jinja addon function already, would exposing current viz type to the Jinja template env be enough?

amitmiran137:
> Could you be more specific with what do those custom params look like? Are they tied to a specific user or a viz type? Does it have to be generated by JavaScript? If not, since you are going to create a custom Jinja addon function already, would exposing current viz type to the Jinja template env be enough?

exposing the current viz type won't be enough.
actually, those custom params won't be anything that exists in the dataset probably
a custom param can look like:
`{ dataValidationParam: { tenantId, chosenSegmentIds:[1,2,3,4,5] , chosenPlatformTuples: [{AB5HZ,55}, {AB5HZ,54} ] }}`
those elements doesn't live within the dataset

ktmud:
Are they from a client request to another API or will they be dynamic based on user interaction with charts?

If they are fixed values per session, I believe Jinja template functions have full access to the `request` context, so if you can somehow tie them to session ids or save them in cookies, you may not need this.

Not trying to push back an otherwise useful addon, just looking for easier solutions (if possible) here.

amitmiran137:
> Are they from a client request to another API or will they be dynamic based on user interaction with charts?
> 
> If they are fixed values per session, I believe Jinja template functions have full access to the `request` context, so if you can somehow tie them to session ids or save them in cookies, you may not need this.
> 
> Not trying to push back an otherwise useful addon, just looking for easier solutions (if possible) here.

They from a client request to another API but and will be dynamic based on  user interaction with charts
But there will be cases that those Param will not appear at all in the formData . This is why we want to send them explicitly

ktmud (approved):
Makes sense. I was also wondering whether it makes sense to add these adhoc params to `url_param` or even `extras` (needs to update it's backend schema), but I guess it makes sense to keep them separate.

PR #930, opened 2021-02-03
fix(plugin-chart-table): refine ordering logic

villebro:
🐛 Bug Fix
Previously table chart would by default order by the first metric if no sort order was specified. This replicates the logic from the old `viz.py` implementation and also fixes a few minor issues in the current code:
- Infer `queryMode` from `formData` contents; if it's undefined and `metrics` is not empty, assume `QueryMode.aggregate`. This is the case in some old metadata, e.g. the World Bank example dashboard.
- `timeseries_limit_metric` is `[]` when the control is unset, and `QueryFormMetric` when set. The type is updated and a util is added for normalizing to `QueryFormMetric[]` + relevant tests.
- Added a missing license header.

### AFTER
image

### BEFORE
image

### DIFF
--- a/plugins/plugin-chart-table/src/buildQuery.ts
+++ b/plugins/plugin-chart-table/src/buildQuery.ts
@@ -1,22 +1,44 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -1,22 +1,44 @@
...
+      queryMode = QueryMode.aggregate;
+    }
     let postProcessing: PostProcessingRule[] = [];
 
     if (queryMode === QueryMode.aggregate) {
       // orverride orderby with timeseries metric when in aggregation mode
-      if (timeseriesLimitMetric && orderDesc != null) {
-        orderby = [[timeseriesLimitMetric, !orderDesc]];
+      if (timeseriesLimitMetric.length > 0 && orderDesc != null) {
+        orderby = [[timeseriesLimitMetric[0], !orderDesc]];
```
villebro, line 38: In practice `timeseriesLimitMetric` is always truthy, as the control always returns either an empty array (nothing selected) or a string/object (=an orderby is set).

ktmud: Wow, didn't notice the control returns an empty array when nothing is selected. That sounds wrong.

ktmud (approved):
LGTM. Thanks for the fix!

PR #935, opened 2021-02-04
fix(plugin-chart-table): always sort descending by first metric

ktmud:
🐛 Bug Fix

Follow up for #930 . When there is not "Sort by" metric specified, [the original logic](https://github.com/apache/superset/blob/32f2c45f93693156f1cf88c2c5223684c31d3f20/superset/viz.py#L746) pre [the API migration](https://github.com/apache-superset/superset-ui/pull/889) is to always sort by the first metric in descending order, regardless if "Sort descending" is set.

Did some refactoring on table chart's `buildQuery`:

- Introduce a more general `ensureIsArray` utils, as replacement for `extractTimeseriesLimitMetric`
- Change `queryMode` inference logic to be the same as in the control panel configs.

I'm also slipping a small refactor on `buildQueryObject` where `null` values are fell back to `undefined` so that `null`s are not sent to the backend. Even thought the backend has now taken care of `null`s as well: https://github.com/apache/superset/pull/12905

### DIFF
--- a/packages/superset-ui-core/src/query/buildQueryObject.ts
+++ b/packages/superset-ui-core/src/query/buildQueryObject.ts
@@ -47,22 +47,24 @@ export default function buildQueryObject<T extends QueryFormData>(
   });
 
   let queryObject: QueryObject = {
-    time_range,
-    since,
-    until,
-    granularity,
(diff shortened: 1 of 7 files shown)


villebro (approved):
Many nice improvements!

PR #936, opened 2021-02-04
feat(core): boring smart date formatting

ktmud:
💔 Breaking Changes

🏆 Enhancements

Make the smart date formatter "boring"---return a more complete (and accurate) but less adaptive (and sometimes longer) representation of the timestamp.

The reason for this change is that time marks like `".111ms"` or `".13s"` are of little use in most business context.

Sometimes the format also doesn't make sense when it is applied to x-axis bounds:

(It's impossible to tell which month `Mon 06` is referring to).

### Before

before-adaptive

### After

adaptive-formatting


ktmud:
This is breaking change. I'll update the test cases if folks think this is the right way to go.

junlincc (approved):
make sense from the product standpoint, also checked with end-users to confirm this is direction we wanna go. thanks for the improvement! @ktmud 🙏

ktmud:
One concern with this may introduce very long x-axis ticks for xy charts (line, bar, area, etc.).

After discussing this with @kristw ,  I'm inclined to limit this to table charts only and leave the work for smarter axis labels to later.

villebro:
I think the old behavior is better for the x-axis on D3 charts, as it adapts the ticks to fit well with the truncated format. Not sure how we'd do this more elegantly, but it seems like "smart" should apply the one proposed here whenever it's used as pure text (tooltips, tables) but the old one when used on chart axes.

villebro:

PR #937, opened 2021-02-04
fix(plugin-chart-table): better rendering for temporal columns

ktmud:
🐛 Bug Fix

Display temporal columns more smartly. I.e. when in "Adaptive formatting":

1. if the backend returns them as string, display as is
2. if they are numbers, then assume they are timestamps and display as full UTC timestamps

Also fixes a bug where wrong query is generated when "Include time" is selected then switch to run raw records queries.

### Test Plan

Tested with this virtual dataset in PostgreSQL:

```
SELECT TIMESTAMP '2021-01-01T11:12:33.113' as ts, '2021-01-01T11:12:33' AS ts_str, 'aa' as text, 10000000 as num
UNION
SELECT TIMESTAMP '2021-01-01T00:00:00', '2021-01-09T10:11:22', 'ccc', 11.114491
UNION
SELECT TIMESTAMP '2021-01-09T11:10:33', '2021-01-09T10:11:22', 'ccc', 11.114491
UNION
SELECT TIMESTAMP '2021-01-09T11:10:00', '2021-01-09T10:11:22', 'abc', 11.114491
UNION
SELECT TIMESTAMP '2021-01-09T11:00:00', '2021-01-09T10:11:22', 'ama', 11.114491
UNION
SELECT TIMESTAMP '2021-01-09T00:00:00', '2021-01-09T10:11:22', 'abc', 11.114491
UNION
SELECT TIMESTAMP '2021-02-07T00:00:00', '2021-01-09T10:11:22', 'abc', 11.114491
UNION
SELECT TIMESTAMP '2021-02-08T00:00:00', '2021-01-09T10:11:22', 'abc', 11.114491
```

### Before

before-format

before-format-aggregate

### After

after-format

after-format-aggregate

### DIFF
--- a/plugins/plugin-chart-table/src/buildQuery.ts
+++ b/plugins/plugin-chart-table/src/buildQuery.ts
@@ -46,7 +46,17 @@ export default function buildQuery(formData: TableChartFormData) {
   const { percent_metrics: percentMetrics, order_desc: orderDesc = false } = formData;
   const queryMode = getQueryMode(formData);
   const sortByMetric = ensureIsArray(formData.timeseries_limit_metric)[0];
-  return buildQueryContext(formData, baseQueryObject => {
+  let formDataCopy = formData;
+
+  // never include time in raw records mode
+  if (queryMode === QueryMode.raw) {
+    formDataCopy = {
+      ...formData,
+      include_time: false,
+    };
+  }
+
+  return buildQueryContext(formDataCopy, baseQueryObject => {
     let { metrics, orderby } = baseQueryObject;
     let postProcessing: PostProcessingRule[] = [];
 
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -46,6 +46,12 @@ export default function buildQuery(formData: TableChartFormData) {
   const { percent_metrics: percentMetrics, order_desc: orderDesc = false } = formData;
   const queryMode = getQueryMode(formData);
   const sortByMetric = ensureIsArray(formData.timeseries_limit_metric)[0];
+
+  // never include time in raw records mode
+  if (queryMode === QueryMode.raw) {
+    formData.include_time = false;
+  }
```
ktmud, line 53: I'm just disabling selecting the Time column for raw records mode. We may want to allow that in the future though because sometimes it's useful to add truncated dates even for raw records and the time column provides an easy way to do that.

villebro (approved):
LGTM - tested and works as expected

ktmud:
Thanks for reviewing and taking care of the release.

PR #942, opened 2021-02-07
fix(plugin-chart-echarts): label type and hover bugs

villebro:
🐛 Bug Fix

Bump ECharts to 5.0.2 to fix hover bug (https://github.com/apache/echarts/pull/14108) and label type bug. Also clean up control panel so that all customization fields are on a single row.

### After
image

### Before
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
@@ -38,8 +38,6 @@ const {
   showLabels,
 } = DEFAULT_FORM_DATA;
 
-const noopControl = { name: 'noop', config: { type: '', renderTrigger: true } };
-
 const config: ControlPanelConfig = {
   controlPanelSections: [
     sections.legacyRegularTime,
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
```diff
@@ -52,17 +50,18 @@ const config: ControlPanelConfig = {
...
-        [legendMarginControl, noopControl],
+        [legendTypeControl],
+        [legendOrientationControl],
+        [legendMarginControl],
         // eslint-disable-next-line react/jsx-key
         [<h1 className="section-header">{t('Labels')}</h1>],
         [
           {
-            name: 'pie_label_type',
+            name: 'label_type',
```
ktmud, line 64: Does renaming this field matter here? Would you need a db migration for this?

ktmud (approved):

plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx
villebro: Yes, we need a migration - I'll add one to the bump PR.

ktmud:
Linking related db migration PR: https://github.com/apache/superset/pull/13052/files

PR #943, opened 2021-02-08
feat(core): expand native filter hook and add chart metadata fields

simcha90:
This PR:

1. Update definition of `SetExtraFormDataHook`
2. change `isNativeFilter` metadata flag to `behaviours` property that contains array of implemented behaviours
3. Removing Antd filters from repository

💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal


packages/superset-ui-demo/storybook/stories/plugins/plugin-filter-antd/Select/Stories.tsx
```diff
@@ -1,42 +0,0 @@
-import React from 'react';
-import { action } from '@storybook/addon-actions';
-import { boolean, withKnobs } from '@storybook/addon-knobs';
-import { SuperChart, getChartTransformPropsRegistry } from '@superset-ui/core';
-import { AntdSelectFilterPlugin } from '@superset-ui/plugin-filter-antd';
-import transformProps from '@superset-ui/plugin-filter-antd/lib/Select/transformProps';
-import data from './data';
-import { withResizableChartDemo } from '../../../../shared/components/ResizableChartDemo';
-import 'antd/dist/antd.css';
```
villebro, line 9: Just a reminder: we need migrate these to the `apache/superset` storybook.

villebro (commented):
A few comments

packages/superset-ui-demo/storybook/stories/plugins/plugin-filter-antd/Select/Stories.tsx
simcha90: ok

packages/superset-ui-core/src/chart/types/Base.ts
```diff
@@ -2,8 +2,19 @@ import { ExtraFormData } from '../../query';
 
 export type HandlerFunction = (...args: unknown[]) => void;
 
+export enum Behaviour {
+  CROSS_FILTER = 'CROSS_FILTER',
+  NATIVE_FILTER = 'NATIVE_FILTER',
+}
```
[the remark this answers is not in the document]
simcha90: done

villebro (approved):
LGTM

packages/superset-ui-core/src/chart/models/ChartMetadata.ts
```diff
@@ -1,3 +1,5 @@
+import { Behavior } from '../types/Base';
```
ktmud, line 1: Would prefer this to be called `ChartFilterBehavior` (after first-time seeing it in https://github.com/apache/superset/pull/13052 and being confused of what it is), but looks like I'm late to the party...

villebro: It won't necessarily be restricted to filtering behavior, for example drill-downs which are essentially groupbys, but `ChartBehavior` could also be ok. Let me know if you want to update this now, otherwise we'll stick to this naming.

ktmud: Let's save the update to when we need to expand this functionality.

PR #955, opened 2021-02-11
fix(legacy-plugin-chart-country-map): Bulgaria map is not correctly displayed in Country Map chart

### LINKED ISSUE #569: Bulgaria map is not correctly displayed in Country Map chart
Hi,
I've tryed to create Country Map chart to locate some objects in Bulgaria and I've run across the following issues:
The codes of region/province/department, required by the the input when creating country map, are not in compliance with ISO-3166-2 in packages/superset-ui-legacy-plugin-chart-country-map/src/countries/bulgaria.geojson. Ok, then I changed the codes in my source DB table to be in the format like in the mentioned .geojson file and the Bulgaria map was displayed. Unfortunately - it is not correctly displayed - it is displayed like a small point in the upper corner of some rectangle.
I have looked at the DEMO  for superset-ui in Github - the issue thing there. When you select Bulgaria from the dropdown, only a point inside a rectangle is shown.
You can check it here:
https://apache-superset.github.io/superset-ui/?path=/story/legacy-chart-plugins-legacy-plugin-chart-country-map--basic

djok:
🐛 Bug Fix #569


zeridon:
Additionally with regards to #569 i would like to point that zooming the map (by clicking on a region) works only if clicking on borders of regions (for bulgaria). Is this related to the geojson?

mayurnewase:
Hi @djok what algo did you use to convert this file,as I am seeing different schema and co-ordinate values as well.
I know one more issue which I think is of same type here (https://github.com/apache/superset/issues/12987).
If you could take a look at it,or just give me this conversion logic,I will fix it as well.

djok:
> Additionally with regards to #569 i would like to point that zooming the map (by clicking on a region) works only if clicking on borders of regions (for bulgaria). Is this related to the geojson?

No, zooming the map is not related to geojson

bdiouih:
> Hi @djok what logic did you use to convert this file,as I am seeing different schema and co-ordinate values as well.
> I know one more issue which I think is of same type here ([apache/superset#12987](https://github.com/apache/superset/issues/12987)).
> If you could take a look at it,or just give me this conversion logic,I will fix it as well.

I second that : the same issue is encountered with other geojson files.
Please detail how the conversion is done.

ktmud:
@mayurnewase Do you need help with this issue? I can also take a look since our users are reporting similar issues with the map of Spain.

mayurnewase:
> @mayurnewase Do you need help with this issue? I can also take a look since our users are reporting similar issues with the map of Spain.

@ktmud Didn't get to spend time on it, tried with different zoom, scale values but didn't help. I will get on it next week for sure.

zeridon:
After some digging the following was discovered:

 * ISO Codes are not compliant to ISO 3661-2 (https://en.wikipedia.org/wiki/ISO_3166-2:BG)
 * The shapes included are for municipalities but are expected to be for districts

@djok Can you please prepare a geojson with districts only and ISO codes as per ISO 3166-2. Good starting point is: https://github.com/yurukov/Bulgaria-geocoding

djok:
Yes, I will prepare geojson according to ISO 3661-2, but we still need geojson map with municipalities.

Can I submit new country like "Bulgaria (municipalities)" so we keep municipalities level?

zeridon:
My personal oppinion is that the country map plugin does not allow for that. The visualization itself asks for ISO-3661-2 Code. Though that field can be overloaded i am not sure if this is a good idea.

Other than that the approach seems fine to me. On the other hand using NUTS3/4 or LAU-1 might be a better option (if possible to integrate cleanly)

mayurnewase:
@ktmud,I tried debugging with a fixed bounds scale but couldn't find the root cause.
Tried with geojson of only 2 provinces,and after removing multiple co-ordinates from polygon, it displays correctly, I am not sure if the issue is in d3 lib itself.

ktmud:
@mayurnewase Thanks for the update! I'll take a look, too, later this week. I think we may need to find a new datasource to re-generate newer geojson files as https://www.diva-gis.org/gdata seems pretty outdated.

zeridon:
@ktmud What about https://gadm.org/data.html

At least to my untrained eye looks decent enough.

ktmud:
@zeridon GADM data does not allow unauthorized use in redistribution and commercial softwares. 

I was checking https://www.naturalearthdata.com/downloads/ the other day and it looked good.

You can download the [1:10m Admin 1](https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-1-states-provinces/) data,  load it with [GeoPandas](https://geopandas.org/) and convert it to GeoJSON country by country. I'd also recommended to [simplify geometry](https://geopandas.org/docs/user_guide/geometric_manipulations.html#GeoSeries.simplify) for bigger countries so the geojson file can be smaller. You'll need to write some Python script for it, but it shouldn't be that hard.

After all these are done, it is recommended to put the Python script within the plugin folder so future maintainers can reproduce the process for any potential updates.

villebro:

PR #957, opened 2021-02-13
fix(plugin-chart-table): hide cell bar for group by fields

ktmud:
🏆 Enhancements

Hide table chart cell bars for group by fields, as normally they could be considered categorical and not numeric metrics.

For raw records, we could theoretically use the "IS DIMENSION" flag, but it seems to be set to True by default, even for numeric columns:

image

So I'm not changing the behavior for raw records mode.

### Before

with-cell-bar

### DIFF
--- a/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/plugins/plugin-chart-table/src/TableChart.tsx
@@ -146,6 +146,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
     height,
     width,
     data,
+    isRawRecords,
     columns: columnsMeta,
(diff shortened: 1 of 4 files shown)


maloun96:
@ktmud is it ok to change directly yarn.lock?

ktmud:
> @ktmud is it ok to change directly yarn.lock?

Depending on how much it changed.

Theoretically you should never need to do it in a non-related PR but this diff pops up after I ran "yarn install" and it looks pretty straightforward so I'm including it nonetheless.

villebro (approved):
I agree with this change

PR #959, opened 2021-02-15
feat(self-trigger): add backend pagination for the table

simcha90:
This PR add backend pagination for `plugin-chart-table`.
Requires: https://github.com/apache/superset/pull/13137
As part of this request done next changes:
1. Getting ownState of table chart from dashboard with `{ pageSize, currentPage }`
2. It's update `buildQuery` of this plugin with adding `row_offset` and `row_limit`:
Because BE still not support returning of total rows available for current request, we do request for number items per page `+1` to know if there is available one more page. For same reason we use only `previous/next` button in pagination
3. Because we needed update definition of `setExtraFormData` hook, we also updated it's name to more relevant for it's action: `setDataMask`
4. Added `custom_form_data` to `extraFormData` object to pass custom props for build necessary request to the server in `buildQuery`

https://user-images.githubusercontent.com/56388545/107976853-e4544a80-6fc2-11eb-9eb6-988b07a755f7.mov

### DIFF
--- a/packages/superset-ui-core/src/chart/types/Base.ts
+++ b/packages/superset-ui-core/src/chart/types/Base.ts
@@ -7,25 +7,25 @@ export enum Behavior {
   NATIVE_FILTER = 'NATIVE_FILTER',
 }
 
+export type DataMaskCurrentState = { value?: any; [key: string]: any };
+
+export type DataMask = {
+  nativeFilters?: {
+    extraFormData?: ExtraFormData;
+    currentState: DataMaskCurrentState;
+  };
+  crossFilters?: {
+    extraFormData?: ExtraFormData;
+    currentState: DataMaskCurrentState;
+  };
+  ownFilters?: {
+    extraFormData?: ExtraFormData;
+    currentState: { [key: string]: any };
+  };
+};
+
 export type SetDataMaskHook = {
-  ({
-    nativeFilters,
-    crossFilters,
-    ownFilters,
-  }: {
-    nativeFilters?: {
-      extraFormData: ExtraFormData;
-      currentState: { value: any; [key: string]: any };
-    };
-    crossFilters?: {
-      extraFormData: ExtraFormData;
-      currentState: { [key: string]: any };
-    };
-    ownFilters?: {
-      extraFormData: ExtraFormData;
-      currentState: { [key: string]: any };
-    };
-  }): void;
+  ({ nativeFilters, crossFilters, ownFilters }: DataMask): void;
 };
 
 export interface PlainObject {
--- a/packages/superset-ui-core/src/chart/types/Base.ts
+++ b/packages/superset-ui-core/src/chart/types/Base.ts
@@ -9,19 +9,19 @@ export enum Behavior {
 
 export type SetDataMaskHook = {
   ({
-    native,
-    cross,
-    own,
+    nativeFilters,
+    crossFilters,
+    ownFilters,
   }: {
-    native?: {
+    nativeFilters?: {
       extraFormData: ExtraFormData;
       currentState: { value: any; [key: string]: any };
     };
-    cross?: {
+    crossFilters?: {
       extraFormData: ExtraFormData;
       currentState: { [key: string]: any };
     };
-    own?: {
+    ownFilters?: {
       extraFormData: ExtraFormData;
       currentState: { [key: string]: any };
     };
--- a/packages/superset-ui-core/src/chart/types/Base.ts
+++ b/packages/superset-ui-core/src/chart/types/Base.ts
@@ -7,13 +7,24 @@ export enum Behavior {
   NATIVE_FILTER = 'NATIVE_FILTER',
 }
 
-export type SetExtraFormDataHook = {
+export type SetDataMaskHook = {
   ({
-    extraFormData,
-    currentState: { value },
+    native,
+    cross,
+    own,
   }: {
-    extraFormData: ExtraFormData;
-    currentState: { value: any; [key: string]: any };
+    native?: {
+      extraFormData: ExtraFormData;
+      currentState: { value: any; [key: string]: any };
+    };
+    cross?: {
+      extraFormData: ExtraFormData;
+      currentState: { [key: string]: any };
+    };
+    own?: {
+      extraFormData: ExtraFormData;
+      currentState: { [key: string]: any };
+    };
   }): void;
 };
 
--- a/packages/superset-ui-core/src/query/processExtraFormData.ts
+++ b/packages/superset-ui-core/src/query/processExtraFormData.ts
@@ -11,7 +11,7 @@
  *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIutils/externalAPIs.tsES OR CONDITIONS OF ANY
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
--- a/packages/superset-ui-core/src/query/processExtraFormData.ts
+++ b/packages/superset-ui-core/src/query/processExtraFormData.ts
@@ -18,7 +18,14 @@
  */
 import { QueryObject } from './types/Query';
 
-const ALLOWED_OVERRIDES = ['time_grain_sqla', 'time_range', 'since', 'until'];
+const ALLOWED_OVERRIDES = [
+  'time_grain_sqla',
+  'time_range',
+  'since',
+  'until',
+  'row_offset',
+  'row_limit',
+];
 const ALLOWED_APPENDS = ['adhoc_filters', 'filters', 'groupby'];
 
 export function overrideExtraFormData(
--- a/packages/superset-ui-core/src/query/processExtraFormData.ts
+++ b/packages/superset-ui-core/src/query/processExtraFormData.ts
@@ -11,21 +11,14 @@
  *
  * Unless required by applicable law or agreed to in writing,
  * software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * "AS IS" BASIS, WITHOUT WARRANTIutils/externalAPIs.tsES OR CONDITIONS OF ANY
  * KIND, either express or implied.  See the License for the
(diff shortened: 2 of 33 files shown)


ktmud (commented):
Thanks for working on this highly-demanded feature! A lot of users would be grateful!

Since the queryContext can now run multiple queries, I'm wondering how difficult would it be to let `buildQuery` add an additional `SELECT count(*) FROM {...subquery...}` query, which may be implemented by adding another [`ChartDataResultType`](https://github.com/apache/superset/blob/ac73991913c596a0b2b37aaf96083c227088dad6/superset/common/query_context.py#L167)?

amitmiran137:
It was difficult to achieve that right now 
I think we could get to that together with the row totals feature

villebro:
> Since the queryContext can now run multiple queries, I'm wondering how difficult would it be to let `buildQuery` add an additional `SELECT count(*) FROM {...subquery...}` query, which may be implemented by adding another [`ChartDataResultType`](https://github.com/apache/superset/blob/ac73991913c596a0b2b37aaf96083c227088dad6/superset/common/query_context.py#L167)?

This is definitely on the short term roadmap. I didn't look to closely into how much work this will be, but I believe we need to do some minor refactoring to achieve this. I had been planning on adding some optional props from `QueryContext` to `QueryObject` to make it possible to introduce support for querying data from multiple different datasources in a single `QueryContext`, and I believe the same could be relevant for `result_type` (=mixing `RESULTS` with `SAMPLES` or the forthcoming `ROWCOUNT`).

Anyway, let's defer this feature until the backend feature is merged, as there's no clean ANSI SQL way to achieve multi-column COUNT DISTINCTs without a subquery.

packages/superset-ui-core/src/chart/types/Base.ts
```diff
@@ -7,13 +7,24 @@ export enum Behavior {
...
+      currentState: { value: any; [key: string]: any };
+    };
+    cross?: {
+      extraFormData: ExtraFormData;
+      currentState: { [key: string]: any };
+    };
+    own?: {
+      extraFormData: ExtraFormData;
+      currentState: { [key: string]: any };
+    };
```
villebro, line 27: When seeing this like this, I think new developers will be confused by what `native`, `cross` and `own` mean. Not sure I have better proposals, but at the risk of being misleading if the scope of this functionality is expanded more, I think `nativeFilters`, `crossFilters` and `ownFilters` could be slightly more descriptive. Alternatively we might want to add a comment per prop to help understand what the context is.

packages/superset-ui-core/src/query/types/QueryFormData.ts
```diff
@@ -100,6 +101,7 @@ export type QueryFormExtraFilter = {
 export type ExtraFormData = {
   append_form_data?: Partial<QueryObject>;
   override_form_data?: Partial<QueryObject>;
+  custom_form_data?: JsonObject;
```
villebro, line 104: A comment here could also be helpful (also add to the other two)

packages/superset-ui-core/src/query/processExtraFormData.ts
```diff
@@ -18,7 +18,14 @@
...
 import { QueryObject } from './types/Query';
 
-const ALLOWED_OVERRIDES = ['time_grain_sqla', 'time_range', 'since', 'until'];
+const ALLOWED_OVERRIDES = [
+  'time_grain_sqla',
+  'time_range',
+  'since',
+  'until',
+  'row_offset',
+  'row_limit',
```
villebro, line 27: These might no longer be needed

plugins/plugin-chart-table/src/DataTable/utils/externalAPIs.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { SetDataMaskHook } from '@superset-ui/core';
+
+export const updateExternalFormData = (
+  setDataMask: SetDataMaskHook,
+  pageNumber: number,
+  pageSize: number,
+) =>
+  setDataMask({
+    // @ts-ignore
```
villebro, line 28: is the ignore needed?

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -29,10 +36,14 @@ import {
   Row,
 } from 'react-table';
 import { matchSorter, rankings } from 'match-sorter';
+import { SetDataMaskHook } from '@superset-ui/core';
 import GlobalFilter, { GlobalFilterProps } from './components/GlobalFilter';
 import SelectPageSize, { SelectPageSizeProps, SizeOption } from './components/SelectPageSize';
 import SimplePagination from './components/Pagination';
 import useSticky from './hooks/useSticky';
+import { updateExternalFormData } from './utils/externalAPIs';
+import BackendPagination from './components/BackendPagination';
```
villebro, line 45: `ServerPagination` or `DatabasePagination` might be a more appropriate name

ktmud:
Another question, how does this interact with percent metrics? Are we OK with percent metrics only be calculated based on limited rows? It'd be somewhat pointless to calculate distribution only on one page of data (10 ~ 200 rows) when the total number of records may be millions.

villebro:
> Another question, how does this interact with percent metrics? Are we OK with percent metrics only be calculated based on limited rows? It'd be somewhat pointless to calculate distribution only on one page of data (10 ~ 200 rows) when the total number of records may be millions.

This will need to be addressed in a forthcoming PR - once we have the second query for totals we can calculate the percent mertrics in the frontend, bypassing the need for the backend postprocessing and always getting percent distributions for the full dataset.

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -109,7 +129,7 @@ export default function DataTable<D extends object>({
     }
     return undefined;
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [initialHeight, initialWidth, wrapperRef, hasPagination, hasGlobalControl]);
+  }, [initialHeight, initialWidth, wrapperRef, hasPagination, hasGlobalControl, showBENextButton]);
```
villebro, line 132: showNextButton

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -29,10 +36,14 @@ import {
...
 } from 'react-table';
 import { matchSorter, rankings } from 'match-sorter';
+import { SetDataMaskHook } from '@superset-ui/core';
 import GlobalFilter, { GlobalFilterProps } from './components/GlobalFilter';
 import SelectPageSize, { SelectPageSizeProps, SizeOption } from './components/SelectPageSize';
 import SimplePagination from './components/Pagination';
 import useSticky from './hooks/useSticky';
+import { updateExternalFormData } from './utils/externalAPIs';
+import ServerPagination from './components/ServerPagination';
+import { BackendPage } from '../types';
```
villebro, line 46: For consistency, this should probably  be `ServerPage`

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -43,9 +54,13 @@ export interface DataTableProps<D extends object> extends TableOptions<D> {
   hooks?: PluginHook<D>[]; // any additional hooks
   width?: string | number;
   height?: string | number;
+  backendPagination?: boolean;
```
villebro, line 57: ..serverPagination

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -66,9 +82,12 @@ export default function DataTable<D extends object>({
   maxPageItemCount = 9,
   sticky: doSticky,
   searchInput = true,
+  setDataMask,
+  showBENextButton,
   selectPageSize,
   noResults: noResultsText = 'No data found',
   hooks,
+  backendPagination,
```
villebro, line 90: ..serverPagination

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -43,9 +54,13 @@ export interface DataTableProps<D extends object> extends TableOptions<D> {
   hooks?: PluginHook<D>[]; // any additional hooks
   width?: string | number;
   height?: string | number;
+  backendPagination?: boolean;
+  setDataMask: SetDataMaskHook;
+  currentPage?: number;
   pageSize?: number;
   noResults?: string | ((filterString: string) => ReactNode);
   sticky?: boolean;
+  showBENextButton: boolean;
```
villebro, line 63: Could this just be `showNextButton`?

plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -192,7 +192,39 @@ const config: ControlPanelConfig = {
...
-        ['row_limit'],
+        [
+          {
+            name: 'backend_pagination',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Backend pagination'),
+              description: t('Enable backend pagination'),
+              default: false,
+            },
```
villebro, line 203: The label should probably be "Server pagination" and the description "Enable server side pagination of results"

villebro (commented):
Mostly minor comments, looking good. However, I'm having trouble getting this to work, but it appears to be related to linking working differently on `npm@7` (I believe I had this working a few days ago, but now I can't get it to work). @ktmud - is it working ok for you?

ktmud:
@villebro I didn't test this PR, but with npm 7, you need to manually run `npm i --legacy-peer-deps` in the plugin package after running `npm link`.

simcha90:
> Mostly minor comments, looking good. However, I'm having trouble getting this to work, but it appears to be related to linking working differently on `npm@7` (I believe I had this working a few days ago, but now I can't get it to work). @ktmud - is it working ok for you?

but it also requires: https://github.com/apache/superset/pull/13137 to pull out

villebro:
Thanks @simcha90 - I had indeed linked from the wrong branch. One last comment, as this feature isn't quite feature ready yet (slightly misleading percent metrics, no totals, no page numbers), perhaps we should the pagination control behind the cross filtering FF? We could then remove it when we have all features implemented and are certain that this works robustly.

villebro (approved):

nuevoleonkx:
On which version is available this feature? I'm testing on 1.1 but apparently is not covered here. In addition, what would be the best approach to upgrade from 0.36? TIA

PR #961, opened 2021-02-16
fix(legacy-plugin-chart-paired-t-test): fix paired t-test table chart

kgabryje:
This PR fixes several bugs that were breaking Paired t-test table chart.
1. Chart was using a metric name as a to transform data, which worked only with saved metrics, since backend returns a saved metric as a string. If an adhoc metric was used, chart was still trying to use it as a key, even though it's an object, which led to an 'undefined' error.
2. `reactable-arc` was not rendering table body. I replaced it with the latest version of `reactable`.
3. CSS for scrollable table content was buggy, caused chart to have 0px height. 

What is left to decide is how to handle a case when no group by column was specified. The chart does render when group by is not specified, but it doesn't make much sense. What we can do is either display an error message (see video) or leave it as it is (see screenshot). If we go for the first option, we need to figure out a proper, descriptive error message - the one in the video is just an example.

https://user-images.githubusercontent.com/15073128/108093700-4aea6e80-707e-11eb-8cda-a83ef14df964.mov
image

UPDATE:
After adding a validator for groupby, user cannot run query without specifying at least 1 group by column.
image

CC: @villebro @ktmud @rusackas @junlincc

### DIFF
--- a/plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
+++ b/plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
@@ -19,7 +19,7 @@
 /* eslint-disable react/no-array-index-key, react/jsx-no-bind */
 import dist from 'distributions';
 import React from 'react';
-import { Table, Tr, Td, Thead, Th } from 'reactable-arc';
+import { Table, Tr, Td, Thead, Th } from 'reactable';
 import PropTypes from 'prop-types';
 
 export const dataPropType = PropTypes.arrayOf(
--- a/plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
+++ b/plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
@@ -166,6 +166,11 @@ class TTestTable extends React.Component {
   render() {
     const { data, metric, groups } = this.props;
     const { control, liftValues, pValues } = this.state;
+
+    if (!Array.isArray(groups) || groups.length === 0) {
+      throw Error('Group by param is required');
+    }
+
     // Render column header for each group
     const columns = groups.map((group, i) => (
       <Th key={i} column={group}>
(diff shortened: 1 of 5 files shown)


plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
```diff
@@ -166,6 +166,11 @@ class TTestTable extends React.Component {
   render() {
     const { data, metric, groups } = this.props;
     const { control, liftValues, pValues } = this.state;
+
+    if (!Array.isArray(groups) || groups.length === 0) {
+      throw Error('Group by param is required');
+    }
```
ktmud, line 172: You can probably add `validators: [validateNonEmpty]` to the chart control, too:

```ts
      controlSetRows: [
        ['metrics'],
        ['adhoc_filters'],
        [
          {
            name: 'groupby',
            override: {
              validators: [validateNonEmpty],
            },
          },
        ],
```

ktmud (approved):
LGTM

plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
kgabryje: Good point! Done.
Is there any documentation for control panel config besides `ControlPanelConfig` interface?

kgabryje:
UPDATE:
After adding a validator for groupby, user cannot run query without specifying at least 1 group by column.
image

plugins/legacy-plugin-chart-paired-t-test/src/TTestTable.jsx
ktmud: There are some comments in `chart-controls`: https://github.com/apache-superset/superset-ui/blob/6b9cc586b3d8b56a53033b34e861f3f4a9e33a10/packages/superset-ui-chart-controls/src/types.ts#L149

But it's not very easy to find. We should probably start writing better README's.

villebro (approved):

junlincc:
Screen Shot 2021-02-25 at 9 37 38 P@kgabryje src=
sorry for the late sign-off, @kgabryje when you have a chance, let's remove the second group-by. thanks 🙏 

we also need to find a dataset to properly test it.

kgabryje:
@junlincc Ugh, what an embarassing bug 🤦 Fixed in https://github.com/apache-superset/superset-ui/pull/979

PR #962, opened 2021-02-17
refactor(chart-controls): improve typing and file organization

ktmud:
🏆 Enhancements
🏠 Internal

Improve typing and file organization for shared controls.

1. Reuse `Metric` definition in `superset-ui/core`.
2. Move the React components for shared controls to its own folder (prepare for adding more custom controls).
3. Export empty objects in pure typing files to fix build errors with `npm link`:
   cannot-read-call

### DIFF
--- a/packages/superset-ui-core/src/chart/index.ts
+++ b/packages/superset-ui-core/src/chart/index.ts
@@ -19,6 +19,6 @@ export * from './types/Base';
 export * from './types/TransformFunction';
 export * from './types/QueryResponse';
 
-export { default as __hask_reexport_chart_Base } from './types/Base';
-export { default as __hask_reexport_chart_TransformFunction } from './types/TransformFunction';
-export { default as __hask_reexport_chart_QueryResponse } from './types/QueryResponse';
+export { default as __hack_reexport_chart_Base } from './types/Base';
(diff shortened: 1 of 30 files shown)


packages/superset-ui-core/src/chart/index.ts
```diff
@@ -18,3 +18,7 @@ export { default as ChartDataProvider } from './components/ChartDataProvider';
 export * from './types/Base';
 export * from './types/TransformFunction';
 export * from './types/QueryResponse';
+
+export { default as __hack_reexport_chart_Base } from './types/Base';
+export { default as __hack_reexport_chart_TransformFunction } from './types/TransformFunction';
+export { default as __hack_reexport_chart_QueryResponse } from './types/QueryResponse';
```
ktmud, line 24: This hack fixes script errors in dev build when re-exporting pure type files:

image

Not sure what is the root cause, but this fixes it.

felixcodes (approved):
From an offline conversation with Jesse, `export default {};` is a hack solution for re-exporting * from pure type files that don't contain real JavaScript variables. 

Jesse explained that there is a failure under certain circumstances that Webpack rebuild will fail and throw an error that files do not exist without the hack.

PR LGTM

PR #965, opened 2021-02-17
fix(legacy-plugin-chart-map-box): fix downloading map as image

kgabryje:
Fixes downloading MapBox chart as image. Part of https://github.com/apache/superset/pull/13181
country-of-citizenship-2021-02-17T17-01-42 411Z

CC: @villebro @junlincc


villebro (approved):
LGTM - I was not able to see any degradation in chart performance.

PR #967, opened 2021-02-19
feat: add dropGroupByControl

zhaoyongjie:
🏆 Enhancements
POC of Explore drag and drop query panel. scope - Groupby control only.
link to: https://github.com/apache/superset/pull/13210


ktmud:
How about we name this `dndColumnSelectControl`? It's a control for selecting columns with drag and drop.

zhaoyongjie:
> How about we name this `dndColumnSelectControl`? It's a control for selecting columns with drag and drop. We have deprecated `groupby` in QueryContext, so it's better to be consistent in control naming, too.

All right. I will rename it.

ktmud (approved):
lgtm

PR #969, opened 2021-02-20
fix(chart-controls): expose datasource for some controls

ktmud:
🏠 Internal

Expose current `datasource` to the state of `adhoc_filters` and `time_range` control and clean up the `provideFormDataToProps` option. We should just be explicit what prop to expose in `mapStateToProps` so controls don't get unnecessarily re-rendered because of form data updates from other unrelated controls.

`superset-frontend` has some related changes: https://github.com/apache/superset/pull/13221

### Test Plan

Manual verification in Superset via npm link:

- Make sure both controls still work
- Make sure changing dataset will reset time filter to default value ("Last week")


villebro (approved):
LGTM

PR #970, opened 2021-02-22
chore: upgrade node to v14

ktmud:
🏆 Enhancements
🏠 Internal

superset-frontend has upgrade node to v14. Let's do it for superset-ui as well.

Related https://github.com/apache/superset/pull/13215


etr2460 (approved):
the yarn.lock file change is a bit weird. Otherwise lgtm

ktmud:
@etr2460 I think it was from #961 but `yarn.lock` was not updated with that PR.

PR #978, opened 2021-02-25
feat: implement drag and drop filters controls

kgabryje:
🏆 Enhancements

Based on https://github.com/apache-superset/superset-ui/pull/967 by @zhaoyongjie 

CC: @villebro @junlincc

### DIFF
--- a/packages/superset-ui-chart-controls/src/shared-controls/dndControls.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/dndControls.tsx
@@ -70,3 +70,18 @@ export const dndEntity: typeof dndGroupByControl = {
   validators: [validateNonEmpty],
   description: t('This defines the element to be plotted on the chart'),
 };
+
+export const dnd_adhoc_filters: SharedControlConfig<'DndFilterSelect'> = {
+  type: 'DndFilterSelect',
+  label: t('Filters'),
+  default: null,
+  description: '',
+  mapStateToProps: ({ datasource, form_data }) => ({
+    columns: datasource?.columns.filter(c => c.filterable) || [],
+    savedMetrics: datasource?.metrics || [],
+    // current active adhoc metrics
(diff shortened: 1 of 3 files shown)


packages/superset-ui-chart-controls/src/shared-controls/dndControls.tsx
```diff
@@ -0,0 +1,87 @@
...
+  ...dndGroupByControl,
+  label: t('Entity'),
+  default: null,
+  multi: false,
+  validators: [validateNonEmpty],
+  description: t('This defines the element to be plotted on the chart'),
+};
+
+export const dnd_adhoc_filters: SharedControlConfig<'FilterColumnSelect'> = {
+  type: 'FilterColumnSelect',
```
zhaoyongjie, line 75: please use `dnd` as the control name prefix.

zhaoyongjie:
I personally suggest not to make generic components now. The main reason is that these components are completely inconsistent in function, and we should not do such maintenance that increases complexity.

#### For dimension(group by) selector
1. columns (can shift between)
2. Level columns(AKA hierarchy) (there may be folding between levels, can not shift between)
3. [columns set](https://help.tableau.com/current/pro/desktop/en-us/sortgroup_sets_create.htm)(such as generate `case...when...` clause)

#### For filter selector, Superset is now just a simple dimension filter
1. If we want to drag saved_metric to generate a having clause,  we need to enable metic drop to filter
2. columns, or top level of hierarchy
3. Drag into different types of columns should have different filter

#### For metric selector
1. columns ( generate ad-hoc metric)
2. saved_metric

packages/superset-ui-chart-controls/src/shared-controls/dndControls.tsx
kgabryje: Done, I renamed it to DndFilterSelect

villebro (approved):
LGTM

PR #982, opened 2021-03-01
feat(core): copy feature flags from main repo

villebro:
🏆 Enhancements
This copies over the feature flag functionality from `apache/superset` to make it easier to conditionally enable features in `superset-ui`.

### DIFF
--- a/packages/superset-ui-core/test/utils/featureFlag.test.ts
+++ b/packages/superset-ui-core/test/utils/featureFlag.test.ts
@@ -16,13 +16,18 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-export { default as convertKeysToCamelCase } from './convertKeysToCamelCase';
-export { default as ensureIsArray } from './ensureIsArray';
-export { default as isDefined } from './isDefined';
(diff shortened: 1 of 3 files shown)


ktmud (approved):

packages/superset-ui-core/test/utils/featureFlag.test.ts
```diff
@@ -0,0 +1,33 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import { FeatureFlag, isFeatureEnabled } from '../../src';
+
+describe('isFeatureFlagEnabled', () => {
+  window.featureFlags = {
+    [FeatureFlag.CLIENT_CACHE]: true,
```
ktmud, line 24: Nit: flags in tests should probably be test specific because other regular features flags may be eventually cleaned up.

villebro: That's a good point, I will address this next time I'm in here

PR #985, opened 2021-03-02
feat: add BE pagination to table with pages

simcha90:
🏆 Enhancements
!Requires: https://github.com/apache/superset/pull/13417/files
This PR:
1. add BE pagination to table with support of pages
2. add setDataMask to buildQuery

📜 Documentation

https://user-images.githubusercontent.com/56388545/109660957-23ad9a00-7b72-11eb-97a7-14575bb09a82.mov


plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -88,21 +89,49 @@ export default function buildQuery(formData: TableChartFormData) {
...
+    }
+    // Because we use same buildQuery for all table on the page we need split them by id
+    formData.setCachedChanges({ [formData.slice_id]: queryObject.filters });
+
+    if (formData.server_pagination) {
+      return [
+        { ...queryObject },
+        { ...queryObject, row_limit: 0, row_offset: 0, post_processing: [], is_rowcount: true },
+      ];
+    }
```
villebro, line 121: Pretty amazing to see this in action 🚀  Now if we can only get the third object in with column totals! 🥇

plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -88,21 +89,49 @@ export default function buildQuery(formData: TableChartFormData) {
...
+    return [queryObject];
   });
 }
+
+// Use this closure to cache changing of external filters, if we have server pagination we need reset page to 0, after
+// external filter changed
+const cachedBuildQuery = () => {
+  let cachedChanges: any = {};
+  const setCachedChanges = (_cachedChanges: any) => {
+    cachedChanges = { ...cachedChanges, ..._cachedChanges };
```
villebro, line 131: could we use `newChanges` instead of `_cachedChanges`?

plugins/plugin-chart-table/src/types.ts
```diff
@@ -74,16 +74,18 @@ export interface TableChartProps extends ChartProps {
     pageSize?: number;
     currentPage?: number;
   };
+  cachedChanges: any;
+  setCachedChanges: Function;
```
villebro, line 78: Could we have the full signature here? See e.g. `SetDataMaskHook` type

villebro (commented):
First pass comments

packages/superset-ui-core/src/query/buildQueryContext.ts
```diff
@@ -24,5 +26,6 @@ export default function buildQueryContext(
     queries: buildQuery(buildQueryObject(formData, queryFields)),
     result_format: formData.result_format || 'json',
     result_type: formData.result_type || 'full',
+    setDataMask: formData.setDataMask ?? emptyFunction,
```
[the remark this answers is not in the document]
simcha90: I did it because it always same function so it give us not re-render pure components or recalc `useMemo/useCallback hooks` that get this function if all other data same

plugins/plugin-chart-table/src/buildQuery.ts
simcha90: 👍

plugins/plugin-chart-table/src/types.ts
simcha90: no need it here, removing

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -42,8 +42,7 @@ import SelectPageSize, { SelectPageSizeProps, SizeOption } from './components/Se
 import SimplePagination from './components/Pagination';
 import useSticky from './hooks/useSticky';
 import { updateExternalFormData } from './utils/externalAPIs';
-import ServerPagination from './components/ServerPagination';
-import { ServerPage } from '../types';
+import { PAGE_SIZE_OPTIONS } from '../controlPanel';
```
ktmud, line 45: When I created the `DataTable` component, the goal was that it should be re-usable outside of the table chart, which means it should not have dependencies from outside of the `DataTable` folder. All the pagination manipulation should happen in the `TableChart` component or `transformProps`. 

`DataTable` should be a generic table renderer with pagination, search, and page size options, and that's it. It can provide hooks to control which controls to display and how to render them, but it should not contain Superset chart specific logics.

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
```diff
@@ -73,17 +72,17 @@ export default function DataTable<D extends object>({
   tableClassName,
   columns,
   data,
-  currentPage = 0,
+  ownCurrentState,
```
ktmud, line 75: Do you really need to pass `ownCurrentState` on to `DataTable`?

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
simcha90: done

packages/superset-ui-core/src/query/types/Query.ts
```diff
@@ -134,6 +135,7 @@ export interface QueryContext {
   /** Response format */
   result_format: string;
   queries: QueryObject[];
+  setDataMask: SetDataMaskHook;
```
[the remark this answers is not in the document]
simcha90: done

plugins/plugin-chart-table/src/DataTable/DataTable.tsx
simcha90: fixed

villebro (commented):
This LGTM to me - I defer to @ktmud to sign off on this as he has most context with the table chart and requested some changes.

plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -43,7 +45,13 @@ export function getQueryMode(formData: TableChartFormData) {
   return hasRawColumns ? QueryMode.raw : QueryMode.aggregate;
 }
 
-export default function buildQuery(formData: TableChartFormData) {
+type Hooks = {
+  setDataMask?: SetDataMaskHook;
+  cachedChanges: any;
+  setCachedChanges: (newChanges: any) => void;
+};
```
ktmud, line 52: Should this typing live in `@superset-ui/core`?

Since you already have the`BuildQueryFunction` generic type, you can just do:

```ts
export cons buildQuery: BuildQuery<TableChartFormData>: (formData, { hooks }) => {
  // ...  hooks would be already typed
};
```

ktmud (commented):
LGTM with one suggestion on reusing typing.

plugins/plugin-chart-table/src/buildQuery.ts
simcha90: I updated now this type to `BuildQuery` type

packages/superset-ui-core/src/chart/registries/ChartBuildQueryRegistrySingleton.ts
```diff
@@ -1,7 +1,17 @@
...
-type BuildQuery = (formData: any) => QueryContext;
+export type BuildQuery<T = any> = (
+  formData: T,
+  options?: {
+    hooks?: {
+      [key: string]: any;
+      setDataMask: SetDataMaskHook;
+      cachedChanges?: any;
+      setCachedChanges: (newChanges: any) => void;
+    };
```
villebro, line 12: I think we should make this either more generic (just a generic `[key: string]: HandlerFunction) or optionally move these new hooks to the `Hooks` type: https://github.com/apache-superset/superset-ui/blob/15a718a6090668122a1b384e7335d5b9d0d22779/packages/superset-ui-core/src/chart/models/ChartProps.ts#L16-L30

villebro (approved):

packages/superset-ui-core/src/chart/registries/ChartBuildQueryRegistrySingleton.ts
```diff
@@ -1,7 +1,18 @@
...
+  options?: {
+    extras?: {
+      cachedChanges?: any;
+    };
+    hooks?: {
+      setDataMask: SetDataMaskHook;
+      setCachedChanges: (newChanges: any) => void;
+    };
+  },
+) => QueryContext;
```
ktmud, line 15: I think it's possible to reuse `BuildQueryFunction` here:

```suggestion
type BuildQuery = BuildQueryFunction<QueryFormData & JsonObject>;
```

You just provide a more relaxed `FormData` type.

PR #986, opened 2021-03-03
fix: show question mark icon when can not infer datatype

zhaoyongjie:
🐛 Bug Fix
minor fix, show question mark icon when can not infer data type in datasource panel

image

### DIFF
--- a/packages/superset-ui-chart-controls/src/components/ColumnTypeLabel.tsx
+++ b/packages/superset-ui-chart-controls/src/components/ColumnTypeLabel.tsx
@@ -23,7 +23,7 @@ export type ColumnTypeLabelProps = {
 };
 
 export function ColumnTypeLabel({ type }: ColumnTypeLabelProps) {
-  let stringIcon = '';
+  let stringIcon;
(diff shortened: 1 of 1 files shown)


ktmud:
Are we sure question mark is the icon we want to use? Adding a question mark here makes it feel like there is something wrong and it needs user attention or intervention, while in fact, they should not have to worry about it because we _should_ be able to infer all column types based on database engine specs.

villebro (approved):
I like it! @mihir174 WDYT?

villebro:
> Are we sure question mark is the icon we want to use? Adding a question mark here makes it feel like there is something wrong and it needs user attention or intervention, while in fact, they should not have to worry about it because we _should_ be able to infer all column types based on database engine specs.

I feel we should aim for being able to infer all column types - therefore having a question mark pop up when this was not the case seems like a good annoyance to provoke fixing it/opening a ticket on GH.

zhaoyongjie:
This is not an elegant solution. But it may be better than blank.

pkdotson (approved):
LGTM

PR #991, opened 2021-03-05
fix(plugin-chart-echarts): fix GraphViz crash when DnD FF enable

zhaoyongjie:
## Summary
fix graph viz crash when dnd FF enable

## use select control in Graph Viz
image

## use DnD control in Graph Viz
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Graph/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Graph/controlPanel.tsx
@@ -18,8 +18,8 @@
  */
 import React from 'react';
 import { t } from '@superset-ui/core';
-import { sections, sharedControls } from '@superset-ui/chart-controls';
-import { DEFAULT_FORM_DATA, EchartsGraphFormData } from './types';
+import { ControlPanelConfig, sections, sharedControls } from '@superset-ui/chart-controls';
+import { DEFAULT_FORM_DATA } from './types';
 import {
   legendMarginControl,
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-echarts/src/Graph/controlPanel.tsx
```diff
@@ -28,22 +28,20 @@ import {
 } from '../controls';
 
 const noopControl = { name: 'noop', config: { type: '', renderTrigger: true } };
-const columnSelectControl = {
+
+const requiredEntity = {
+  ...sharedControls.entity,
+  clearable: false,
+  // do not export private name 'filterOption'
+  filterOption: sharedControls.entity,
```
ktmud, line 36: Shouldn't `filterOption` be a function? I didn't get what the comment `do not export private name 'filterOption'` means. Are you just trying to disable the type-to-search filter?

zhaoyongjie: Hi, @ktmud
typescript raise a type error `TS4082: Default export of the module has or is using private name 'FilterOption'.`, so I guess `filterOption` is a private name.

image

zhaoyongjie: thanks for this suggestion. I have changed with this suggestion. Unfortunately, The main repo also needs some repairs.

ktmud (approved):
LGTM

PR #992, opened 2021-03-05
feat(core): add extra form data fields for native filters

villebro:
🏆 Enhancements

List of changes:
- add placeholder props `time_grain` and `time_column` fields to `QueryFormData` to get ready to streamline `granularity`, `granularity_sqla` and `time_grain_sqla`.
- fix appending of filters and adhoc filters + add tests
- remove `appendExtraFormData`, as we should only enable appending filters (are handled by `processFilters.ts` now to centralize filter handling in `buildQueryObject`)
- add new override props + add tests
- fix a few typos in `buildQueryContext.ts`

### DIFF
--- a/packages/superset-ui-core/src/query/buildQueryObject.ts
+++ b/packages/superset-ui-core/src/query/buildQueryObject.ts
@@ -1,10 +1,11 @@
 /* eslint-disable camelcase */
-import { QueryObject } from './types/Query';
+import { QueryObject, QueryObjectFilterClause } from './types/Query';
 import { QueryFieldAliases, QueryFormData } from './types/QueryFormData';
 import processFilters from './processFilters';
 import extractExtras from './extractExtras';
 import extractQueryFields from './extractQueryFields';
-import { appendExtraFormData, overrideExtraFormData } from './processExtraFormData';
+import { overrideExtraFormData } from './processExtraFormData';
+import { AdhocFilter } from './types';
 
 export const DTTM_ALIAS = '__timestamp';
 
@@ -42,8 +43,17 @@ export default function buildQueryObject<T extends QueryFormData>(
   const { metrics, columns, orderby } = extractQueryFields(residualFormData, queryFields);
 
   const extras = extractExtras(formData);
+  // collect all filters for conversion to simple filters/freeform clauses
+  const filterFormData: {
+    filters: QueryObjectFilterClause[];
+    adhoc_filters: AdhocFilter[];
+  } = {
+    filters: [...(formData.filters || []), ...(append_form_data.filters || [])],
+    adhoc_filters: [...(formData.adhoc_filters || []), ...(append_form_data.adhoc_filters || [])],
+  };
   const extrasAndfilters = processFilters({
     ...formData,
+    ...filterFormData,
     ...extras,
   });
 
@@ -68,8 +78,7 @@ export default function buildQueryObject<T extends QueryFormData>(
     url_params: url_params || undefined,
     custom_params,
   };
-  // append and override extra form data used by native filters
-  queryObject = appendExtraFormData(queryObject, append_form_data);
+  // override extra form data used by native and cross filters
   queryObject = overrideExtraFormData(queryObject, override_form_data);
   return { ...queryObject, custom_form_data };
 }
--- a/packages/superset-ui-core/src/query/buildQueryObject.ts
+++ b/packages/superset-ui-core/src/query/buildQueryObject.ts
@@ -42,9 +42,9 @@ export default function buildQueryObject<T extends QueryFormData>(
   const numericRowOffset = Number(row_offset);
   const { metrics, columns, orderby } = extractQueryFields(residualFormData, queryFields);
 
-  const extras = extractExtras(formData);
   // collect all filters for conversion to simple filters/freeform clauses
-  const { filters: extraFilters = [] } = extras;
+  const extras = extractExtras(formData);
+  const { filters: extraFilters } = extras;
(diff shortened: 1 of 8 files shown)


ktmud:
> add placeholder props time_grain and time_column fields to QueryFormData to get ready to streamline granularity, granularity_sqla and time_grain_sqla.

This is long overdue! I still have a hard time to wrap my head around every time I see `granularity`.

packages/superset-ui-core/src/query/buildQueryObject.ts
```diff
@@ -42,8 +43,17 @@ export default function buildQueryObject<T extends QueryFormData>(
   const { metrics, columns, orderby } = extractQueryFields(residualFormData, queryFields);
 
   const extras = extractExtras(formData);
+  // collect all filters for conversion to simple filters/freeform clauses
+  const filterFormData: {
+    filters: QueryObjectFilterClause[];
+    adhoc_filters: AdhocFilter[];
+  } = {
+    filters: [...(formData.filters || []), ...(append_form_data.filters || [])],
+    adhoc_filters: [...(formData.adhoc_filters || []), ...(append_form_data.adhoc_filters || [])],
```
simcha90, line 52: I think better use `??`... to do all like:
```
...(formData.filters ?? [])
```

packages/superset-ui-core/src/query/buildQueryObject.ts
```diff
@@ -42,8 +43,17 @@ export default function buildQueryObject<T extends QueryFormData>(
   const { metrics, columns, orderby } = extractQueryFields(residualFormData, queryFields);
 
   const extras = extractExtras(formData);
+  // collect all filters for conversion to simple filters/freeform clauses
+  const filterFormData: {
```
simcha90, line 47: 2 questions:
1. Should this functionality to be in `extractExtras`?
2. Should we add here also `filters and adhoc_filters` from `custom_form_data and override_form_data`?

simcha90 (approved):
approve, but some questions left

packages/superset-ui-core/src/query/buildQueryObject.ts
villebro: 1. `extractExtras` and `processFilters` probably need to be slightly refactored now that there's more stuff going on here, but since this code is slightly convoluted I prefer to keep these changes as small as possible for now. Let's do a proper refactor soon when we know the full scope of changes required by native and cross filters.
2. I don't think we should to allow overriding filters, as it could break charts, but happy to open up this discussion if there are known use cases for it. Also, `custom_form_data` probably shouldn't include `filters` and `adhoc_filters`, as those are already defined in `QueryFormData`  and `QueryObject`.

simcha90: About 2. For example what will be behavior for a case:
I have table with server pagination and cross filtering: 
1. User clicks `Page 2` we have in `custom_form_data` data about server pagination
2. We have button in dashboards like `resetDashboard` so we need reset it to `Page 1` or stuff like that

packages/superset-ui-core/test/query/buildQueryObject.test.ts
```diff
@@ -58,6 +58,41 @@ describe('buildQueryObject', () => {
...
+      extra_form_data: {
+        append_form_data: {
+          adhoc_filters: [
+            {
+              expressionType: 'SQL',
+              clause: 'WHERE',
+              sqlExpression: '(1 = 1)',
+            },
+          ],
+        },
```
ktmud, line 86: Question about `adhoc_filters` and `filters`. Is my understanding correct that `filters` are only used for native filters and can only be set from `append_form_data`, since all filters defined in chart `formData` are basically `adhoc_filters`? 

I'm a little confused why do we need to allow appending and overriding both `filters` and `adhoc_filters`. The overrides are only from a Dashboard's native/cross filters, which would be only in `QueryObjectFilterClause` or `QueryFormExtraFilter` format.

It seems to me the original design of an `adhoc_filters` vs `extra_filters` was cleaner. 

Not sure if this is the direction of future refactoring, but we probably shouldn't add extra stuff that are not configurable by chart controls to `formData`. This `extra_form_data ` could be moved to the extra options of `buildQueryObject` and `buildQuery`. E.g.

```ts
buildQueryObject(formData, {
  queryFields: ...         // configured in chart metatada,
  extraFormData: {
    append: {
      adhoc_filter: [...],
    },
    override: {
    }
  }                         // collected from dashboard state
});
```

ktmud (approved):
LGTM with some questions. Looking forward to further cleanups in this area.

packages/superset-ui-core/test/query/buildQueryObject.test.ts
villebro: AFAIK `filters` are just the simple filters that preceeded `adhoc_filters` (before my time). I'm not sure why we support both, as `adhoc_filters` are a superset of `filters`, but my assumption is it's a leftover from old times. I've been using the `filters` property when adding filters that are simple, i.e. don't require custom SQL, but we could equally well ditch `filters` and just make it possible append to `adhoc_filters`.

ktmud: We should probably:

1. Rename the Python attribute [`query_object.filter`](https://github.com/apache/superset/blob/694ae6f90e76b864d5d721c4223d36e01f7fd4cf/superset/common/query_object.py#L178) to `query_object.filters`
2. Rename `adhoc_filters` on the client side to `filters` and have it support both simple filter and adhoc filters.
3. Move the filter format consolidation logic to the server side. Either in QueryObject itself or when actually building the SQL queries.

PR #993, opened 2021-03-05
feat: echarts gauge chart

krsnik93:
🏆 Enhancements

Echarts gauge type chart in Superset.

I would appreciate feedback as the echarts chart API is really flexible, but when I look at other chart types already supported in Superset, they all expose a fairly simple API, so there are questions of just how to simplify.

For example I don't think the initial attempt would do well at imitating something like this: https://echarts.apache.org/examples/en/editor.html?c=gauge-ring.

A couple of screenshots of what's currently possible:
superset_gauge_1
superset_gauge_2

I am also a Typescript newbie so some of the typing might be weird.


villebro:
@krsnik93 this is AWESOME! ❤️  I'll post a review shortly - in the meantime feel free to annotate the code with comments if there are any areas you have questions about so we can address those properly.

eugeniamz:
I can't wait to have this in master!! @krsnik93 Thanks!!

rusackas:
While perhaps not everyone loves gauge charts, I think this is a great thing to have in our collective arsenal. They _are_ easily understandable in certain dashboards, where you want to see a benchmark or report-card KPI for some data point within known bounds and good/warning/bad thresholds. Examples:
• ISP speed tests compared to expected ranges (e.g. "am I actually GETTING gigabit?)
• A brewery's utilization of theoretical fermenter capacity, with R/A/G coloring of zones for loss/break-even/profit

Thanks for the work on this! 

Small detail, but don't forget to add a thumbnail ;)

eugeniamz:
@rusackas gouge is a key chart for KPI dashboards, having a visual image of how is the performance makes a big difference, big number says nothing about the performance. I am very excited about this chart and I can't wait to see in Master (I already said that 😃 )

junlincc:
Thank you so much @krsnik93 for your contribution! this is really awesome. ♥️

I agreed with @rusackas; Gauge is pretty common in modern BI products. It's not only great at capturing attention from the audience by providing a single quantitative value, but also indicating where the value stands within a standard range... anyway, it's an ideal chart for business users/executives, whom we haven't been focusing on. 

I would be curious to know where the contentious points are, if there are any?  Hopefully, we can address the concerns before bringing the chart to the codebase? @williaster 🙏

junlincc:
@mayurnewase would you like to review this PR? 🙏

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+      ],
+    },
+    {
+      label: t('Chart Options'),
+      expanded: true,
+      controlSetRows: [
+        [<h1 className="section-header">{t('General')}</h1>],
+        [
+          {
+            name: 'title',
```
mayurnewase, line 87: I think chart title is not required in superset context.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+          {
+            name: 'number_format',
+            config: {
+              type: 'SelectControl',
+              freeForm: true,
+              label: t('Number format'),
+              renderTrigger: true,
+              default: numberFormat,
+              choices: D3_FORMAT_OPTIONS,
+              description: `${t('D3 format syntax: https://github.com/d3/d3-format')} ${t(
```
mayurnewase, line 107: Is it required to put it like this separately?
And Second string about Label type may not be required for this chart.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+              label: t('Roundcap'),
+              renderTrigger: true,
+              default: roundcap,
+              description: t('Style the ends of the progress bar with a roundcap'),
+            },
+          },
+        ],
+        [
+          {
+            name: 'show_split_line',
```
mayurnewase, line 142: These names need to be in lower camel case.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show split lines'),
+              renderTrigger: true,
+              default: showSplitLine,
+              description: t('Show split lines'),
```
mayurnewase, line 148: Can you add more description, if you feel it's self-explanatory and is the same as the label does it need to be here?
And does echart allow to modify this value with respect to the number of split lines and angle between them?

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+        ],
+        [
+          {
+            name: 'show_axis_tick',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show axis line ticks'),
+              renderTrigger: true,
+              default: showAxisTick,
+              description: t('Show axis line ticks'),
```
mayurnewase, line 160: Same as split lines

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+        [<h1 className="section-header">{t('Intervals')}</h1>],
+        ['color_scheme'],
+        [
+          {
+            name: 'intervals',
+            config: {
+              type: 'TextControl',
+              label: t('Interval bounds'),
+              description: t(
+                'Comma-separated interval bounds, e.g. 2,4,5 for intervals 0-2, 2-4 and 4-5. Last number should match the value provided for MAX.',
```
mayurnewase, line 197: I tried adding values here,but chart partially disappeared.
Screenshot_2021-03-10  DEV  Explore - covid_vaccines(1)

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+        ],
+        [
+          {
+            name: 'value_animation',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Value animation'),
+              renderTrigger: true,
+              default: valueAnimation,
+              description: t('Animate the value or just display it'),
```
mayurnewase, line 249: I am not sure if this is required.
Does it only control value animation or arrow animation as well?
If it only for value, then I think we should give the single checkbox to enable/disable both arrow and value animation.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+              type: 'TextControl',
+              label: t('End angle'),
+              renderTrigger: true,
+              default: endAngle,
+              description: t('Angle at which to end progress axis'),
+            },
+          },
+        ],
+        [<h1 className="section-header">{t('Intervals')}</h1>],
+        ['color_scheme'],
```
mayurnewase, line 189: Changing colors is not working.
Screenshot_2021-03-10  DEV  Explore - energy_usage

plugins/plugin-chart-echarts/src/Gauge/buildQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+import { buildQueryContext, QueryFormData } from '@superset-ui/core';
+
+export default function buildQuery(formData: QueryFormData) {
+  return buildQueryContext(formData, baseQueryObject => [
+    {
+      ...baseQueryObject,
+      groupby: formData.groupby,
+      metrics: [formData.metric, formData.min_val, formData.max_val],
```
mayurnewase, line 26: If all we need is a single numeric value, is it possible to form query just for that, instead of getting all rows from the backend?

plugins/plugin-chart-echarts/src/Gauge/buildQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { buildQueryContext, QueryFormData } from '@superset-ui/core';
+
+export default function buildQuery(formData: QueryFormData) {
+  return buildQueryContext(formData, baseQueryObject => [
+    {
+      ...baseQueryObject,
+      groupby: formData.groupby,
```
mayurnewase, line 25: I think groupby here is getting deprecated for columns.
You can check graph chart's buildQuery for reference.

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,226 @@
...
+    showAxisTick,
+    startAngle,
+    endAngle,
+    intervals,
+    intervalColorIndices,
+    valueName,
+    valueAnimation,
+    valueFormatter,
+  }: EchartsGaugeFormData = { ...DEFAULT_GAUGE_FORM_DATA, ...formData };
+  const data = queriesData[0].data as TimeseriesDataRecord[];
```
mayurnewase, line 71: should we check if it's null then use empty array?

mayurnewase: Is timeSeriesRecord dataRecord?
Are we using the timestamp field ?

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,226 @@
...
+  );
+  const transformedData = data.map(data_point => ({
+    value: data_point[getMetricLabel(metric as QueryFormMetric)] as number,
+    name: valueName,
+  }));
+
+  const formatValue = (value: number) => valueFormatter.replace('{value}', numberFormatter(value));
+
+  console.log('formData via TransformProps.ts', formData);
+  console.log('queriesData', queriesData);
```
mayurnewase, line 89: Cleanup?

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,226 @@
...
+  let axisLine;
+  let axisLabel;
+  let axisTick;
+  let pointer;
+  let itemStyle;
+  let detail;
+
+  if (!intervalBoundsAndColors.length) {
+    progress = {
+      show: true,
```
mayurnewase, line 102: Can this be moved to a separate constants file and type it as well?

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,226 @@
...
+      detail,
+      data: transformedData,
+      itemStyle,
+    },
+  ];
+
+  const echartOptions: EChartsOption = {
+    grid: defaultGrid,
+    tooltip: defaultTooltip,
+    title: {
```
mayurnewase, line 214: This isn't required.

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,226 @@
...
+      },
+    };
+    pointer = {};
+    detail = {
+      valueAnimation,
+      formatter: (value: number) => formatValue(value),
+      fontSize: 1.5 * fontSize,
+    };
+    itemStyle = {};
+  } else {
```
mayurnewase, line 140: If the only difference is line and item style, should we just create a function to get it from interval bound setting?

mayurnewase (commented):
First pass review.
Having problems with reading values with the default setting, I might be doing something wrong.
Screenshot_2021-03-10  DEV  Explore - energy_usage(2)

plugins/plugin-chart-echarts/src/Gauge/buildQuery.ts
mayurnewase, line 26: Is min,max value intended to work as data filter,or just limits on chart?
If used for filter,then I think this wont work.
If only for chart,then it's not required to put it here.

plugins/plugin-chart-echarts/src/Gauge/buildQuery.ts
mayurnewase: Also how does it handle multiple rows are returned from backend for this query?

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,267 @@
...
+        ['groupby'],
+        ['metric'],
+        ['adhoc_filters'],
+        ['row_limit', null],
+        [
+          {
+            name: 'min_val',
+            config: {
+              ...sharedControls.metric,
+              type: 'MetricsControl',
```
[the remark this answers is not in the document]
zhaoyongjie: @krsnik93 Thanks for your contribution,
please do not direct uses 'MetricsControl`, when `ENABLE_EXPLORE_DRAG_AND_DROP` is enabled(from config.py), the control type will be dynamically selected.

plugins/plugin-chart-echarts/src/Gauge/buildQuery.ts
ktmud: I think `min/max` is analogous to [`x_bounds` and `y_bounds`](https://github.com/apache-superset/superset-ui/blob/645539b133e1e02b70fae8002f7a4ad6158d605e/plugins/legacy-preset-chart-nvd3/src/NVD3Controls.tsx#L230) in other charts, which should just use a `BoundsControl`.

ktmud: You probably don't even need a custom `buildQuery` if you just use a regular `metric` and `bounds` control.

plugins/plugin-chart-echarts/src/utils/controls.ts
```diff
@@ -16,10 +16,20 @@
...
+
 // eslint-disable-next-line import/prefer-default-export
 export function parseYAxisBound(bound?: string | number | null): number | undefined {
   if (bound === undefined || bound === null || Number.isNaN(Number(bound))) {
     return undefined;
   }
   return Number(bound);
 }
+
+export function tokenizeToNumericArray(value: string): number[] | null {
```
ktmud, line 30: How about this:

```ts
export function parseNumbersList(value: string, delim = ';') {
  return (value || '').split(delim).map(x => {
    const num = Number(x);
    if (Number.isNaN(num)) throw new Error('All values must be numeric');
    return num;
  });
}
```

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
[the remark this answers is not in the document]
krsnik93: I wanted to give the option of having a query result as either the start or the end of the axis. For example, user sets `min_val` as `0`, `max_val` as `sum(some_column)` and then sets `metric` as `avg(some_column)`. If you think this is overengineering, I can replace as suggested above.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: Sorry, I saw `adhoc_filters` and `row_limit`, so I assumed the custom ones should follow.

mayurnewase: oh my bad, this is correct.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: I agree on the description and will try to improve it. There is `splitNumber` in echarts which is the number of split lines and the angle is implied from it. I will add it.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: Passing interval bounds implies passing interval color indices as well. Have you added those? I can't be sure from the screenshot. Also 0 as the first one might not be needed.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: This controls only value animation but there is actually a setting for both so I will change this.

krsnik93: It does not seem to work great inside Superset so I will just drop it.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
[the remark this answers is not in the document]
ktmud: I can understand the intention but this might not be necessary even from a user-experience point of view. Having a dynamic range could be useful in some cases, but for most cases, as a dashboard consumer, it'd be confusing to see the goal changes from time to time. The purpose of a gauge chart is normally for monitoring the progress towards a certain goal; if the goal is dynamic, then it's a little bit difficult to reason whether the progress is satisfying or not.

Considering it's very easy to change the bounds and no other charts have similar capability, I'd vote for leave it to chart builders to manually manage the bounds.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,287 @@
...
+            name: 'end_angle',
+            config: {
+              type: 'TextControl',
+              label: t('End angle'),
+              description: t('Angle at which to end progress axis'),
+              renderTrigger: true,
+              default: DEFAULT_FORM_DATA.endAngle,
+            },
+          },
+        ],
```
EBoisseauSierra, line 217: Imho, [most gauges](http://www.agcoauto.com/content/images/electrical/typical_dash_gauges.jpg) are likely to be vertically symmetrical, with the “empty” at the bottom (think speedometer). Thus, we could save one parameter by only specifying something like “Gauge angle/aperture”. The input could be a [0, 360] slider (I'd say stepped every 5 or 10°).
We can then compute:
* `start_angle` = 90 + `gauge_angle` ÷ 2
* `end_angle` = 90 − `gauge_angle` ÷ 2
(apparently, there is no need for modulo)


… yet at the same time, it makes more sense to make 360°/[ring gauges](https://echarts.apache.org/examples/en/editor.html?c=gauge-ring) with the start on top.

EBoisseauSierra: P.S.

I'm here thinking how to improve UX/reduce user's cognitive load. (A proper UX designer would be more adequate to advise on this, though.)

I eventually believe one could be even more prescriptive, and hardcode start and end angle. (E.g. at 225 and −45 deg, respectively, as in [the basic example](https://echarts.apache.org/examples/en/editor.html?c=gauge-simple)).
Or these could be preset, but modifiable in a section like Explore/Customize/Advanced charts options — possibly still through one unique slider. (Off-topic: accessing this section could be linked to ad hoc privileges — i.e. to prevent “basic” users to tinker with it and decrease charts homogeneity.)

This chart type could be for < 360° gauges only, vertically symmetrical, with one unique data point visualized. Then, one could define _a separate_ chart type for [“ring gauges”](https://echarts.apache.org/examples/en/editor.html?c=gauge-ring). This type would have `startAngle` and `endAngle` equal to 90° and −270°, respectively (i.e. clockwise, starting and ending at noon) _and could display_ one or multiple _data point(s)_. Multiplying chart types isn't ideal, but I guess that in this case, it could simplify UI and actually address to not-entirely-overlapping use cases.

EBoisseauSierra:
Regarding UI, another thing to consider is which option to offer to users: [handle only](https://echarts.apache.org/examples/en/editor.html?c=gauge), [progress bar only](https://echarts.apache.org/examples/en/editor.html?c=gauge-temperature), or [both](https://echarts.apache.org/examples/en/editor.html?c=gauge-speed).
In a way, these options are orthogonal (so one tick box for each to [dis]able it)… yet if both boxes are unticked, then it doesn't make sense. So could a dropdown menu/radio buttons to select from handle/progress bar/both be better?

It's also to be noted that “progress bar only” might be quite confusing if using [stages](https://echarts.apache.org/examples/en/editor.html?c=gauge-stage) (i.e. the coloured-by-sector background). Should logic be added to disable the stages colouring if “progress bar only” is selected? Or is it fine to have stages with progress bar only?

À propos stages, it would be great to enable users to colour the gauge only _partially_ — cf. e.g. [this gauge](https://community.webcore.co/uploads/default/original/2X/6/62cc6f1f2336da7cb5c93abbff6452c7c8665637.jpg).

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,197 @@
...
+    value: data_point[getMetricLabel(metric as QueryFormMetric)] as number,
+    name: columns.map(column => `${column}: ${data_point[column]}`).join(),
+    itemStyle: {
+      color: colorFn(index),
+    },
+    title: {
+      offsetCenter: ['0%', `${index * 32 + 20}%`],
+    },
+    detail: {
+      offsetCenter: ['0%', `${index * 32 + 20 + 12}%`],
```
maloun96, line 100: what do these numbers? maybe move to constants

krsnik93:
ping

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,310 @@
...
+  optionRenderer: (c: ColumnMeta) => <ColumnOption column={c} showType />,
+  valueRenderer: (c: ColumnMeta) => <ColumnOption column={c} />,
+  valueKey: 'column_name',
+  allowAll: true,
+  filterOption: ({ data: opt }: { data: ColumnMeta }, text = '') =>
+    opt.column_name?.toLowerCase().includes(text.toLowerCase()) ||
+    opt.verbose_name?.toLowerCase().includes(text.toLowerCase()),
+  promptTextCreator: (label: string) => label,
+  commaChoosesOption: false,
+};
```
villebro, line 46: I think we can remove this and just use the `sharedControls.groupby` config (see below)

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,310 @@
...
+    {
+      label: t('Query'),
+      expanded: true,
+      controlSetRows: [
+        [
+          {
+            name: 'columns',
+            config: {
+              ...columnSelectControl,
+              clearable: false,
```
villebro, line 59: I think we should allow clearing the control (not applying groupby should be allowed IMO)
```suggestion
              ...sharedControls.groupby,
```

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,310 @@
...
+        ],
+        [
+          {
+            name: 'start_angle',
+            config: {
+              type: 'TextControl',
+              label: t('Start angle'),
+              description: t('Angle at which to start progress axis'),
+              renderTrigger: true,
+              default: DEFAULT_FORM_DATA.startAngle,
```
villebro, line 109: It might be cleaner to do object destructuring and then use those variables here:

```javascript
const { startAngle, endAngle, fontSize, ... } = DEFAULT_FORM_DATA;
```

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,197 @@
...
+  getNumberFormatter,
+  getMetricLabel,
+} from '@superset-ui/core';
+import { EChartsOption, GaugeSeriesOption } from 'echarts';
+import { GaugeDataItemOption } from 'echarts/types/src/chart/gauge/GaugeSeries';
+import { parseNumbersList } from '../utils/controls';
+import { DEFAULT_FORM_DATA as DEFAULT_GAUGE_FORM_DATA, EchartsGaugeFormData } from './types';
+
+// https://stackoverflow.com/a/10284006/2858258
+const zip = (rows: any[][]) => rows[0].map((_, c) => rows.map(row => row[c]));
```
villebro, line 34: I think this is a little difficult to read, and I'm not sure it's necessarily needed here (see below)

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,197 @@
...
+  intervalColorIndices: string,
+  colorFn: CategoricalColorScale,
+  normalizer: number,
+): Array<[number, string]> => {
+  const intervalBoundsNonNormalized = parseNumbersList(intervals, ',') || [];
+  const intervalColorIndicesArray = parseNumbersList(intervalColorIndices, ',') || [];
+  const intervalBounds = intervalBoundsNonNormalized.map(bound => bound / normalizer);
+  const intervalColors = intervalColorIndicesArray.map(ind => colorFn.colors[ind - 1]);
+  const intervalBoundsAndColors = zip([intervalBounds, intervalColors]) as Array<[number, string]>;
+  return intervalBoundsAndColors;
```
villebro, line 47: I think this works equally well:
```suggestion
  return intervalBounds.map((val, idx) => [val, intervalColors[idx]]);
```
Btw, I noticed that when typing the interval colors/bounds, if the parse isn't successful, the chart errors out and you have to refresh your browser to get it to work again. It might be a good idea to make this more resilient to errors (adding a try-catch to avoid having `parseNumbersList` throwing an error), and also perhaps default to sequentially stepping through the available colors if none are specified. So something like this may be preferable:

```javascript
  return intervalBounds.map((val, idx) => {
    const color = intervalColors[idx];
    return [val, color ?? colorFn.colors[idx]];
  });
};
```

villebro (commented):
This is looking really great! I left a few comments, please let me know if I can help get this across the finish line!

junlincc:
@villebro @mayurnewase looks like most of the comments are addressed?

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,285 @@
...
+            config: {
+              ...sharedControls.groupby,
+              label: t('Group by'),
+              description: t('Columns to group by'),
+            },
+          },
+        ],
+        ['metric'],
+        ['adhoc_filters'],
+        ['row_limit'],
```
mayurnewase, line 42: I am still not sure, if this control is required as the chart only needs 1 row.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
mayurnewase: Sorry,I asked for it because jesse suggested similar for graph pr.

villebro:
> @villebro @mayurnewase looks like most of the comments are addressed?

Thanks for the heads up @junlincc and addressing the comments @krsnik93; I'll rereview and retest today (UTC)

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
villebro: This does return multiple rows when using groupbys: 
image

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,285 @@
...
+      expanded: true,
+      controlSetRows: [
+        [<h1 className="section-header">{t('General')}</h1>],
+        [
+          {
+            name: 'min_val',
+            config: {
+              type: 'TextControl',
+              isInt: true,
+              default: DEFAULT_FORM_DATA.minVal,
```
villebro, line 56: There seems to be a bug in how number values are handled here, as it doesn't populate the default value of `0` here (probably some truthiness checking gone wrong in the control). For now it appears we need to put in something like this to make it work:
```suggestion
              default: String(DEFAULT_FORM_DATA.minVal),
```

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
mayurnewase: So chart expected to show multiple arrows?

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+  controlOverrides: {
+    columns: {
+      validators: [validateNonEmpty],
+    },
+    row_limit: {
+      choices: [...Array(10).keys()].map(n => n + 1),
+      default: DEFAULT_FORM_DATA.rowLimit,
+    },
+  },
```
villebro, line 282: Some comments on this section:
- In general I recommend avoiding the `controlOverrides` section and moving the logic directly into the control.
- `columns` is not used in this chart, so that section can be removed.

villebro (commented):
Just a few small comments, almost ready for merge 👍

junlincc:
> Regarding UI, another thing to consider is which option to offer to users: [handle only](https://echarts.apache.org/examples/en/editor.html?c=gauge), [progress bar only](https://echarts.apache.org/examples/en/editor.html?c=gauge-temperature), or [both](https://echarts.apache.org/examples/en/editor.html?c=gauge-speed).

Ideally all. Can we allow user to select from the Customize tab?

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
junlincc: what does it look like when hovering on each series?

junlincc:
in terms of default start-end degree. either so 0, 180 or -15, 195. for all variation. 
i personally like -15, 195
Screen Shot 2021-03-26 at 11 59 28 AM

EBoisseauSierra:
Well, @ikrsnik is actually pulling all the work there :rocket: !
I understand we prefer to get this merge quickly, and only then look at how to refine the UX, right?

junlincc:
@EBoisseauSierra @krsnik93 is unit test included in the PR?

krsnik93 (commented):
> @EBoisseauSierra @krsnik93 is unit test included in the PR?

I did not see unit tests for other chart types, at least for echarts ones, so I did not bother adding one. Sorry about that...

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -0,0 +1,217 @@
...
+  const data = (queriesData[0]?.data || []) as DataRecord[];
+  const numberFormatter = getNumberFormatter(numberFormat);
+  const colorFn = CategoricalColorNamespace.getScale(colorScheme as string);
+  const normalizer = maxVal;
+  const axisLineWidth = calculateAxisLineWidth(data, fontSize, overlap);
+  const axisTickLength = fontSize / 4;
+  const splitLineLength = fontSize;
+  const titleOffsetFromTitle = 2 * fontSize;
+  const detailOffsetFromTitle = 0.9 * fontSize;
+  const detailFontSize = 1.2 * fontSize;
```
mayurnewase, line 104: Can we move these(4,2,0.9) constants to constants.ts?

plugins/plugin-chart-echarts/src/Gauge/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  LegendOrientation,
+  LegendType,
+} from '../types';
+
+export type AxisTickLineStyle = {
+  width: number;
+  color: string;
+};
+
+export type EchartsGaugeFormData = EchartsLegendFormData & {
```
mayurnewase, line 31: EchartsLegendFormData may not be required,as this chart doesn't have legend.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,284 @@
...
+          },
+        ],
+        [
+          {
+            name: 'interval_color_indices',
+            config: {
+              type: 'TextControl',
+              label: t('Interval colors'),
+              description: t(
+                'Comma-separated color picks for the intervals, e.g. 1,2,4. Integers denote colors from the chosen color scheme and are 1-indexed. Length must be matching that of interval bounds.',
```
mayurnewase, line 272: Should we add a validator here to take care of these restrictions or handle those explicitly?
like, omit 0 if present,if the last number is not max then override it in transformprops.

plugins/plugin-chart-echarts/src/Gauge/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  showPointer: boolean;
+  intervals: string;
+  intervalColorIndices: string;
+  valueFormatter: string;
+};
+
+export const DEFAULT_FORM_DATA: EchartsGaugeFormData = {
+  ...DEFAULT_LEGEND_FORM_DATA,
+  legendOrientation: LegendOrientation.Top,
+  legendType: LegendType.Scroll,
```
mayurnewase, line 58: legend info may not be required.

mayurnewase (commented):
Solid work...
Can you also add tests for buildquery with some groupby columns and for transformprops with different control values?
You can check existing tests at src/plugin/plugin-chart-echarts/test

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: sounds useful but I don't see a way to fetch full form data in the validator... I would need to check the last number against maxVal, and validator only gets the field it's validating but not the rest

krsnik93: @mayurnewase do you think dropping this would be ok?

mayurnewase: yeah, I guess other than this no other comments from me.

villebro: I think we could take the modulo on the the length of the color scheme when using it, thereby looping and avoiding errors.

villebro (approved):
LGTM - just a minor comment about the color scheme, but unless there are other comments/objections I'll merge this in 24 h

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
krsnik93: Please see here:
https://github.com/apache-superset/superset-ui/pull/993/commits/07763d34cb388130ad462061bdf689dbff7f1213
I can squash all commits before merging if needed.

villebro: Thanks @krsnik93 - my bad! No need to squash, as me always squash merge PRs. Looking forward to getting this merged soon!

villebro:
Merged - thanks so much for your patience with the review process @krsnik93 ! We need to bump the core package first, but once that's done we can proceed to open up a PR on the main repo to get this in :) I'll keep you posted when we can do that (you should do the honors!)

junlincc:
Let's bump one chart at a time.

villebro:
FYI @krsnik93 - `superset-ui` has now been bumped on the main repo (https://github.com/apache/superset/pull/14045), so feel free to open a PR adding this plugin to Superset! 🙂  Let me know if you need help.

plugins/plugin-chart-echarts/src/Gauge/controlPanel.tsx
```diff
@@ -0,0 +1,284 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import React from 'react';
+import { t, validateNonEmpty, validateInteger } from '@superset-ui/core';
+import { sharedControls, ControlPanelConfig, D3_FORMAT_OPTIONS } from '@superset-ui/chart-controls';
+import { DEFAULT_FORM_DATA } from './types';
+
+const config: ControlPanelConfig = {
+  controlPanelSections: [
```
maltoze, line 25: Can we add `Time` section? I think it's very useful.
Like: https://github.com/apache-superset/superset-ui/blob/d31e6ba52ce0335cf9d4198cfc475600fc6651ca/plugins/plugin-chart-echarts/src/Pie/controlPanel.tsx#L50

EBoisseauSierra: @maltoze would you like to add your idea to [this issue](https://github.com/apache/superset/issues/14209)?

PR #994, opened 2021-03-08
fix(plugin-chart-table): ignore duplicate percent metrics

ktmud:
🐛 Bug Fix

Silently ignore duplicate percent metrics.

Currently we do this for metrics, but adding percent metrics with the same name may still cause an error. This fixes it.

### Before

duplicate-percent-metric-error

### After

duplicate-percent-metrics


ktmud:
Please ignore the bad branch name.

villebro (approved):

plugins/plugin-chart-table/src/buildQuery.ts
```diff
@@ -146,4 +146,4 @@ const cachedBuildQuery = (): BuildQuery<TableChartFormData> => {
     );
 };
 
-export default cachedBuildQuery;
+export default cachedBuildQuery();
```
ktmud, line 149: I changed how the default `buildQuery` function is exported.

cc @simcha90

PR #995, opened 2021-03-08
fix(core): don't add metrics to query object when in raw records mode

ktmud:
🏆 Enhancements
🐛 Bug Fix

Don't set `metrics` in QueryObject is it is in raw records mode. Then we can assume it is an aggregation query as long as `metrics` is defined, even if it's just empty array.

This allows a clean deprecation of the `groupby` field.

See discussions in https://github.com/apache/superset/pull/13434 for details.

### DIFF
--- a/packages/superset-ui-core/test/query/extractQueryFields.test.ts
+++ b/packages/superset-ui-core/test/query/extractQueryFields.test.ts
@@ -20,6 +20,7 @@ import extractQueryFields from '@superset-ui/core/src/query/extractQueryFields';
 import { configure } from '../../src/translation';
 import { QueryMode } from '../../src';
 import { DTTM_ALIAS } from '../../src/query/buildQueryObject';
+import { NUM_METRIC } from '../fixtures';
 
 configure();
 
@@ -28,7 +29,7 @@ describe('extractQueryFields', () => {
     expect(extractQueryFields({})).toEqual({
       columns: [],
       metrics: [],
-      orderby: [],
+      orderby: undefined,
     });
   });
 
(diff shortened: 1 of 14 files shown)


packages/superset-ui-core/src/query/types/Column.ts
```diff
@@ -41,7 +41,15 @@ export enum ColumnType {
...
-  columnName: string;
+  column_name: string;
+  groupby?: boolean;
+  is_dttm?: boolean;
+  filterable?: boolean;
+  verbose_name?: string | null;
+  description?: string | null;
+  expression?: string | null;
+  database_expression?: string | null;
+  python_date_format?: string | null;
```
ktmud, line 52: Consolidate types from `Column` and `ColumnMeta`.

packages/superset-ui-core/test/query/extractQueryFields.test.ts
```diff
@@ -28,7 +29,7 @@ describe('extractQueryFields', () => {
     expect(extractQueryFields({})).toEqual({
       columns: [],
       metrics: [],
-      orderby: [],
+      orderby: undefined,
```
kristw, line 32: Should we test when `metrics` and/or `orderby` are not set too?

kristw (approved):

packages/superset-ui-core/test/query/extractQueryFields.test.ts
ktmud: Some [existing tests](https://github.com/apache-superset/superset-ui/blob/f5c2a8d9ad5c77c06597d082a01e17b1cc009fa0/packages/superset-ui-core/test/query/extractQueryFields.test.ts#L55-L69) already covered this.

PR #998, opened 2021-03-08
refactor(chart-controls): optimize typing

ktmud:
🏠 Internal

Optimize typing for chart controls. So they can be more easily referenced in various util functions in `superset-frontend`.

See: https://github.com/apache/superset/pull/13520/files

### Test Plan

- `yarn type` should not throw error
- In `superset-frontend`, `npm run type` should not throw error with changes in https://github.com/apache/superset/pull/13520/files

### DIFF
--- a/packages/superset-ui-chart-controls/src/types.ts
+++ b/packages/superset-ui-chart-controls/src/types.ts
@@ -18,7 +18,7 @@
  * under the License.
  */
 import React, { ReactNode, ReactText, ReactElement } from 'react';
-import { QueryFormData, DatasourceType, Metric } from '@superset-ui/core';
+import { QueryFormData, DatasourceType, Metric, JsonValue } from '@superset-ui/core';
 import sharedControls from './shared-controls';
 import sharedControlComponents from './shared-controls/components';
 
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart-controls/src/types.ts
```diff
@@ -179,7 +181,7 @@ export type TabOverride = 'data' | 'customize' | boolean;
 export interface BaseControlConfig<
   T extends ControlType = ControlType,
   O extends SelectOption = SelectOption,
-  V = unknown
+  V = JsonValue
```
ktmud, line 184: It's a little complicated, but setting this allows `ControlState` to be used in [`controlUtils`](https://github.com/apache/superset/pull/13520/files#diff-e07e5c8190a73296c5a8192ac79a72cfe9eb1e93a076a4511b8879131c54a669) without a value incompatibility warning.

williaster (approved):

plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -52,7 +52,7 @@ function getQueryMode(controls: ControlStateMapping): QueryMode {
   if (mode === QueryMode.aggregate || mode === QueryMode.raw) {
     return mode as QueryMode;
   }
-  const rawColumns: QueryFormColumn[] | undefined = controls?.all_columns?.value;
+  const rawColumns = controls?.all_columns?.value as QueryFormColumn[] | undefined;
```
kristw, line 55: Is this because of linting?

kristw (approved):

plugins/plugin-chart-table/src/controlPanel.tsx
ktmud: It's because `value` now has its own type (`JsonValue`), previously it was `any` from `AnyDict`.

PR #1004, opened 2021-03-11
feat: implement Metrics dnd

kgabryje:
🏆 Enhancements

If `ENABLE_EXPLORE_DRAG_AND_DROP` is set, enables drag and drop feature for metrics control.

CC @villebro @junlincc


zhaoyongjie (approved):
LGTM

PR #1009, opened 2021-03-16
feat(plugin-chart-echarts): emit filter events when clicking on pie slice

villebro:
🏆 Enhancements
Add cross filtering to Pie Chart.
pie-filter

TODO:
- add checkbox to control panel to enable/disable emitting filters (default to disabled)
- add multiselect functionality
- add option to clear selection (button in corner?)

### DIFF
--- a/plugins/plugin-chart-echarts/src/Pie/EchartsPie.tsx
+++ b/plugins/plugin-chart-echarts/src/Pie/EchartsPie.tsx
@@ -17,9 +17,52 @@
  * under the License.
  */
 import React from 'react';
-import { EchartsProps } from '../types';
+import { PieChartTransformedProps } from './types';
 import Echart from '../components/Echart';
+import { EventHandlers } from '../types';
 
-export default function EchartsPie({ height, width, echartOptions }: EchartsProps) {
-  return <Echart height={height} width={width} echartOptions={echartOptions} />;
+export default function EchartsPie({
+  height,
+  width,
+  echartOptions,
+  setDataMask,
+  labelMap,
+  groupby,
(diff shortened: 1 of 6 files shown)


junlincc:
@mihir174 ^

plugins/plugin-chart-echarts/src/Pie/EchartsPie.tsx
```diff
@@ -16,10 +16,67 @@
...
-  return <Echart height={height} width={width} echartOptions={echartOptions} />;
+export default function EchartsPie({
+  height,
+  width,
+  echartOptions,
+  setDataMask,
+  labelMap,
+  groupby,
+}: PieChartTransformedProps) {
+  const [selectedValues, setSelectedValues] = useState<string[]>([]);
```
simcha90, line 33: I think it shouldn't be managed here, it should be saved in `currenState.value` and then injected to here from dashboard.
The question that I'm not sure if we need to pass it as `currentValue` of formData, or better pass entire currentState, so here also can be passed also other data

plugins/plugin-chart-echarts/src/Pie/EchartsPie.tsx
```diff
@@ -16,10 +16,67 @@
...
+              return {
+                col,
+                op: '==',
+                val: val as string | number | boolean,
+              };
+            }),
+          },
+        },
+        currentState: {
+          value: groupbyValues ?? null,
```
simcha90, line 60: also need to take current value from outside and set it to the chart as in `SelectFilterPlugin`

PR #1011, opened 2021-03-17
feat: support warning_markdown on metrics

etr2460:
🏆 Enhancements
Precursor to https://github.com/apache/superset/pull/13606, supports both warning_text and warning_markdown side by side in the metric option component.

Tested by npm linking to superset:
Screen Shot 2021-03-17 at 10 33 39 AM
Screen Shot 2021-03-17 at 10 33 31 AM
Screen Shot 2021-03-17 at 10 33 23 AM

### DIFF
--- a/packages/superset-ui-core/package.json
+++ b/packages/superset-ui-core/package.json
@@ -55,6 +55,7 @@
     "lodash": "^4.17.11",
     "pretty-ms": "^7.0.0",
     "react-error-boundary": "^1.2.5",
+    "react-markdown": "^4.3.1",
     "reselect": "^4.0.0",
     "rison": "^0.1.1",
     "seedrandom": "^3.0.5",
--- a/packages/superset-ui-core/src/components/SafeMarkdown.tsx
+++ b/packages/superset-ui-core/src/components/SafeMarkdown.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import ReactMarkdown, { MarkdownAbstractSyntaxTree } from 'react-markdown';
+// @ts-ignore no types available
+import htmlParser from 'react-markdown/plugins/html-parser';
+
+import { FeatureFlag, isFeatureEnabled } from '../utils';
+
+interface SafeMarkdownProps {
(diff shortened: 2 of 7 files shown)


packages/superset-ui-core/src/components/SafeMarkdown.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
etr2460, line 1: Moved from apache/superset, will delete from that codebase when I bump the superset-ui-core version used in superset

packages/superset-ui-core/package.json
```diff
@@ -55,6 +55,7 @@
     "lodash": "^4.17.11",
     "pretty-ms": "^7.0.0",
     "react-error-boundary": "^1.2.5",
+    "react-markdown": "^4.3.1",
```
ktmud, line 58: Could you update `yarn.lock` as well?

packages/superset-ui-core/src/components/SafeMarkdown.tsx
```diff
@@ -0,0 +1,34 @@
...
+import ReactMarkdown, { MarkdownAbstractSyntaxTree } from 'react-markdown';
+// @ts-ignore no types available
+import htmlParser from 'react-markdown/plugins/html-parser';
+
+import { FeatureFlag, isFeatureEnabled } from '../index';
```
ktmud, line 6:
```suggestion
import { FeatureFlag, isFeatureEnabled } from '../utils';
```

In general I think imports within a library should be more explicit to save any future trouble of circular imports.

ktmud (approved):
LGTM.

packages/superset-ui-core/package.json
etr2460: yarn.lock is git ignored for me. or do you mean the one in the root directory?

etr2460: ah, i see the yarn.lock in root is updated a lot. i'll do that

PR #1013, opened 2021-03-18
fix(generator-superset): update plugin template to 0.17 standards

villebro:
🐛 Bug Fix
When doing some viz plugin work I noticed there were some changes during the migration from `superset-ui` 0.15->0.17 that hadn't made it into the templates.

Tested locally building both regular and timeseries chart and both worked as expected.


simcha90 (approved):

zhaoyongjie (approved):
LGTM

PR #1014, opened 2021-03-19
fix(plugin-chart-echarts): fill missing values when stacked chart

villebro:
🐛 Bug Fix

Fill missing values that are next to non-null values with 0 when stacking is enabled to avoid gaps. Avoids imputing all missing values with zero to avoid overlapping lines (empirically observed that the lines became blurry due to many overlapping lines).

### AFTER
image

### BEFORE
image


junlincc:
LGTM! thanks for the fix!

jc-qa (approved):
ltgm! have to get a grey checkmark to approve lol

zhaoyongjie (approved):
LGTM

PR #1023, opened 2021-03-23
feat(viz): Pivot table chart POC

kgabryje:
This PR implements a POC of a new, enhanced pivot table chart.
Te chart is based on `react-pivottable` library (https://github.com/plotly/react-pivottable). However, it lacked several required features, so I forked the repository and implemented them myself (https://github.com/kgabryje/react-pivottable).
Features:

- Multiple visualization types (table, heatmap table, columns heatmap table, rows heatmap table
- Multiple aggregation functions (count, count unique, list unique, sum, integer sum, average, median, sample variance, sample std deviation, min, max, first, last, sum as fraction of total, sum as fraction of rows, sum as fraction of cols, count as fraction of total, count as fraction of rows, count as fraction of cols)
- custom callbacks for value cells, column headers and row headers
- customize subtotals position, grand totals visibility, sorting from "Customize" tab
- collapse column and row groups

In order to test with Superset:

1. Build `plugin-chart-pivot-table` plugin (in `superset-ui/plugins/plugin-chart-pivot-table` run `yarn`, then in `superset-ui` run `yarn && yarn build`)
2.  Link `plugin-chart-pivot-table` plugin (from `superset/superset-frontend` run `npm link path/to/superset-ui/plugins/plugin-chart-pivot-table --legacy-peer-deps`)
3. In `superset/superset-frontend/src/visualizations/presets/MainPreset.js` import the new plugin:
```
import { PivotTableChartPlugin as NewPivotTableChartPlugin } from '@superset-ui/plugin-chart-pivot-table';
```
and append it to `plugins` array:
```
new NewPivotTableChartPlugin().configure({ key: 'new_pivot_table' }),
```

image
image

CC: @villebro @junlincc


zhaoyongjie:
AWESOME new pivot table. thank you for this improvement.

could we adjust metric order use metric control?
image

zhaoyongjie:
Could we keep the function of `transpose pivot`?

image

zhaoyongjie:
### caculate sum in pivottable

image

### caculate sum in google spreadsheets
image


If we depend on the front-end calculation ground total, I think we may need more precision settings.

ktmud:
Ah, so now the pivoting happens on the client side? This may have significant performance impact in dashboards, especially those with multiple pivot tables applied to a large number of rows.

Can we do some stress testing to make sure both the downloading & rendering performance is within an acceptable criteria for dashboards with such data?

Another way to do this is probably to still do the pivoting on the backend and [flatten the pivoted table](https://stackoverflow.com/questions/39273441/flatten-pandas-pivot-table) before sending it to the client (using post-processing for both steps).

kgabryje:
@zhaoyongjie Transpose and keeping the order of metrics done (see the video). Could you please try to reproduce the miscalculation of the subtotal on one of test datasets? I've done some tests and can't reproduce and I don't have access to the dataset that you used.

https://user-images.githubusercontent.com/15073128/112300483-301ca280-8c99-11eb-8183-8145d64874fe.mov

@ktmud I've run a stress test with a pivot table of 140k cells (10k rows x 14 columns). The new pivot table renders in 13 seconds, whereas the old pivot table needs 55 seconds on my computer (and often the page crashes before rendering is done). I haven't tested it on a dashboard with multiple pivot tables, but it's hard to imagine that the new pivot table won't bring a huge improvement to what we have now.

zhaoyongjie:
I will give you this dataset offline.

villebro:
> Ah, so now the pivoting happens on the client side? This may have significant performance impact in dashboards, especially those with multiple pivot tables applied to a large number of rows.
> 
> Can we do some stress testing to make sure both the downloading & rendering performance is within an acceptable criteria for dashboards with such data?
> 
> Another way to do this is probably to still do the pivoting on the backend and [flatten the pivoted table](https://stackoverflow.com/questions/39273441/flatten-pandas-pivot-table) before sending it to the client (using post-processing for both steps).

I don't think this should have any major performance consequences, as we're doing all cell-level aggregation in the database (=a select that groups by row and column columns): this way we essentially have as many rows being sent from the backend to the frontend as there will be populated cells in the pivot table, which means that the network IO is probably mostly the same. In addition, we retain the freedom to render the table in the frontend, which IMO brings with it more pros than cons.

ktmud:
It's not about number of rows in the pivoted table, but the raw data. If we do pivoting on the client side, then all data needs to be downloaded first. The risk is at large network payloads (which can be slow to download sometimes) and data transformation for pivoting (which is CPU intensive).

Not sure which case is more often, but if the pivoting is transforming a very long table into a very short one, the old pivot table might actually renders faster---just because it doesn't have to download that much data.

ktmud:
> I don't think this should have any major performance consequences, as we're doing all cell-level aggregation in the database (=a select that groups by row and column columns): this way we essentially have as many rows being sent from the backend to the frontend as there will be populated cells in the pivot table, which means that the network IO is probably mostly the same. In addition, we retain the freedom to render the table in the frontend, which IMO brings with it more pros than cons.

Oh, this makes sense, then. I was more mostly thinking in a timeseries table there could be repetitive groups from different time points, so the data sent to the pivot function could be much larger than the output table. Looks like this will never be the case with the required groupby columns.

I'm sure the new implementation has more pros than cons, just wanna be prudent here.

kgabryje:
The issue described by @zhaoyongjie has been resolved by allowing user to choose number formatter from Customize tab (we're using d3 formatters).
https://user-images.githubusercontent.com/15073128/112489713-ff169d80-8d7e-11eb-99f4-351b3a19f8dc.mov

kgabryje:
Known bug/limitation - if there are more than 1000 columns, the "metric" header cell has a colspan that exceeds a maximum allowed value (1000). If a user scrolls past 1000th column, metrics row disappears as it cannot span over more than 1000 columns. CC @junlincc

plugins/plugin-chart-pivot-table/package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@superset-ui/chart-controls": "0.17.22",
     "@superset-ui/core": "0.17.22",
-    "@kgabryje/react-pivottable": "^0.12.2"
+    "@kgabryje/react-pivottable": "^0.12.3"
```
ktmud, line 31: @kgabryje  Can we move this under `@superset-ui` ? Just invited you on both GitHub and NPM. You should have access to fork this repo into `apache-superset` and publish npm packages under `@superset-ui`. If not, let me know!

kgabryje: Done 👍

junlincc:
> Known bug/limitation - if there are more than 1000 columns, the "metric" header cell has a colspan that exceeds a maximum allowed value (1000). If a user scrolls past 1000th column, metrics row disappears as it cannot span over more than 1000 columns. CC @junlincc

Thanks for surfacing it! no need to address this issue in this PR. super edge case plus user can always collapse to see total.

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -0,0 +1,159 @@
...
+const clickRowHeaderCallback = (
+  e: MouseEvent,
+  value: string,
+  filters: Record<string, any>,
+  pivotData: Record<string, any>,
+  isSubtotal: boolean,
+  isGrandTotal: boolean,
+) => {
+  console.log('ROW CLICKED', { e, value, filters, pivotData, isSubtotal, isGrandTotal });
+};
```
villebro, line 65: Maybe we could comment out the `console.log`s here until we implement filters/other events.

plugins/plugin-chart-pivot-table/src/plugin/controlPanel.ts
```diff
@@ -0,0 +1,250 @@
...
+            config: {
+              type: 'SelectControl',
+              label: t('Cols sort by'),
+              default: 'key_a_to_z',
+              choices: [
+                // [value, label]
+                ['key_a_to_z', t('key alphabetically')],
+                ['value_a_to_z', t('value ascending')],
+                ['value_z_to_a', t('value descending')],
+              ],
```
villebro, line 185: Should we also support `key_z_to_a`?

plugins/plugin-chart-pivot-table/src/plugin/index.ts
```diff
@@ -0,0 +1,52 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import { t, ChartMetadata, ChartPlugin } from '@superset-ui/core';
+import buildQuery from './buildQuery';
+import controlPanel from './controlPanel';
+import transformProps from './transformProps';
+import thumbnail from '../images/thumbnail.png';
```
villebro, line 23: We should add a proper thumbnail instead of the placeholder. I'm ok using the one from the legacy plugin, as it's mostly indistinguishable from this one at that resolution.

villebro (commented):
A few comments

junlincc:
@kgabryje Hey Kamil, let's address the last batch of comments, get all the checks passed and bring this PR to the finish line. after your break oc. 🙂

plugins/plugin-chart-pivot-table/src/plugin/index.ts
kgabryje: Done. I used the one from the old pivot table

plugins/plugin-chart-pivot-table/src/plugin/controlPanel.ts
kgabryje: Done. It's not supported in the original `react-pivottable`, I implemented it in our fork `@superset-ui/react-pivottable`

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
kgabryje: Done. I replaced `console.log`s with `TODO` comments

kgabryje:
@junlincc All of the comments have been addressed, the PR is ready for the final review :)

junlincc:
@kgabryje just can't say it enough, you are awesome! 
If the code looks good, let's get this PR to the mainrepo for testing. I believe we are not doing the db migration yet have we?

villebro (approved):
LGTM, can't wait to get this in! @junlincc - we need to come up with a plan for migrating to this new plugin. The difficulty is that migrations currently can't be placed behind a feature flag, and since this version is incompatible with the old plugin's metadata, we can't migrate these for only a subset of users. I propose we introduce a new feature flag "VIZ_PLUGIN_BLEEDING_ENGE" that adds new experimental charts to those that have it enabled, but hides it from the rest. Once a plugin has been found to be stable, we can then do a migration that changes all existing charts to the new one.

junlincc:
instead of putting them under a FF, can we have both Pivot tables available until the new one is fully migrated? @villebro

plugins/plugin-chart-pivot-table/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Usage
+
+Configure `key`, which can be any `string`, and register the plugin. This `key` will be used to
+lookup this chart throughout the app.
+
+```js
+import PivotTableChartPlugin from '@superset-ui/plugin-chart-pivot-table';
+
+new PivotTableChartPlugin().configure({ key: 'pivot-table' }).register();
```
rusackas, line 15:
```suggestion
new PivotTableChartPlugin().configure({ key: 'pivot-table-v2' }).register();
```

plugins/plugin-chart-pivot-table/README.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Then use it via `SuperChart`. See
+[storybook](https://apache-superset.github.io/superset-ui/?selectedKind=plugin-chart-pivot-table)
+for more details.
+
+```js
+<SuperChart
+  chartType="pivot-table"
```
rusackas, line 24:
```suggestion
  chartType="pivot-table-v2"
```
There might be some other places to catch this as well, but let's give it a unique key so it can be loaded concurrently with the other one.

rusackas (changes requested):
LGTM, but in order to load it concurrently with the existing one, we should give it a unique key and refer to it everywhere.

plugins/plugin-chart-pivot-table/README.md
kgabryje: Done

plugins/plugin-chart-pivot-table/README.md
kgabryje: Done

PR #1028, opened 2021-03-27
feat(plugin-chart-table): add small number formatter

ktmud:
🐛 Bug Fix

Realized I didn't commit the changes about small number formatter before merging... https://github.com/apache-superset/superset-ui/pull/1019

### DIFF
--- a/plugins/plugin-chart-table/src/utils/DateWithFormatter.ts
+++ b/plugins/plugin-chart-table/src/utils/DateWithFormatter.ts
@@ -43,12 +43,11 @@ export default class DateWithFormatter extends Date {
 
     this.input = input;
     this.formatter = formatter;
-  }
-
-  toString(): string {
-    if (this.formatter === String) {
-      return String(this.input);
-    }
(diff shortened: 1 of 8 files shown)


plugins/plugin-chart-table/src/utils/DateWithFormatter.ts
```diff
@@ -43,12 +43,11 @@ export default class DateWithFormatter extends Date {
 
     this.input = input;
     this.formatter = formatter;
-  }
-
-  toString(): string {
-    if (this.formatter === String) {
-      return String(this.input);
-    }
-    return this.formatter ? this.formatter(this) : Date.toString.call(this);
```
ktmud, line 52: While this code works in production, somehow it doesn't work in the dev mode of Storybook. Moving to instance in constructor to make sure it works in both cases.

ktmud:
@kristw do you mind stamping this as well?

etr2460 (approved):
lgtm!

PR #1033, opened 2021-03-29
feat: add default series_limit

etr2460:
💔 Breaking Changes
🏆 Enhancements
Makes the change recommended in https://github.com/apache/superset/pull/13802 to help protect performance when grouping by high cardinality dimensions in the default case

Test Plan:
CI

to: @ktmud @michael-s-molina


ktmud (approved):
Sounds like a reasonable change.

ktmud:
10 may have data accuracy issue if people didn't notice the default is set to 10; 100 sounds about right because it's already very hard to check whether some series is missing among the 100.

etr2460:
My thoughts exactly. The row limit is set to 10k (which is crazy large too) so I figured this should err on the larger side too.

PR #1035, opened 2021-03-30
fix(chart-controls): export/import DTTM_ALIAS properly

ktmud:
🐛 Bug Fix

Fixes a bad import introduced by https://github.com/apache-superset/superset-ui/pull/1019

See https://github.com/apache-superset/superset-ui/pull/1034#discussion_r603935519 for details


etr2460 (approved):
nice, lgtm

PR #1036, opened 2021-04-01
fix(chart-controls): transform emotion css prop

ktmud:
🐛 Bug Fix

Emotion's [css prop](https://emotion.sh/docs/css-prop#babel-preset) must be transpiled to use emotion's `jsx` function. Not doing this is will result in missing styles when the built React components (`plugin/lib/../Component.js`) are used in `superset-frontend`. This affect only production builds, and not `npm link` because `npm link` uses `plugin/src/../Components.tsx`, which can be picked up by `superset-frontend`'s Babel emotion plugin.

Tested by temporarily disable `src` replacement logic in `superset-frontend`'s `webpack.config.js`.

Related: https://github.com/apache/superset/pull/13758#issuecomment-810604105

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -20,4 +20,11 @@ config.plugins = [
   ['babel-plugin-typescript-to-proptypes', { loose: true }],
   ['@babel/plugin-proposal-class-properties', { loose: true }],
 ];
+config.presets.push([
+  '@emotion/babel-preset-css-prop',
+  {
+    autoLabel: 'dev-only',
+    labelFormat: '[local]',
+  },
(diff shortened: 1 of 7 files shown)


kristw (approved):

kristw:
Ah wait. The code looks like it makes sense but the build is failing.

ktmud:
> sense

Just pushed a fix.  (Hope it works!)

ktmud:
> Ah wait. The code looks like it makes sense but the build is failing.

Just pushed a fix. (Hope it works!)

PR #1038, opened 2021-04-04
feat(native-filters): add AppSection type

simcha90:
🏆 Enhancements
Add `AppSection` metadata to support functionality in https://github.com/apache/superset/pull/13726#pullrequestreview-620185791
Changes:
1. Added AppSection type that specify place of chart on the screen (it's parent component)

### DIFF
--- a/packages/superset-ui-core/src/chart/models/ChartProps.ts
+++ b/packages/superset-ui-core/src/chart/models/ChartProps.ts
@@ -55,7 +55,7 @@ export interface ChartPropsConfig {
   ownCurrentState?: JsonObject;
   /** Set of actual behaviors that this instance of chart should use */
   behaviors?: Behavior[];
-  /** Place of the chart on the screen */
+  /** Application section of the chart on the screen (in what components/screen it placed) */
   appSection?: AppSection;
 }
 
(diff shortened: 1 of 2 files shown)


packages/superset-ui-core/src/chart/models/ChartProps.ts
```diff
@@ -1,5 +1,5 @@
 import { createSelector } from 'reselect';
-import { Behavior, convertKeysToCamelCase, Datasource, JsonObject } from '../..';
+import { Behavior, convertKeysToCamelCase, Datasource, JsonObject, AppSection } from '../..';
```
villebro, line 2: nit: ABC

villebro (commented):
LGTM - please update the title and description to reflect the renamed property

packages/superset-ui-core/src/chart/types/Base.ts
```diff
@@ -7,6 +7,13 @@ export enum Behavior {
   NATIVE_FILTER = 'NATIVE_FILTER',
 }
 
+export enum AppSection {
+  EXPLORE = 'EXPLORE',
+  DASHBOARD = 'DASHBOARD',
+  FILTER_BAR = 'FILTER_BAR',
+  FILTER_CONFIG_MODAL = 'FILTER_CONFIG_MODAL',
```
villebro, line 14: ..might also be a good idea to add `EMBEDDED` as an option

villebro (approved):
LGTM - leaving open for a while for others to chime in.

PR #1041, opened 2021-04-06
feat(plugin-chart-table): enable emitting cross-filters

kgabryje:
This PR enables emitting cross-filters from Table chart.
To enable the feature, set the feature flag `DASHBOARD_CROSS_FILTERS` in `superset/config.py` and select "ENABLE EMITTING FILTERS" checkbox in Customize tab.
To select filter values, click on cells in groupby columns. Selecting a value emits a filter with an `IN` operator.

https://user-images.githubusercontent.com/15073128/113731815-5c084100-96f9-11eb-92c5-bdfb969ae9e9.mov
image

CC @junlincc @villebro @ktmud

### DIFF
--- a/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/plugins/plugin-chart-table/src/TableChart.tsx
@@ -161,12 +161,50 @@ export default function TableChart<D extends DataRecord = DataRecord>(
     showCellBars = true,
     emitFilter = false,
     sortDesc = false,
-    onChangeFilter,
-    filters: initialFilters,
+    filters: initialFilters = {},
     sticky = true, // whether to use sticky header
   } = props;
 
   const [filters, setFilters] = useState(initialFilters);
+
+  const handleChange = useCallback(
+    (filters: { [x: string]: DataRecordValue[] }) => {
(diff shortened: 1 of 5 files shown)


plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -161,13 +161,51 @@ export default function TableChart<D extends DataRecord = DataRecord>(
...
+      if (!emitFilter) {
+        return;
+      }
+
+      const groupBy = Object.keys(filters);
+      const groupByValues = Object.values(filters);
+      setDataMask({
+        crossFilters: {
+          extraFormData: {
+            append_form_data: {
```
ktmud, line 181: I'm wondering whether there is room for a simpler public API to set cross filters. Now It feels a little tedious that every chart with corss-filter support has to set this very long nested object themselves.

villebro: We're actually currently cleaning up the API (some of this is leftovers from during development when it wasn't clear in which direction this was going to ultimately go)

villebro (approved):
Tested, works like a charm!

PR #1043, opened 2021-04-07
fix(plugin-chart-echarts): default to standard x-axis format

villebro:
🏆 Enhancements
Default to using the default time axis formatter for the ECharts Timeseries chart and remove default 45 degree rotation. Also introduce a new adaptive time formatter for use in the tooltip which shows increasing level of detail based on the timestamp. This is in contrast with the current adaptive time formatter, which e.g. only shows the seconds/milliseconds when a timestamp has second/subsecond granularity.

Last full year (only months show up):
image

Last quarter (months with weekly ticks):
image

Last month (four day ticks):
image

Last week (daily ticks):
image

Two days (6 hour ticks):
image

Tooltip with millisecond granularity data:
image


packages/superset-ui-chart-controls/src/utils/D3Formatting.ts
```diff
@@ -46,7 +46,7 @@ export const D3_TIME_FORMAT_OPTIONS: [string, string][] = [
   ['%m/%d/%Y', '%m/%d/%Y | 01/14/2019'],
   ['%Y-%m-%d', '%Y-%m-%d | 2019-01-14'],
   ['%Y-%m-%d %H:%M:%S', '%Y-%m-%d %H:%M:%S | 2019-01-14 01:32:10'],
-  ['%d-%m-%Y %H:%M:%S', '%Y-%m-%d %H:%M:%S | 14-01-2019 01:32:10'],
+  ['%d-%m-%Y %H:%M:%S', '%d-%m-%Y %H:%M:%S | 14-01-2019 01:32:10'],
```
villebro, line 49: Bycatch - typo in the description

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -184,7 +193,7 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
         const value: number = !richTooltip ? params.value : params[0].value[0];
         const prophetValue = !richTooltip ? [params] : params;
 
-        const rows: Array<string> = [`${xAxisFormatter(value)}`];
+        const rows: Array<string> = [`${tooltipFormatter(value)}`];
```
zhaoyongjie, line 196: I tested and found that the `String` constructor did not work.

image

zhaoyongjie (approved):
LGTM, Ignore my previous reply.

PR #1044, opened 2021-04-07
fix(plugin-chart-echarts): fix unnecessary chart clearing

villebro:
🐛 Bug Fix
This removes flicker in the ECharts charts introduced in #1010 caused by calling `clear()` on the ECharts instance. This was previously done to clear old highlighted values, as old highlights have to be explicitly "downplayed". This PR removes the `clear()` call, and to downplay non-selected values, the previous selection is compared to the current selection; if a previously selected value is no longer present, it is"downplayed".

### BEFORE
Previously, the chart would rerender every time the `useEffect` hook was called:
bloxplot-before

### AFTER
After the PR, no flashing occurs:
boxplot-after

### DIFF
--- a/plugins/plugin-chart-echarts/src/components/Echart.tsx
+++ b/plugins/plugin-chart-echarts/src/components/Echart.tsx
@@ -26,14 +26,6 @@ const Styles = styled.div<EchartsStylesProps>`
   width: ${({ width }) => width};
 `;
 
-const usePrevious = <T extends unknown>(value: T): T | undefined => {
-  const ref = useRef<T>();
-  useEffect(() => {
-    ref.current = value;
-  });
-  return ref.current;
-};
-
 export default function Echart({
   width,
   height,
@@ -44,7 +36,7 @@ export default function Echart({
   const divRef = useRef<HTMLDivElement>(null);
   const chartRef = useRef<ECharts>();
   const currentSelection = Object.keys(selectedValues) || [];
-  const previousSelection = usePrevious<string[]>(currentSelection) || [];
+  const previousSelection = useRef<string[]>([]);
 
   useEffect(() => {
     if (!divRef.current) return;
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-echarts/src/components/Echart.tsx
```diff
@@ -26,38 +26,54 @@ const Styles = styled.div<EchartsStylesProps>`
   width: ${({ width }) => width};
 `;
 
+const usePrevious = <T extends unknown>(value: T): T | undefined => {
```
ktmud, line 29:
```suggestion
const usePrevious = <T>(value: T): T | undefined => {
```

I think this is good enough?

plugins/plugin-chart-echarts/src/components/Echart.tsx
```diff
@@ -26,38 +26,54 @@ const Styles = styled.div<EchartsStylesProps>`
...
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  });
```
ktmud, line 33: Wouldn't this only store the initial value?

plugins/plugin-chart-echarts/src/components/Echart.tsx
```diff
@@ -26,38 +26,54 @@ const Styles = styled.div<EchartsStylesProps>`
...
   height,
   echartOptions,
   eventHandlers,
-  selectedValues,
+  selectedValues = {},
 }: EchartsProps) {
   const divRef = useRef<HTMLDivElement>(null);
   const chartRef = useRef<ECharts>();
+  const currentSelection = Object.keys(selectedValues) || [];
+  const previousSelection = usePrevious<string[]>(currentSelection) || [];
```
ktmud, line 47: I think `usePrevious` can already infer types from `currentSelection`?

ktmud: Can we maybe just always downplay all previous highlighted selection and highlight all current selection?

```ts
const previousSelection = useRef(currentSelection);

useEffect(() => {
  if (previousSelection.current.length) {
    // downplay
  }
  if (currentSelection.length) {
    // highlight
  }
  previousSelection.current = currentSelection;
});
```

ktmud (commented):
Thanks for the fix! I'm not sure whether `usePrevious` is necessary is in this case. I often find directly using `useRef` to be more readable.

plugins/plugin-chart-echarts/src/components/Echart.tsx
villebro: I initially did that, but as I didn't know how expensive the operation is, so I erred on the side of caution. But I'll change it and if it does turn out to become a bottleneck.

plugins/plugin-chart-echarts/src/components/Echart.tsx
villebro: I picked up this from StackOverflow and didn't look too closely at the details - I pulled this out and replaced with a simple `useRef` instead

plugins/plugin-chart-echarts/src/components/Echart.tsx
villebro: I changed it so that we highlight all currently selected values, but only downplay the previously selected values that aren't currently selected to avoid flipflopping.

zhaoyongjie (approved):
LGTM!!

### before in dashboard auto-refresh
https://user-images.githubusercontent.com/2016594/114030285-efce3e80-98ac-11eb-8370-4ff1dfd91cfc.mov

### after in dashboard auto-refresh

https://user-images.githubusercontent.com/2016594/114030437-06749580-98ad-11eb-9791-904d7a54436d.mov

ktmud (approved):

PR #1046, opened 2021-04-08
fix(plugin-chart-echarts): add series deduplication

villebro:
🐛 Bug Fix
If an annotation layer has a series name that already exists in the chart, the chart raises an error. This PR adds series name deduplication to ensure the chart can be drawn.

### AFTER
image

### BEFORE
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/utils/series.ts
+++ b/plugins/plugin-chart-echarts/src/utils/series.ts
@@ -184,3 +184,23 @@ export function getChartPadding(
     bottom: bottom + (orientation === LegendOrientation.Bottom ? legendMargin : 0),
   };
 }
+
+export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
+  const counter: Record<string, number> = {};
+  return series.map(row => {
+    const { id } = row;
+    if (id === undefined) return row;
+    const count = counter[id as string];
+    if (count === undefined) {
+      counter[id] = 1;
+    } else {
+      counter[id] += 1;
+    }
+    const prefix = count ? ` (${count})` : '';
+
+    return {
+      ...row,
+      id: `${id}${prefix}`,
+    };
+  });
+}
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -184,3 +184,23 @@ export function getChartPadding(
     bottom: bottom + (orientation === LegendOrientation.Bottom ? legendMargin : 0),
   };
 }
+
+export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
```
zhaoyongjie, line 188: how about this

```suggestion
export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
  const counter = new Map<string, number>();
  return series.map(row => {
    let { id } = row;
    if (id === undefined) return row;

    id = String(id);
    if (counter.has(id)) {
      counter.set(id, counter.get(id) as number + 1);
    } else {
      counter.set(id, 0);
    }
    const suffix = counter.get(id) as number > 0 ? ` (${counter.get(id)})` : '';

    return {
      ...row,
      id: `${id}${suffix}`,
    };
  });
}
```

plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -184,3 +184,23 @@ export function getChartPadding(
...
+      counter[id] += 1;
+    }
+    const prefix = count ? ` (${count})` : '';
+
+    return {
+      ...row,
+      id: `${id}${prefix}`,
+    };
+  });
+}
```
zhaoyongjie, line 206: How about this?

```suggestion
export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
  const counter = new Map<string, number>();
  return series.map(row => {
    let { id } = row;
    if (id === undefined) return row;

    id = String(id);
    if (counter.has(id)) {
      counter.set(id, counter.get(id) as number + 1);
    } else {
      counter.set(id, 0);
    }
    const suffix = counter.get(id) as number > 0 ? ` (${counter.get(id)})` : '';

    return {
      ...row,
      id: `${id}${suffix}`,
    };
  });
}
```

zhaoyongjie:
Thank you for this fix, tested in my local env. 

image

zhaoyongjie (approved):
LGTM!

PR #1047, opened 2021-04-08
fix(plugin-chart-table): fix empty metrics

villebro:
🐛 Bug Fix
The totals query should not be added if there are no metrics present. The check needs to verify that 1) `metrics` is not `undefined`/`null` and 2) it's `length` is truthy.

### AFTER
image

# BEFORE
image


simcha90 (approved):

zhaoyongjie (approved):
LGTM

PR #1049, opened 2021-04-09
fix(legacy-preset-chart-nvd3): fix tokenizer input type

villebro:
🐛 Bug Fix
In the bullet chart, the value for ranges can be undefined, causing the chart to break if no ranges are specified.

### AFTER
image

### BEFORE
image


zhaoyongjie (approved):
LGTM, thanks for this fix!

PR #1051, opened 2021-04-13
chore(plugin-chart-table): Add muted attribute of video tag to whitelist

maltoze:
🏆 Enhancements

Add `muted` attribute of `video` tag to whitelist.

The `muted` is very useful because `autoplay` attribute will not work if not added `muted` on Chrome.

Default: https://github.com/leizongmin/js-xss/blob/master/lib/default.js#L75
```js
video: ["autoplay", "controls", "loop", "preload", "src", "height", "width"]
```


villebro (approved):
Thanks for the contribution @maltoze !

PR #1052, opened 2021-04-13
fix(plugin-chart-echarts): fix unnecessary highlight

maltoze:
🐛 Bug Fix
Currently, `Timeseries` chart highlighted by default.
This will fix it.

## BEFORE
image

## AFTER
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/components/Echart.tsx
+++ b/plugins/plugin-chart-echarts/src/components/Echart.tsx
@@ -55,7 +55,7 @@ export default function Echart({
       type: 'downplay',
       dataIndex: previousSelection.current.filter(value => !currentSelection.includes(value)),
     });
-    if (currentSelection.length > 0) {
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-echarts/src/components/Echart.tsx
```diff
@@ -55,10 +55,12 @@ export default function Echart({
       type: 'downplay',
       dataIndex: previousSelection.current.filter(value => !currentSelection.includes(value)),
     });
-    chartRef.current.dispatchAction({
-      type: 'highlight',
-      dataIndex: currentSelection,
-    });
+    if (currentSelection.length > 0) {
```
villebro, line 58: Thanks for the fix! Can we simplify this a bit?
```suggestion
    if (currentSelection.length) {
```

villebro (approved):

PR #1053, opened 2021-04-13
chore(core): simplify ExtraFormData

villebro:
💔 Breaking Changes
This is a follow-up to #1040 to further simplify the data mask hook and related types.

Before the `ExtraFormData` contained three properties

```
extraFormData = {
  append_form_data: {
    adhoc_filters: [],
    filters: [],
  },
  override_form_data: {
    time_range: '',
  },
  custom_form_data: {},
}
```

After this change, `append_form_data`, `override_form_data` and `custom_form_data`  are all in the same object, with the implicit assumption that filters and `custom_form_data` are appended to fields with the same name in `QueryFormData`/`QueryObject`, and the rest (mostly temporal filters, like time column, time range, time grain etc) are overridden. Example:

```
extraFormData = {
  adhoc_filters: [],
  filters: [],
  time_range: '',
  custom_form_data: {},
}
```

In addition, `Behavior.CROSS_FILTER` is renamed to `Behavior.INTERACTIVE_CHART` to more clearly communicate that it applies to any interactive chart behavior as opposed to native filters.

The details will be fully documented shortly as we stabilize these interfaces.

### DIFF
--- a/packages/superset-ui-core/src/query/buildQueryObject.ts
+++ b/packages/superset-ui-core/src/query/buildQueryObject.ts
@@ -20,7 +20,7 @@ export default function buildQueryObject<T extends QueryFormData>(
 ): QueryObject {
   const {
     annotation_layers = [],
-    extra_form_data = {},
+    extra_form_data,
     time_range,
     since,
     until,
@@ -34,7 +34,12 @@ export default function buildQueryObject<T extends QueryFormData>(
     custom_params = {},
     ...residualFormData
   } = formData;
-  const { append_form_data = {}, override_form_data = {}, custom_form_data = {} } = extra_form_data;
+  const {
+    adhoc_filters: appendAdhocFilters = [],
+    filters: appendFilters = [],
+    custom_form_data = {},
+    ...overrides
+  } = extra_form_data || {};
 
   const numericRowLimit = Number(row_limit);
   const numericRowOffset = Number(row_offset);
@@ -43,7 +48,6 @@ export default function buildQueryObject<T extends QueryFormData>(
   // collect all filters for conversion to simple filters/freeform clauses
   const extras = extractExtras(formData);
   const { filters: extraFilters } = extras;
-  const { adhoc_filters: appendAdhocFilters = [], filters: appendFilters = [] } = append_form_data;
   const filterFormData: {
     filters: QueryObjectFilterClause[];
     adhoc_filters: AdhocFilter[];
@@ -79,7 +83,7 @@ export default function buildQueryObject<T extends QueryFormData>(
     custom_params,
   };
   // override extra form data used by native and cross filters
-  queryObject = overrideExtraFormData(queryObject, override_form_data);
+  queryObject = overrideExtraFormData(queryObject, overrides);
 
   return { ...queryObject, custom_form_data };
 }
--- a/packages/superset-ui-core/src/query/constants.ts
+++ b/packages/superset-ui-core/src/query/constants.ts
@@ -26,25 +26,12 @@ import {
  */
 export const DTTM_ALIAS = '__timestamp';
 
-export const EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS: (keyof ExtraFormDataOverrideRegular)[] = [
-  'granularity',
-  'granularity_sqla',
-  'time_column',
-  'time_grain',
-  'time_range',
-];
-
 export const EXTRA_FORM_DATA_OVERRIDE_EXTRA_KEYS: (keyof ExtraFormDataOverrideExtras)[] = [
   'druid_time_origin',
   'relative_start',
   'relative_end',
   'time_grain_sqla',
   'time_range_endpoints',
-];
-
-export const EXTRA_FORM_DATA_OVERRIDE_KEYS: (keyof ExtraFormDataOverride)[] = [
-  ...EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS,
-  ...EXTRA_FORM_DATA_OVERRIDE_EXTRA_KEYS,
 ];
 
 export const EXTRA_FORM_DATA_APPEND_KEYS: (keyof ExtraFormDataAppend)[] = [
@@ -66,3 +53,12 @@ export const EXTRA_FORM_DATA_OVERRIDE_REGULAR_MAPPINGS: Record<
   time_grain: 'time_grain',
   time_range: 'time_range',
 };
+
+export const EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS = Object.keys(
+  EXTRA_FORM_DATA_OVERRIDE_REGULAR_MAPPINGS,
+) as (keyof ExtraFormDataOverrideRegular)[];
+
+export const EXTRA_FORM_DATA_OVERRIDE_KEYS: (keyof ExtraFormDataOverride)[] = [
+  ...EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS,
+  ...EXTRA_FORM_DATA_OVERRIDE_EXTRA_KEYS,
+];
--- a/packages/superset-ui-core/src/query/constants.ts
+++ b/packages/superset-ui-core/src/query/constants.ts
@@ -1,3 +1,11 @@
+import {
+  ExtraFormDataAppend,
+  ExtraFormDataOverrideExtras,
+  ExtraFormDataOverrideRegular,
+  ExtraFormDataOverride,
+  QueryObject,
+} from './types';
+
 /**
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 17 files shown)


packages/superset-ui-core/src/query/buildQueryObject.ts
```diff
@@ -20,7 +20,7 @@ export default function buildQueryObject<T extends QueryFormData>(
 ): QueryObject {
   const {
     annotation_layers = [],
-    extra_form_data = {},
```
simcha90, line 23: why not leave it?

villebro: I was getting a weird TS error - it was moved down to line 42 in the form of `extra_form_data || {}`.

packages/superset-ui-core/src/query/constants.ts
```diff
@@ -17,3 +25,44 @@
...
+  ...EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS,
+  ...EXTRA_FORM_DATA_OVERRIDE_EXTRA_KEYS,
+];
+
+export const EXTRA_FORM_DATA_APPEND_KEYS: (keyof ExtraFormDataAppend)[] = [
+  'adhoc_filters',
+  'filters',
+  'cross_groupby',
+  'cross_highlight',
+  'cross_drilldown',
```
simcha90, line 55: what is it property `cross_drilldown`? why we all these properties have prefix `cross`? may be also native filters can do same thing?

packages/superset-ui-core/src/query/constants.ts
```diff
@@ -17,3 +25,44 @@
...
  */
 export const DTTM_ALIAS = '__timestamp';
+
+export const EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS: (keyof ExtraFormDataOverrideRegular)[] = [
+  'granularity',
+  'granularity_sqla',
+  'time_column',
+  'time_grain',
+  'time_range',
+];
```
simcha90, line 35: better do it as I think as:
```
export const EXTRA_FORM_DATA_OVERRIDE_REGULAR_KEYS: (keyof ExtraFormDataOverrideRegular)[] = Object.values(EXTRA_FORM_DATA_OVERRIDE_REGULAR_MAPPINGS);
```

packages/superset-ui-core/src/query/processExtraFormData.ts
```diff
@@ -16,39 +16,29 @@
...
-  overrideFormData: Partial<QueryObject>,
+  overrideFormData: ExtraFormDataOverride,
 ): QueryObject {
   const overriddenFormData: QueryObject = { ...queryObject };
   const { extras: overriddenExtras = {} } = overriddenFormData;
-  Object.entries(ALLOWED_OVERRIDES).forEach(([key, target]) => {
-    if (key in overrideFormData) {
-      overriddenFormData[target] = overrideFormData[key];
+  Object.entries(EXTRA_FORM_DATA_OVERRIDE_REGULAR_MAPPINGS).forEach(([key, target]) => {
+    // @ts-ignore
```
simcha90, line 32: why ignore?

packages/superset-ui-core/src/query/processExtraFormData.ts
```diff
@@ -16,39 +16,29 @@
...
+    if (value !== undefined) {
+      overriddenFormData[target] = value;
     }
   });
-  const { extras: overrideExtras = {} } = overrideFormData;
-  ALLOWED_EXTRAS_OVERRIDES.forEach(key => {
-    if (key in overrideExtras) {
+  EXTRA_FORM_DATA_OVERRIDE_EXTRA_KEYS.forEach(key => {
+    if (key in overrideFormData) {
       // @ts-ignore
```
simcha90, line 40: and also why here? :)

packages/superset-ui-core/test/query/processExtraFormData.test.ts
```diff
@@ -70,6 +70,7 @@ describe('overrideExtraFormData', () => {
           time_range: '100 years ago',
         },
         {
+          // @ts-expect-error
```
simcha90, line 73: what this mean?

plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -177,24 +177,22 @@ export default function TableChart<D extends DataRecord = DataRecord>(
...
-                ? []
-                : groupBy.map(col => {
-                    const val = filters?.[col];
-                    if (val === null || val === undefined)
-                      return {
-                        col,
-                        op: 'IS NULL',
-                      };
+          filters:
+            groupBy.length === 0
```
simcha90, line 181: looks like EchartsPie data, may be we can create common utils if it relevant also for other charts?

simcha90 (approved):
Left some comments

packages/superset-ui-core/src/query/constants.ts
villebro: We discussed it with @amitmiran137 - we just needed to come up with a prefix that clearly distinguishes these from regular form data fields, and `cross` seemed like the most descriptive one we could come up with. `cross_drilldown` is a placeholder for the forthcoming drilldown action.

packages/superset-ui-core/src/query/constants.ts
villebro: Good idea 👍  I assume you mean `Object.keys`.

packages/superset-ui-core/src/query/processExtraFormData.ts
villebro: Since the value types aren't all the same (most are `string`, one is `TimeRange` and one is `TimeGranularity`), TS is unable to distinguish which value from `ExtraFormDataOverride` is being retrieved, hence this needs to be muted.

packages/superset-ui-core/src/query/processExtraFormData.ts
villebro: Same here

packages/superset-ui-core/test/query/processExtraFormData.test.ts
villebro: This was added in TS 3.9: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

plugins/plugin-chart-table/src/TableChart.tsx
villebro: Good idea - however, let's leave that for a follow-up PR (we should avoid doing bycatch fixes/refactors)

amitmiran137 (approved):

packages/superset-ui-core/src/query/buildQueryObject.ts
simcha90: 👍

packages/superset-ui-core/src/query/constants.ts
simcha90: but if native filter will use `cross_groupby` it can cause some misunderstanding, no? may be something more common? also what the difference between `cross_highlight` and `cross_drilldown`?

packages/superset-ui-core/src/query/constants.ts
simcha90: I still think `values` because in the end we use `values`

packages/superset-ui-core/src/query/processExtraFormData.ts
simcha90: so `string | TimeRange | TimeGranularity` will not work?

packages/superset-ui-core/test/query/processExtraFormData.test.ts
simcha90: 👍

plugins/plugin-chart-table/src/TableChart.tsx
simcha90: 👍

packages/superset-ui-core/src/query/constants.ts
villebro: We don't use the values in `ExtraFormData` - AFAIK `Object.values` is actually `keyof QueryObject` as stated in the `EXTRA_FORM_DATA_OVERRIDE_REGULAR_MAPPINGS` declaration.

packages/superset-ui-core/src/query/constants.ts
villebro: I'm open to another name, but in the interest of time let's decide these asap so we can get this in. @amitmiran137 thoughts?

packages/superset-ui-core/src/query/processExtraFormData.ts
villebro: Not that I know of - because we're indexing `overrideFormData` on `key`, TS won't know what type we're getting at which iteration of the loop.

packages/superset-ui-core/src/query/constants.ts
amitmiran137: drilldown/highlight are almost the same expect that there indication might look different on the chart.
We mostly gonna use drilldown to completely required and rebuild the visual.
And highlight might just highlight what was filtered instead of rebuilding

amitmiran137: Does calling it cross_groupby or groupby  matters now that we explicitly support  it as an appendFormData Option ?
If not so we can just call it groupby and then make sure we document it correctly

villebro: ok, so rename to `groupby`, highlight` and `drilldown`?

villebro: Hmm, `groupby` will be dangerous, as it's practically a reserved keyword on `QueryFormData`. I'd prefer to prefix these to avoid collisions and to make it obvious that they belong to the same category of actions.

amitmiran137: Maybe we can consider the following naming : 

filter_highlight
filter_drilldown
groupby

rusackas (approved):

packages/superset-ui-core/src/query/processExtraFormData.ts
kgabryje: How about `const value = overrideFormData[key as keyof ExtraFormDataOverride];`?

kgabryje (commented):
1 suggestion about removing ts-ignore, other than that LGTM!

packages/superset-ui-core/src/query/processExtraFormData.ts
villebro: That worked well, thanks!

packages/superset-ui-core/src/query/processExtraFormData.ts
villebro: This one seemed trickier to fix, so leaving as-is for now.

PR #1058, opened 2021-04-14
feat(legacy-plugin-chart-country-map): update jupyter notebook script (Slovenia map)

dkrat7:
🏆 Enhancements

I missed updating the Jupyter notebook file when adding Slovenia map.
This is the update.


ktmud (commented):
Thank you!

ktmud (approved):

PR #1059, opened 2021-04-15
fix: Adding codecov script locally

craig-rueda:
In light of recent issues with the remote codecov bash script, this PR downloads the script locally and migrates CI actions to utilize the local script instead.


etr2460 (approved):
thanks!

PR #1066, opened 2021-04-20
feat: add tooltip for full metric names

etr2460:
🏆 Enhancements

Adds a tooltip to the metric name so that if it overflows the container we can still read it.

Test plan:
CI. new unit test, test in superset
Screen Shot 2021-04-20 at 10 54 11 AM
Screen Shot 2021-04-20 at 10 55 03 AM

to: @pkdotson @ktmud @graceguo-supercat


ktmud (approved):
Looks like a reasonable change. We already have tooltips in the Dataset panel.

Although I'm personally a little more in favor of using html title attribute over tooltips for text cutoffs, especially when most of the time the text can be fully displayed. It'd be a little distracting seeing too many tooltips popping out when you move your mouse.

PR #1070, opened 2021-04-23
feat(plugin-chart-echarts): [wip] gauge chart enhancements and fixes

krsnik93:
🏆 Enhancements

This PR addresses feature requests and bugs raised here: https://github.com/apache/superset/issues/14209


junlincc:
@krsnik93 is it ready to turn draft to open?

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -127,7 +127,15 @@ export default function transformProps(chartProps: ChartProps) {
       fontSize: FONT_SIZE_MULTIPLIERS.detailFontSize * fontSize,
     },
   }));
-
+  const min = minVal || calculateMin(transformedData);
```
stephenLYZ, line 130: Should we use `??` to replace `||` ？I'm not sure if it needs to be recalculated only when `minVal` is `null` instead of `0`

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -195,14 +203,16 @@ export default function transformProps(chartProps: ChartProps) {
       show: showPointer,
     };
   }
+  const min = minVal ? minVal : calculateMin(transformedData);
```
xiezhongfu: here `minVal ` is  `null` or `0`?    The same question to `maxVal `

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -70,6 +70,14 @@ const setIntervalBoundsAndColors = (
 const calculateAxisLineWidth = (data: DataRecord[], fontSize: number, overlap: boolean): number =>
   overlap ? fontSize : data.length * fontSize;
 
+const calculateMin = (data: GaugeDataItemOption[]) => {
+  return 2 * Math.min(...data.map(d => d.value).concat([0]));
```
xiezhongfu: why here must be `2`?

xiezhongfu: Why here must be `2`?

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
```diff
@@ -70,6 +71,12 @@ const setIntervalBoundsAndColors = (
 const calculateAxisLineWidth = (data: DataRecord[], fontSize: number, overlap: boolean): number =>
   overlap ? fontSize : data.length * fontSize;
 
+const calculateMin = (data: GaugeDataItemOption[]) =>
+  2 * Math.min(...data.map(d => d.value as number).concat([0]));
```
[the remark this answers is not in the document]
krsnik93: It does not have to be, I was looking at how other BI tools do this and came up with `2`. We could have `1.5`, `1.2` or similar.   The point is to ensure that the max value fits into the chart, but that there is a bit of additional "space" left on the gauge.

plugins/plugin-chart-echarts/src/Gauge/transformProps.ts
krsnik93: I was not aware of the `??` operator. Exactly what is needed here. Thanks a lot!

stephenLYZ: It's ok :). `??` is called [Nullish Coalescing](https://github.com/tc39/proposal-nullish-coalescing) which just for the case of `null` and `undefined` values. Here, if we use the `||` operator, it will also be calculated when the value of minVal is `0`, which is actually unnecessary.So it may be better to use nullish coalescing.

villebro:

PR #1071, opened 2021-04-24
fix(plugin-chart-echarts): remove label line if below threshold

villebro:
🐛 Bug Fix

When `showLabelsThreshold` is set, the label lines are shown despite the labels being missing. This PR removes the hack to make labels empty strings when the threshold is not met and implements the `minShowLabelAngle` (set to `3.6 * showLabelsThreshold` as 100 % == 360 degrees) property that provides native support for removal of thin slices.

### BEFORE
As can be seen in the screenshot, the top slices show a line despite no labels being shown:
image

### AFTER
After the fix, no label lines are shown for the slices that don't exceed the threshold:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
@@ -102,6 +102,7 @@ export default function transformProps(chartProps: EchartsPieChartProps): PieCha
     emitFilter,
   }: EchartsPieFormData = { ...DEFAULT_LEGEND_FORM_DATA, ...DEFAULT_PIE_FORM_DATA, ...formData };
   const metricLabel = getMetricLabel(metric);
+  const minShowLabelAngle = (showLabelsThreshold || 0) * 3.6;
 
   const keys = data.map(datum =>
     extractGroupbyLabel({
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-echarts/src/Pie/transformProps.ts
```diff
@@ -102,6 +102,7 @@ export default function transformProps(chartProps: EchartsPieChartProps): PieCha
     emitFilter,
   }: EchartsPieFormData = { ...DEFAULT_LEGEND_FORM_DATA, ...DEFAULT_PIE_FORM_DATA, ...formData };
   const metricLabel = getMetricLabel(metric);
+  const minShowLabelAngle = (showLabelsThreshold || 0) * 3.6;
```
villebro, line 105: It turns out the `TextControl` with `isFloat: true` returns an empty string when unset, making it important to replace a falsy value for `showLabelsThreshold` with zero.

zhaoyongjie (approved):
LGTM! Thank you for this fix.

junlincc:
Thank you @villebro , i believe the actual implementation/fix is happening in Echarts repo and will be available in their next release?

villebro:
@junlincc you are right - this fix only partly works on the 5.1.0 release - the rest of it is implemented in 5.1.1 (https://github.com/apache/echarts/pull/14702) which is being bumped here: #1074

PR #1072, opened 2021-04-26
fix(native-filters): support group by feature for server side pagination in table

simcha90:
🐛 Bug Fix
1. Add support for group by filter in server side pagination case in table
2. Add support only for aggregate mode


villebro (commented):
Code LGTM; however, this looks like a fix - can you update the title and add some context on what the problem is?

villebro (approved):

villebro (approved):

PR #1073, opened 2021-04-26
chore: adds Emotion's CacheProvider and createCache as exports from Core package

rusackas:
💔 Breaking Changes

🏆 Enhancements
We're now exporting cache features to namespace classes and such. We're also continuing this habit of exporting all things Emotion from Superset-UI Core to avoid multiple instantiated versions in Superset, which causes hard-to-diagnose bugs. 

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-core/package.json
+++ b/packages/superset-ui-core/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
+    "@emotion/cache": "^11.1.3",
     "@emotion/core": "^10.0.28",
     "@emotion/react": "^11.1.5",
     "@emotion/styled": "^10.0.27",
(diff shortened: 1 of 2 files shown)


pkdotson (approved):

packages/superset-ui-core/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
+    "@emotion/cache": "^11.1.3",
```
nytai, line 35: Tested locally, this should work!

packages/superset-ui-core/src/style/index.ts
```diff
@@ -20,7 +20,8 @@ import emotionStyled, { CreateStyled } from '@emotion/styled';
 import { useTheme as useThemeBasic } from 'emotion-theming';
 
 export { ThemeProvider, withTheme } from 'emotion-theming';
-export { css, InterpolationWithTheme } from '@emotion/core';
+export { css, InterpolationWithTheme, CacheProvider } from '@emotion/core';
```
nytai, line 23: Let's also add `ClassNames` and `Global` as I see those are also used in superset

rusackas: Done.

PR #1076, opened 2021-04-27
chore: remove class support from plugin yeoman generator

rusackas:
💔 Breaking Changes
Not sure if this is _really_ a breaking change, but it seems like nobody's in favor of class components these days, so I figured maybe we might as well simplify the Yeoman generator so we don't wind up with a bunch of plugins that are just on the road to a refactor.

This just seemed like a logical thing to do, but wondering if anyone feels otherwise, e.g. @ktmud @villebro @amitmiran137 

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/generator-superset/generators/plugin-chart/index.js
+++ b/packages/generator-superset/generators/plugin-chart/index.js
@@ -24,21 +24,6 @@ module.exports = class extends Generator {
       },
       {
         type: 'list',
-        name: 'componentType',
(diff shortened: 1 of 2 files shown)


suddjian (approved):

ktmud (approved):
LGTM!

amitmiran137 (approved):

villebro (approved):

PR #1078, opened 2021-04-28
feat(plugin-chart-table): disable show_totals by default

ktmud:
🐛 Bug Fix

Show totals adds an additional SQL query to the request, which slows things down. Since it's not a critical feature, we shouldn't enable it by default.

### DIFF
--- a/plugins/plugin-chart-table/src/controlPanel.tsx
+++ b/plugins/plugin-chart-table/src/controlPanel.tsx
@@ -245,7 +245,7 @@ const config: ControlPanelConfig = {
             config: {
               type: 'CheckboxControl',
               label: t('Show totals'),
-              default: true,
+              default: false,
               description: t(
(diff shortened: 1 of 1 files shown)


zhaoyongjie (approved):
LGTM. thanks Jesse

junlincc:
@ktmud it IS a critical feature in table :)  but I agree that showing total should be a user choice, since we have not had a conclusion what to do with the non-additive columns. and perf is def important.

zhaoyongjie:
@junlincc 
Since this is a database calculate aggregate, there are many database optimization strategies here rather than on the BI side.

ktmud:
No it is not.

villebro:
Tableau also has totals disabled by default, which I think makes sense. In general I like the idea that designing charts we want to incrementally add stuff and only prepopulate the options that are needed in the overwhelming majority of cases. I think totals are often not needed, hence would require an extra step for someone who only needs a simpler chart (which is counter intuitive).

PR #1083, opened 2021-04-30
feat(plugin-chart-pivot-table): enable cross filtering

kgabryje:
Enable cross filtering in pivot table chart.
Clicking on any header cell that represents selected rows and columns emits a filter `group_name IN ('value1', 'value2', ...)`.

image
https://user-images.githubusercontent.com/15073128/116720817-e97b4000-a9dc-11eb-800e-9003538d0031.mov

CC @villebro @junlincc

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -104,12 +104,11 @@ export default function PivotTableChart(props: PivotTableProps) {
   const handleChange = useCallback(
     (filters: SelectedFiltersType) => {
       const groupBy = Object.keys(filters);
-      const groupByValues = Object.values(filters);
       setDataMask({
         extraFormData: {
           filters:
             groupBy.length === 0
-              ? []
+              ? undefined
               : groupBy.map(col => {
                   const val = filters?.[col];
                   if (val === null || val === undefined)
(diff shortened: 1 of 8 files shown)


plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -118,16 +90,92 @@ export default function PivotTableChart(props: PivotTableProps) {
...
+
+  const handleChange = useCallback(
+    (filters: SelectedFiltersType) => {
+      const groupBy = Object.keys(filters);
+      const groupByValues = Object.values(filters);
+      setDataMask({
+        extraFormData: {
+          filters:
+            groupBy.length === 0
+              ? []
```
villebro, line 112: To avoid unnecessary data requests, it is now (since https://github.com/apache/superset/pull/14408) preferable not to set `extraFormData` values with empty arrays. Instead they should be set to `undefined. This will help the frontend detect that a new chart data request isn't needed. So a small change here:
```suggestion
              ? undefined
```

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -118,16 +90,92 @@ export default function PivotTableChart(props: PivotTableProps) {
...
+                    val: val as (string | number | boolean)[],
+                  };
+                }),
+        },
+        filterState: {
+          value: groupByValues.length ? groupByValues : null,
+        },
+        ownState: {
+          selectedFilters: filters && Object.keys(filters).length ? filters : null,
+        },
```
villebro, line 132: Do we need `ownState`? I think we should be able to get by with just `filterState`. Usually `ownState` is only needed for more complex use cases where the chart's own `buildQuery` needs to be customized based on interaction with the chart.

villebro (commented):
A few comments

villebro (approved):
LGTM!

PR #1095, opened 2021-05-06
fix(plugin-chart-table): check if control values are set

villebro:
🐛 Bug Fix
When checking for visibility of server side related controls, make sure the control value is set before reading the control value.


zhaoyongjie (approved):
LGTM

PR #1096, opened 2021-05-06
fix: avoid unnecessary requeries caused by ownState

villebro:
🐛 Bug Fix
Using `ownState` will retrigger a new query, which is unnecessary in the majority of cross filtering cases. This moves the selected labels in the ECharts plugin into `filterState`, thus avoiding the need to set `ownState`.

### BEFORE
Notice how clicking on a box causes the chart to trigger a query:
own-before

### AFTER
After moving the selection state to `filterState`, the chart no longer needs to trigger a query when clicking a box:
own-after

### DIFF
--- a/plugins/plugin-chart-echarts/src/BoxPlot/EchartsBoxPlot.tsx
+++ b/plugins/plugin-chart-echarts/src/BoxPlot/EchartsBoxPlot.tsx
@@ -60,8 +60,6 @@ export default function EchartsBoxPlot({
         },
         filterState: {
           value: groupbyValues.length ? groupbyValues : null,
-        },
-        ownState: {
           selectedValues: values.length ? values : null,
         },
       });
--- a/plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/BoxPlot/transformProps.ts
@@ -36,7 +36,7 @@ import { defaultGrid, defaultTooltip, defaultYAxis } from '../defaults';
 export default function transformProps(
   chartProps: EchartsBoxPlotChartProps,
 ): BoxPlotChartTransformedProps {
-  const { width, height, formData, hooks, ownState, queriesData } = chartProps;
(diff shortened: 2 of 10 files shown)


simcha90 (approved):

plugins/plugin-chart-echarts/src/BoxPlot/EchartsBoxPlot.tsx
```diff
@@ -60,8 +60,6 @@ export default function EchartsBoxPlot({
         },
         filterState: {
```
kgabryje, line 61: Why do we still need `value` in `filterState`? I can't see it being used

villebro: It's kind of a reserved property in `filterState`, and is used to indicate what the current selection is. If it's unset, e.g. the filter indicator will think that the filter is unset (this will all be documented)

villebro: Here's with the current master, where neither emitted or received filter indicators are shown:
image
With this change it now looks like this (the indicator is looking at `filterState.value`):
image

kgabryje: Got it, thanks for explanation!

villebro: As a future development, I think it might be a good idea to always populate `filterState.value` with a preformatted string. For example, if we have a range filter that's set to the interval  `10000 < x <= 100000000`, and we're using a number formatter that displays `10000` and `100000000` as `10k` and `100M` respectively, a good indicator string would probably be `(10k, 100M]` or `10k < my_col <= 100M`. Now it would just show it as `[10000, 100000000]`, because `filterState.value` is set to an array with that value.

zhaoyongjie (approved):
Thanks for this optimization. LGTM!

PR #1099, opened 2021-05-09
fix(plugin-chart-echarts): add orderby

xiezhongfu:
🐛 Bug Fix
https://github.com/apache/superset/issues/14534
- add orderby for the fist one
- Maybe should ignore the second one. because user SHOULD write a `null filter` or preprocess the data source for the visualization.


zhaoyongjie (approved):
LGTM!

villebro (approved):

junlincc:
Thanks for adding it! with ORDERBY added, null should be at the bottom showing as a line, which is good enough. 👍
image

PR #1102, opened 2021-05-10
feat(core): clean up datatype logic

villebro:
🐛 Bug Fix

Since the backend now does a better job at inferring generic datatypes, it makes sense to remove some of the old logic that looks at raw datatypes coming back from the backend. Dependent on the following PR: https://github.com/apache/superset/pull/14547


zhaoyongjie (approved):
LGTM, Thanks for the works

PR #1103, opened 2021-05-11
feat(plugin-chart-echarts): Enhancements of treemap tooltip

stephenLYZ:
🏆 Enhancements

This PR contains two enhancements:
1. Tooltip shows the hierarchical path.
2. Rename 'Treemap' to 'Treemap v2'.
3. Set upper label checked by default.

**Note:** The content format of the tooltip is：
```
groupby1 ▸ groupby2 ▸ ... ▸ ...
metric: value
```

### Before
image

### After
image


plugins/plugin-chart-echarts/src/Treemap/transformProps.ts
```diff
@@ -62,6 +62,25 @@ export function formatLabel({
...
+  params,
+  numberFormatter,
+}: {
+  params: TreemapSeriesCallbackDataParams;
+  numberFormatter: NumberFormatter;
+}): string {
+  const { value, treePathInfo } = params;
+  const formattedValue = numberFormatter(value as number);
+
+  const treePath = (treePathInfo ?? []).map(pathInfo => pathInfo?.name || '').filter(Boolean);
```
zhaoyongjie, line 75: There is no need to filter for null values. for instance: ["a", null, "b"] => ["a", "b"]

zhaoyongjie:
> "▸" ">" might be better than / , could you try see how they look on the tooltip?

">" greater than sign with space much butter.

junlincc:
image
image

zhaoyongjie (approved):
LGTM, Thanks for the improvements

PR #1104, opened 2021-05-11
fix: use proper time grain when using native filters or filter box

villebro:
🐛 Bug Fix

Currently the following plugins don't apply the proper time grain formatter when the time grain has been overridden by either a Filter Box or Native Filter:
- Legacy Pivot Table
- Table Chart
- Big Number

This PR adds a util to `@superset-ui/core` for extracting the time grain from the raw form data and implements said function in all charts that support time grain formatting. After the change both Filter Box and Native Filter timegrains will be properly reflected in the time formatter.

### BEFORE
Currently overriding the time grain on the dashboard isn't reflected in the time formatter if it's set to timegrain (see time grain `Week` on the Filter Box):
image

### AFTER
After the change all charts that use the time grain formatter will properly pick up the time grain even on the dashboard:
image

### DIFF
--- a/plugins/plugin-chart-table/src/utils/isEqualColumns.ts
+++ b/plugins/plugin-chart-table/src/utils/isEqualColumns.ts
@@ -31,6 +31,10 @@ export default function isEqualColumns(propsA: TableChartProps[], propsB: TableC
       JSON.stringify(b.formData.columnConfig || null) &&
     isEqualArray(a.formData.metrics, b.formData.metrics) &&
     isEqualArray(a.queriesData?.[0]?.colnames, b.queriesData?.[0]?.colnames) &&
-    isEqualArray(a.queriesData?.[0]?.coltypes, b.queriesData?.[0]?.coltypes)
+    isEqualArray(a.queriesData?.[0]?.coltypes, b.queriesData?.[0]?.coltypes) &&
+    JSON.stringify(a.formData.extraFilters || null) ===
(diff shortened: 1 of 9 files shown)


plugins/plugin-chart-table/src/utils/isEqualColumns.ts
```diff
@@ -31,6 +31,10 @@ export default function isEqualColumns(propsA: TableChartProps[], propsB: TableC
       JSON.stringify(b.formData.columnConfig || null) &&
     isEqualArray(a.formData.metrics, b.formData.metrics) &&
     isEqualArray(a.queriesData?.[0]?.colnames, b.queriesData?.[0]?.colnames) &&
-    isEqualArray(a.queriesData?.[0]?.coltypes, b.queriesData?.[0]?.coltypes)
+    isEqualArray(a.queriesData?.[0]?.coltypes, b.queriesData?.[0]?.coltypes) &&
+    JSON.stringify(a.formData.extraFilters || null) ===
+      JSON.stringify(b.formData.extraFilters || null) &&
+    JSON.stringify(a.formData.extraFormData || null) ===
+      JSON.stringify(b.formData.extraFormData || null)
```
villebro, line 38: This is used to check if chart props have changed by a memoization function, hence needs to ensure that the relevant extra filter data hasn't changed.

zhaoyongjie (approved):
LGTM, thanks for your works!

PR #1110, opened 2021-05-14
fix(plugin-chart-echarts): funnel chart improvements

xiezhongfu:
🏆 Enhancements

###### 1.1: before
onInit before
###### 1.2: after
onInit after

###### 2.1: before
rm LABEL LINE before
###### 2.2: after
rm LABEL LINE after

###### 3.1: before
label's textBorderColor before
###### 3.2: after
label's textBorderColor after

###### 4.1: rm Funnel shape before
Funnel shape before
###### 4.1: rm Funnel shape after
Funnel shape after

###### 5.1: before
legend type before
###### 5.2: after
legend type after


zhaoyongjie (approved):
LGTM

PR #1111, opened 2021-05-15
feat(plugin-chart-echarts): set up orderby by default on Pie chart

xiezhongfu:
SORT BY METRIC checkbox check by default in Pie chart. so the query is:
<pre>
SELECT value AS value,
       SUM(value) AS sum__value
FROM energy_usage
GROUP BY value
ORDER BY sum__value DESC
LIMIT 10000;
</pre>

### before
pie  sort by metric  after

### after
pie  sort by metric  before


zhaoyongjie (approved):
LGTM

PR #1112, opened 2021-05-15
feat(plugin-chart-echarts): add orderby on Radar chart

xiezhongfu:
add SORT BY metrics to Radar chart to avoid random values.

### before
radar sort by metrics before

### after
radar sort by metrics after 2

### DIFF
--- a/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
+++ b/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
@@ -16,14 +16,17 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { buildQueryContext, QueryFormData } from '@superset-ui/core';
+import { buildQueryContext, QueryFormData, getMetricLabel } from '@superset-ui/core';
 
 export default function buildQuery(formData: QueryFormData) {
-  const { metric, sort_by_metric } = formData;
-  return buildQueryContext(formData, baseQueryObject => [
-    {
-      ...baseQueryObject,
-      ...(sort_by_metric && { orderby: [[metric, false]] }),
-    },
-  ]);
+  const { sort_by_metric } = formData;
+  return buildQueryContext(formData, baseQueryObject => {
+    const { metrics = [] } = baseQueryObject;
+    return [
+      {
+        ...baseQueryObject,
+        ...(sort_by_metric && { orderby: metrics.map(metric => [getMetricLabel(metric), false]) }),
+      },
+    ];
+  });
 }
--- a/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
+++ b/plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
@@ -16,16 +16,27 @@
  * specific language governing permissions and limitations
  * under the License.
  */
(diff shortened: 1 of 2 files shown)


zhaoyongjie:
Thanks for the fix!

How do we face sort by specific metric? for instance, order by max(value).

image

xiezhongfu:
> Thanks for the fix!
> 
> How do we face sort by specific metric? for instance, order by max(value).
> 
> image

sort by specific metric.
refer:  https://github.com/apache-superset/superset-ui/blob/master/plugins/plugin-chart-table/src/buildQuery.ts#L68
> * default to ordering by first metric in descending order when no "sort by" metric is set
> * orverride orderby with timeseries metric
em. I do not know why `timeseries metric` is time-series.

wjl520:
how to add radar chart on the dashboard, namely i see nothing when i choose a visualization type for radar chart. Looking forward to your reply. thx.

xiezhongfu:
> how to add radar chart on the dashboard, namely i see nothing when i choose a visualization type for radar chart. Looking forward to your reply. thx.

Please let me know more about your issue. You can show something more about error.

zhaoyongjie:
> how to add radar chart on the dashboard, namely i see nothing when i choose a visualization type for radar chart. Looking forward to your reply. thx.

Hi, the Radar chart is already available on the master branch.

plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
```diff
@@ -70,6 +70,7 @@ const config: ControlPanelConfig = {
       controlSetRows: [
         ['groupby'],
         ['metrics'],
+        ['timeseries_limit_metric'],
```
zhaoyongjie, line 73: I will file a new PR to add orderby shared-control. we don't need this control in the non-timeseries chart.

junlincc:
> how to add radar chart on the dashboard, namely i see nothing when i choose a visualization type for radar chart. Looking forward to your reply. thx.

please file improvement or bug requests once you get to it, we are looking for ways to improve the chart. thank you :)

plugins/plugin-chart-echarts/src/Radar/buildQuery.ts
```diff
@@ -16,14 +16,26 @@
...
 
 export default function buildQuery(formData: QueryFormData) {
-  const { metric, sort_by_metric } = formData;
-  return buildQueryContext(formData, baseQueryObject => [
-    {
-      ...baseQueryObject,
-      ...(sort_by_metric && { orderby: [[metric, false]] }),
-    },
-  ]);
+  const { orderby } = formData;
```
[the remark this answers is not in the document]
zhaoyongjie: You are right. revert change, use `timeseries_limit_metric` back. the `orderby` leave it to future new sortby control.

plugins/plugin-chart-echarts/src/Radar/controlPanel.tsx
xiezhongfu: refer:https://github.com/apache-superset/superset-ui/pull/1112#discussion_r636270729

zhaoyongjie (approved):
LGTM, Thanks for this.

PR #1113, opened 2021-05-15
feat(plugin-chart-echarts): SORT BY METRIC add checkbox to Gauge

stephenLYZ:
🏆 Enhancements
support sort by metric for the Gauge chart

cc @zhaoyongjie @junlincc


zhaoyongjie (approved):
LGTM

PR #1114, opened 2021-05-17
feat(core): remove defaults for time range filter and Metrics

kgabryje:
Changes default Time Range from "Last week" to "No filter"
Changes using "COUNT(*)" or other saved metric as a default metric to empty metrics control.

In the case of Table chart, removing default metric was a bit more complex. We use Groupby, Metrics and Percentage Metrics there, and at least one of them must have a value to construct a correct query. However, we can only write control validator that use values of its own control - for instance, we can't create a validator for groupby that checks if metrics or percentage_metrics are empty or not. We also couldn't simply mark all 3 controls as required, as only 1 is really required. The result was that upon creating a table chart, an incorrect request to backend was sent and we displayed an `Empty query?` error. It was quite a bad user experience - a strange error was the first thing that user sees after creating a chart (and what's more, we sent a query that we knew would fail).
For that reason, I created something like a "dynamic" validator - we check the state of groupby, metrics and percentage_metrics in `mapStateToProps` field of groupby. If all of those fields are empty, we set an error in a new field `externalValidationErrors`. Since `mapStateToProps` function is run only when the field changes, we needed a way to trigger it when not only groupby, but also metrics or percentage_metrics fields change. To achieve that, I added a field `rerender: ['groupby']` to metrics and percentage_metrics. It is later checked in `superset/superset-frontend/src/explore/reducers` - if `rerender` array exists, it triggers `getControlStateFromControlConfig` for each control in `rerender` array, which in turn triggers `mapStateToProps` of that control. Then, `externalValidationErrors` are merged with calculated `validationErrors`.
Checkout the video to see the result.

https://user-images.githubusercontent.com/15073128/118494563-94506580-b722-11eb-9015-845ff0cdb632.mov

Screenshot 2021-05-17 at 10 45 42

To test, see https://github.com/apache/superset/pull/14661.
CC: @villebro @junlincc

### DIFF
--- a/plugins/plugin-chart-table/src/controlPanel.tsx
+++ b/plugins/plugin-chart-table/src/controlPanel.tsx
@@ -36,10 +36,11 @@ import {
   ControlPanelsContainerProps,
   ControlStateMapping,
   D3_TIME_FORMAT_OPTIONS,
-  ExtraControlProps,
   QueryModeLabel,
   sections,
   sharedControls,
+  ControlPanelState,
+  ExtraControlProps,
 } from '@superset-ui/chart-controls';
 
 import i18n from './i18n';
@@ -160,6 +161,18 @@ const config: ControlPanelConfig = {
             name: 'groupby',
             override: {
               visibility: isAggMode,
+              mapStateToProps: (state: ControlPanelState, controlState: ExtraControlProps) => {
+                const originalMapStateToProps = sharedControls?.groupby?.mapStateToProps;
+                // @ts-ignore
+                const newState = originalMapStateToProps?.(state, controlState) ?? {};
+                newState.externalValidationErrors =
+                  ensureIsArray(state.controls.metrics?.value).length === 0 &&
+                  ensureIsArray(state.controls.percent_metrics?.value).length === 0 &&
+                  ensureIsArray(controlState.value).length === 0
+                    ? ['Group by, metrics or percentage metrics must have a value']
+                    : [];
+                return newState;
+              },
             },
           },
         ],
@@ -169,6 +182,7 @@ const config: ControlPanelConfig = {
             override: {
               validators: [],
               visibility: isAggMode,
+              rerender: ['groupby'],
             },
           },
           {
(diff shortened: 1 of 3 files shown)


plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -136,6 +140,23 @@ const config: ControlPanelConfig = {
...
+                  newState.options = options;
+                }
+                newState.externalValidationErrors =
+                  ensureIsArray(state.controls.metrics?.value).length === 0 &&
+                  ensureIsArray(state.controls.percent_metrics?.value).length === 0 &&
+                  ensureIsArray(controlState.value).length === 0
+                    ? ['Group by, metrics or percentage metrics must have a value']
+                    : [];
+                return newState;
+              },
```
villebro, line 159: I think it would be nice if we did the same check on `metrics` and `percent_metrics` to paint them red, too.

villebro (commented):
LGTM - my only concern is only painting `groupby` red, as all three controls are jointly in an error state if at least one value is unpopulated. Would be curious to hear others thoughts on this.

plugins/plugin-chart-table/src/controlPanel.tsx
kgabryje: Done. Check out the new video in PR description

plugins/plugin-chart-table/src/controlPanel.tsx
```diff
@@ -103,12 +112,20 @@ const percent_metrics: typeof sharedControls.metrics = {
...
     savedMetrics: datasource?.metrics || [],
     datasourceType: datasource?.type,
     queryMode: getQueryMode(controls),
+    externalValidationErrors:
+      isAggMode({ controls }) &&
+      ensureIsArray(controls.groupby?.value).length === 0 &&
+      ensureIsArray(controls.metrics?.value).length === 0 &&
+      ensureIsArray(controlState.value).length === 0
+        ? [t('Group By, Metrics or Percentage Metrics must have a value')]
+        : [],
```
villebro, line 126: I'm temped to ask if we could break this out so we don't have to repeat the same logic three times?

villebro (approved):
LGT with one minor nit

plugins/plugin-chart-table/src/controlPanel.tsx
kgabryje: Done

villebro (approved):

PR #1121, opened 2021-05-18
feat(core): add support for empty results and refresh indicator

villebro:
🏆 Enhancements

Add support for empty datasets without rendering the "No Results" placeholder text and an flag to notify the chart that it's refreshing data without being forced to use the default spinner. Needed to be able to add support for "Search all filter options" feature in Select Filter.

See below an example of the native select filter with the feature turned on (the chart loading times have been exaggerated by placing a sleep on the backend): 
https://user-images.githubusercontent.com/33317356/118638272-5402ed00-b7df-11eb-9a2c-a9e78b297c63.mp4


kgabryje (approved):
LGTM

kgabryje (approved):
Re-review - LGTM again!

PR #1122, opened 2021-05-18
feat(core): add orderyby shared control

zhaoyongjie:
🏆 Enhancements
Add a new shared control "orderby" for non-timeseries chart.

related PR: https://github.com/apache-superset/superset-ui/pull/1112

### DIFF
--- a/packages/superset-ui-chart-controls/src/shared-controls/index.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/index.tsx
@@ -69,7 +69,7 @@ import {
   dnd_adhoc_filters,
   dnd_adhoc_metric,
   dnd_adhoc_metrics,
-  dnd_timeseries_limit_metric,
+  dnd_sort_by,
   dndColumnsControl,
   dndEntity,
   dndGroupByControl,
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart-controls/src/shared-controls/index.tsx
```diff
@@ -481,7 +481,8 @@ const sharedControls = {
   time_range,
   row_limit,
   limit,
-  timeseries_limit_metric: enableExploreDnd ? dnd_timeseries_limit_metric : timeseries_limit_metric,
+  timeseries_limit_metric: enableExploreDnd ? dnd_sort_by : sort_by,
+  orderby: enableExploreDnd ? dnd_sort_by : sort_by,
```
villebro, line 485: Since this control relates to ordering by metric (as opposed to ordering by e.g. a column name), perhaps we should rename `orderby` to `orderby_metric`?

zhaoyongjie: no problem at all.

But this control is just for declaring the orderby clause in the SQL, and I guess if we add dimension(column) sorting, we'll probably do it in this control too.

villebro: Good points - if we don't think there will be a need for separate ordering controls for 1) only columns 2) only metrics 3) columns and metrics, then this is probably the right solution for now 👍

villebro (approved):

junlincc:
@zhaoyongjie please list all chart that are added this control.

zhaoyongjie:
> @zhaoyongjie please list all chart that are added this control.
Now there is no chart with this control.

PR #1125, opened 2021-05-20
feat(plugin-chart-echarts): Treemap improvements

stephenLYZ:
🏆 Enhancements
- contribution mode show % of parent in the tooltip
- change upper label color to white
- remove label position control
- support only one metric

image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Treemap/buildQuery.ts
+++ b/plugins/plugin-chart-echarts/src/Treemap/buildQuery.ts
@@ -16,12 +16,29 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { buildQueryContext, QueryFormData } from '@superset-ui/core';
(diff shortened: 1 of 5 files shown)


junlincc:
@stephenLYZ this is great!! 
please provide screenshot for each new feature. i think showing % in tooltip is sufficient without adding contribution mode

zhaoyongjie:
What does the largest color block represent? How are multiple metrics represented on the treemap?

image

zhaoyongjie:
not using order by correctly

image

stephenLYZ:
> What does the largest color block represent? How are multiple metrics represented on the treemap?
> 
> image

I think we can only use one metric(first one) for the treemap viz cause this defines the size (% of total) of each rectangle in one dimension. Maybe we can place the other metrics in the label or tooltip. @junlincc What do u think about it?

zhaoyongjie (approved):
LGTM

PR #1126, opened 2021-05-21
fix(plugin-chart-echarts): sanitize series from html tags

villebro:
🐛 Bug Fix

Remove html tags from series names in tooltip.

### BEFORE
Currently a series names that could be html tags don't show correctly on the tooltip(the series name `<NULL>` which can be seen on the legend does not show):
image

### AFTER
After they show correctly (legend omitted in screenshot):
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/utils/series.ts
+++ b/plugins/plugin-chart-echarts/src/utils/series.ts
@@ -200,3 +200,7 @@ export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
     };
   });
 }
+
+export function sanitizeHtml(text: string): string {
+  return text.replace('<', '&lt;').replace('>', '&gt;');
+}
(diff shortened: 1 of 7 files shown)


stephenLYZ:
@villebro great job! A small suggestion, maybe we can sanitize some general string like `echarts.format.encodeHTML`:

image

villebro:
> @villebro great job! A small suggestion, maybe we can sanitize some general string like `echarts.format.encodeHTML`:

Yes, much cleaner approach. I'll update accordingly 👍

plugins/plugin-chart-echarts/src/utils/series.ts
```diff
@@ -200,3 +200,7 @@ export function dedupSeries(series: SeriesOption[]): SeriesOption[] {
     };
   });
 }
+
+export function sanitizeHtml(text: string): string {
+  return format.encodeHTML(text);
+}
```
villebro, line 206: @stephenLYZ I left the original `sanitizeHtml` function in place so we can more easily swap `encodeHTML` out if needed later.

kgabryje (approved):
LGTM

PR #1128, opened 2021-05-21
feat: Add force db FeatureFlag to force users to use SSL on DB connections

hughhhh:
🏠 Internal
- Adding with new feature flag to allow users to force ssl on database connections


suddjian:
Should we add the flag to `DEFAULT_FEATURE_FLAGS` in `config.py` as well?

PR #1129, opened 2021-05-24
feat(core): add support for case insensitive LIKE operator

villebro:
🏆 Enhancements

Adds support for the `ILIKE` binary operator: https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.operators.ColumnOperators.ilike


kgabryje (approved):
Lgtm

PR #1132, opened 2021-05-24
fix(plugin-chart-pivot-table): fix displaying column names when labels are available

kgabryje:
Fixes https://github.com/apache/superset/issues/14716

In order to test this PR, link `plugin-chart-pivot-table` to Superset and create a Pivot Table chart with rows and columns which have labels assigned in dataset

Before:
image

After:
image

CC: @junlincc


rusackas (approved):
LGTM!

PR #1134, opened 2021-05-26
fix(core): missing translation 

maltoze:
🏆 Enhancements

Fix missing translation on `NoResultsComponent`.


maltoze:
https://github.com/isaacs/github/issues/750

PR #1143, opened 2021-06-02
fix(plugin-chart-echarts): fix tooltip format in mixed ts chart

villebro:
🐛 Bug Fix

Currently the Mixed Timeseries chart doesn't consider the secondary axis formatter when rendering the tooltip.

### BEFORE
Despite "girls" referencing the secondary axis (right) that applies a different formatter than the primary axis (left), the secondary axis format isn't applied in the tooltip (now displays "658k", should be "657586"):
image

### AFTER
Now the tooltip format is set based on the format defined for the axis.
image

Fixes https://github.com/apache/superset/issues/14943


zhaoyongjie (approved):
LGTM! thanks for this fix.

PR #1144, opened 2021-06-02
fix(core): missing translation

maltoze:
🏆 Enhancements

Fix missing translation on NoResultsComponent.

### DIFF
--- a/packages/superset-ui-core/src/chart/components/NoResultsComponent.tsx
+++ b/packages/superset-ui-core/src/chart/components/NoResultsComponent.tsx
@@ -1,11 +1,13 @@
 import React, { CSSProperties, useMemo } from 'react';
+import { t } from '../../translation';
 
 const MESSAGE_STYLES: CSSProperties = { maxWidth: 800 };
 const TITLE_STYLES: CSSProperties = { fontSize: 16, fontWeight: 'bold', paddingBottom: 8 };
 const BODY_STYLES: CSSProperties = { fontSize: 14 };
 const MIN_WIDTH_FOR_BODY = 250;
-const BODY_STRING =
-  'No results were returned for this query. If you expected results to be returned, ensure any filters are configured properly and the datasource contains data for the selected time range.';
+const BODY_STRING = t(
+  'No results were returned for this query. If you expected results to be returned, ensure any filters are configured properly and the datasource contains data for the selected time range.',
+);
 
 const generateContainerStyles: (
   height: number | string,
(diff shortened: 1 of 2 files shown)


packages/superset-ui-core/src/chart/components/NoResultsComponent.tsx
```diff
@@ -39,11 +40,11 @@ const NoResultsComponent = ({ className, height, id, width }: Props) => {
...
       id={id}
       style={containerStyles}
-      title={shouldRenderBody ? undefined : BODY_STRING}
+      title={shouldRenderBody ? undefined : t(BODY_STRING)}
     >
       <div style={MESSAGE_STYLES}>
-        <div style={TITLE_STYLES}>No Results</div>
-        {shouldRenderBody && <div style={BODY_STYLES}>{BODY_STRING}</div>}
+        <div style={TITLE_STYLES}>{t('No Results')}</div>
+        {shouldRenderBody && <div style={BODY_STYLES}>{t(BODY_STRING)}</div>}
```
nytai, line 47: I think it would be better to add the translation to the actual constant value on line 8, eg
```js
const BODY_STRING =
  t('No results were returned for this query. If you expected results to be returned, ensure any filters are configured properly and the datasource contains data for the selected time range.');
```

maltoze: This will not pass CI...

nytai: what's the error

maltoze: @nytai https://github.com/apache-superset/superset-ui/runs/2671917720

image

etr2460:
Do translations actually work in superset-ui? I thought this was still unsupported....

maltoze:
@etr2460 seems work well.

image

suddjian (approved):
LGTM

ktmud (approved):

packages/superset-ui-core/src/chart/components/NoResultsComponent.tsx
ktmud: You can also move `const BODY_STRING = t(...)` into `NoResultsComponent`. In general, we want translation strings to use literal keys instead of variables (it makes future refactoring easier).

maltoze: @ktmud Done!

PR #1150, opened 2021-06-06
feat(plugin-chart-echarts): Echarts Waterfall

stephenLYZ:
🏆 Enhancements

Waterfall charts are useful for understanding how an initial value (like net income) is affected by a series of positive and negative changes. 

For example, we can see the trend of personal annual income. Here we use the year as the `series`.
image

Moreover, we can use the value in `breakdowns` to add additional data to the chart. It adds the contributors to increases or decreases for each series.
image

This pr refers to #705 (Thanks for Simcha's work.


xiezhongfu:
Thanks stephen.
Core-code is LGTM for me. Next, I will review the code in detail.

zhaoyongjie (commented):
Awesome works! Leave some issues.

1. the `null` value needs to be handled in tooltip
image


2. please use `seriese` control as category, and need to DND support
image

3. I can't hidden `metric value` and `tooltip` on the viz

image

4. the `Breakdown` is broken
image

plugins/plugin-chart-echarts/src/Waterfall/index.ts
```diff
@@ -0,0 +1,55 @@
...
+   * It is worth noting that `buildQuery` and is optional, and only needed for
+   * advanced visualizations that require either post processing operations
+   * (pivoting, rolling aggregations, sorting etc) or submitting multiple queries.
+   */
+  constructor() {
+    super({
+      buildQuery,
+      controlPanel,
+      loadChart: () => import('./EchartsWaterfall'),
+      metadata: new ChartMetadata({
```
zhaoyongjie, line 46: needs category property here. Might `Ranking`  is  is an appropriate category

rusackas: `Evolution` seems reasonable to me.

stephenLYZ:
> Awesome works! Leave some issues.
> 
> 1. the `null` value needs to be handled in tooltip
>    image
> 2. please use `seriese` control as category, and need to DND support
>    image
> 3. I can't hidden `metric value` and `tooltip` on the viz
> 
> image
> 
> 1. the `Breakdown` is broken
> 
> image

1. fixed.
2. fixed.
3. add `show_value` control and `rich_tooltip` control for this chart. I'm not sure 
4. The logic of the previous breakdown was to hide the breakdown of the first series and only show the total (a reference to powerBI?). Maybe it is strange because it will cause the data and the chart to not correspond, I removed this special logic first and need to discuss whether this special logic is needed.

villebro:

PR #1156, opened 2021-06-08
fix(plugin-chart-echarts): order by timeseries limit metric

villebro:
🐛 Bug Fix
Order by timeseries limit metric if no orderby is provided (for parity with NVD3 Timeseries viz'):
image

Added tests to ensure that ordering is correct both when `orderby` is specified/unspecified combined with timeseries limit metric.


zhaoyongjie (approved):
LGTM

junlincc:
can we get this PR to superset main repo before im out~ i want to help testing! @villebro

PR #1157, opened 2021-06-08
feat(plugin-chart-table): added emit target name

cccs-jc:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
@@ -60,7 +60,6 @@ export default function ColumnConfigControl<T extends ColumnConfig>({
     });
   } else {
     Object.values(configFormLayout).forEach(array_of_array => {
-      console.log(array_of_array);
       const index = array_of_array.findIndex(arr => arr.includes('emitTarget'));
       if (index > -1) {
         array_of_array.splice(index, 1);
--- a/packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
+++ b/packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
@@ -32,6 +32,7 @@ export type ColumnConfigControlProps<T extends ColumnConfig> = ControlComponentP
   queryResponse?: ChartDataResponseResult;
   configFormLayout?: ColumnConfigFormLayout;
   appliedColumnNames?: string[];
+  emitFilter: boolean;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


junlincc:
@cccs-jc Hey Jc, can you give a more context what this PR is about? thanks!

cccs-jc:
yes, sorry. It's for this issue
https://github.com/apache/superset/issues/13966

packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/constants.tsx
```diff
@@ -33,12 +33,22 @@ export type SharedColumnConfigProp =
...
   | 'd3NumberFormat'
   | 'd3SmallNumberFormat'
   | 'd3TimeFormat'
   | 'horizontalAlign'
   | 'showCellBars';
 
+const emitTarget: ControlFormItemSpec<'Input'> = {
+  controlType: 'Input',
+  label: t('Emit Target'),
+  description: D3_FORMAT_DOCS,
```
villebro, line 46: We should add a proper description here. Something like `t('If you wish to specify a different target column than the original column, it can be entered here')`.

packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/constants.tsx
```diff
@@ -33,12 +33,22 @@ export type SharedColumnConfigProp =
...
+  defaultValue: 'this',
```
villebro, line 47: I think the default value should be `''`, and in that case we should emit the original column name (as the logic states in `getEmitTarget`).

villebro (commented):
Great feature! A few minor tweaks needed, other than that looks good.

packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
```diff
@@ -48,8 +49,25 @@ export default function ColumnConfigControl<T extends ColumnConfig>({
...
 }: ColumnConfigControlProps<T>) {
+  if (emitFilter) {
+    Object.values(configFormLayout).forEach(array_of_array => {
+      if (!array_of_array.some(arr => arr.includes('emitTarget'))) {
+        array_of_array.push(['emitTarget']);
+      }
+    });
+  } else {
+    Object.values(configFormLayout).forEach(array_of_array => {
+      console.log(array_of_array);
```
villebro, line 63: This shouldn't be here
```suggestion
```

villebro (changes requested):
Need to remove the `console.log`

packages/superset-ui-chart-controls/src/shared-controls/components/ColumnConfigControl/ColumnConfigControl.tsx
cccs-jc: Sorry about that. I should have spotted that.

villebro (approved):

PR #1161, opened 2021-06-09
fix(plugin-chart-echarts): fix time-series chart xAxisShowMin(Max)Label default value

xiezhongfu:
🐛 Bug Fix
https://github.com/apache/superset/issues/15042

### before
https://github.com/apache/superset/issues/15042#issuecomment-856947476

### after
* the reason is https://github.com/apache/superset/issues/15042#issuecomment-858616822
* the solution is https://github.com/apache-superset/superset-ui/pull/1161#issuecomment-873357465

echarts-label

### DIFF
--- a/plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
@@ -56,8 +56,6 @@ const {
   yAxisBounds,
   zoomable,
   xAxisLabelRotation,
-  xAxisShowMinLabel,
-  xAxisShowMaxLabel,
 } = DEFAULT_FORM_DATA;
 const config: ControlPanelConfig = {
   controlPanelSections: [
@@ -331,30 +329,6 @@ const config: ControlPanelConfig = {
               description: `${D3_TIME_FORMAT_DOCS}. ${t(
                 'When using other than adaptive formatting, labels may overlap.',
               )}`,
-            },
-          },
-        ],
-        [
-          {
-            name: 'xAxisShowMinLabel',
-            config: {
-              type: 'CheckboxControl',
-              label: t('Show Min Label'),
-              default: xAxisShowMinLabel,
-              renderTrigger: true,
-              description: t('Show Min Label'),
-            },
-          },
-        ],
-        [
-          {
-            name: 'xAxisShowMaxLabel',
-            config: {
-              type: 'CheckboxControl',
-              label: t('Show Max Label'),
-              default: xAxisShowMaxLabel,
-              renderTrigger: true,
-              description: t('Show Max Label'),
             },
           },
         ],
--- a/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -79,8 +79,6 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
     stack,
     truncateYAxis,
     yAxisFormat,
-    xAxisShowMinLabel,
-    xAxisShowMaxLabel,
     xAxisTimeFormat,
     yAxisBounds,
     yAxisTitle,
@@ -148,8 +146,6 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
     xAxis: {
       type: 'time',
       axisLabel: {
-        showMinLabel: xAxisShowMinLabel,
-        showMaxLabel: xAxisShowMaxLabel,
         formatter: xAxisFormatter,
         rotate: xAxisLabelRotation,
       },
--- a/plugins/plugin-chart-echarts/src/Timeseries/types.ts
+++ b/plugins/plugin-chart-echarts/src/Timeseries/types.ts
@@ -58,6 +58,8 @@ export type EchartsTimeseriesFormData = {
   truncateYAxis: boolean;
   yAxisFormat?: string;
   yAxisTitle: string;
+  xAxisShowMinLabel?: boolean;
+  xAxisShowMaxLabel?: boolean;
   xAxisTimeFormat?: string;
   timeGrainSqla?: TimeGranularity;
   yAxisBounds: [number | undefined | null, number | undefined | null];
--- a/plugins/plugin-chart-echarts/src/Timeseries/types.ts
+++ b/plugins/plugin-chart-echarts/src/Timeseries/types.ts
@@ -58,8 +58,6 @@ export type EchartsTimeseriesFormData = {
   truncateYAxis: boolean;
   yAxisFormat?: string;
   yAxisTitle: string;
-  xAxisShowMinLabel?: boolean;
-  xAxisShowMaxLabel?: boolean;
   xAxisTimeFormat?: string;
   timeGrainSqla?: TimeGranularity;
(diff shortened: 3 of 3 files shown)


plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
```diff
@@ -352,9 +355,12 @@ const config: ControlPanelConfig = {
             config: {
               type: 'CheckboxControl',
               label: t('Show Max Label'),
-              default: xAxisShowMaxLabel,
+              default: xAxisShowMinLabel,
```
villebro, line 358: Shouldn't this be unchanged?

xiezhongfu: em, I thought `default ` should be removed. I will correct it.

plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
```diff
@@ -340,9 +340,11 @@ const config: ControlPanelConfig = {
             config: {
               type: 'CheckboxControl',
               label: t('Show Min Label'),
-              default: xAxisShowMinLabel,
               renderTrigger: true,
               description: t('Show Min Label'),
+              mapStateToProps: () => ({
+                value: xAxisShowMinLabel,
```
zhaoyongjie, line 346: could we just use the following code without mapStateToProps?
```
{
    ....
    ....
    description: t('Show Min Label'),
    value: xAxisShowMinLabel,
}
```

zhaoyongjie:
I tested this PR using `birth_names` dataset

0. open the `explore` page
1. switch to `time-series chart`
2. add `sum__num` to metrics control
3. change `time range` to `Year`
4. change `time range` to `No filter`(may be default value)
5. `customize tab` are default values at all

observe chart, the `min label of xaxis` still in xaxis

image

plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
xiezhongfu: I tried. It does not work as expected.

xiezhongfu:
@zhaoyongjie  I test the pr as you commented. It does not work as expected indeed. Maybe I need to dig the source code of `echarts`. Do you have any idea about it?

image

plugins/plugin-chart-echarts/src/Timeseries/controlPanel.tsx
zhaoyongjie: `mapStateToProps` means that "mapping state of the store to the property of the component".

villebro:
@xiezhongfu it appears the checkbox control doesn't properly map defaults to values. When a `CheckboxControl` value is set to `true`, it maps correctly. However, when it is set to `false`, the value becomes `undefined`. So I think this comes down to two problems:

1. ECharts apparently has trouble handling `showMinLabel` being set to `undefined. So a simple fix here would be to just double bang the value:
```diff
diff --git a/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts b/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
index 39f63ecf..a97f7ff1 100644
--- a/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -149,8 +149,8 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
     xAxis: {
       type: 'time',
       axisLabel: {
-        showMinLabel: xAxisShowMinLabel,
-        showMaxLabel: xAxisShowMaxLabel,
+        showMinLabel: !!xAxisShowMinLabel,
+        showMaxLabel: !!xAxisShowMaxLabel,
         formatter: xAxisFormatter,
         rotate: xAxisLabelRotation,
       },
```
We could also open a PR on ECharts to make sure it doesn't assume `undefined` is `true` (I don't think it's expected behavior).

2. The `CheckboxControl` should not map `false` default value to `undefined`. I believe this change needs to be done on `apache/superset`.

xiezhongfu:
@villebro  thanks very much. 

I found Superset's source code in https://github.com/apache/superset/issues/15042#issuecomment-857850979. and Echarts' source code in https://github.com/apache/superset/issues/15042#issuecomment-858616822.

> From an Echarts team member said, setting `showMinLabel` to false does not force it to be displayed (setting it to true can force the display).

I will spend time digging into Echart's source deeply.

xiezhongfu:
@villebro @zhaoyongjie  thanks you very much.

I try to make `showMinLabel` and `showMaxLabel` in a control state by forcing them to display and controlling their visibility.

But there is a difficult scene to deal with. eg: `Adaptative formating` option of TIME FORMAT is `undefined` not `Function`. So we can not use a custom `formatter`.
Maybe that scene can be ignored?

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -149,9 +153,25 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
...
+                const isMaxLabel = timestamps[timestamps.length - 1] === value;
+                if (isMinLabel) {
+                  return xAxisShowMinLabel ? xAxisFormatter(value) : '';
+                }
+                if (isMaxLabel) {
+                  return xAxisShowMaxLabel ? xAxisFormatter(value) : '';
+                }
+
+                return xAxisFormatter(value);
+              },
```
villebro, line 174: I don't think we need to explicitly remove the min/max values if the checkboxes are unchecked. I only think we need to make sure that the min and max values are showing if the checkboxes are checked. Therefore, I think the only change required here is
```js
axisLabel: {
    showMinLabel: !!xAxisShowMinLabel,
    showMaxLabel: !!xAxisShowMaxLabel,
}
```
to make sure that if the control value is 
- `undefined` or `false`: let ECharts decide what ticks to show.
`true`: forces showing the min/max values.

@zhaoyongjie wha do you think?

villebro (changes requested):
To keep this change as simple as possible, I'd still prefer to just add the double bang to the control values, as it should fix the bug.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
xiezhongfu: `undefined ` and `null` are `auto` for ECharts.  And `false` does not force to be not displayed.

@junlincc Do you have any suggestions?

junlincc: sorry for being late to the party! please correct me if i understand the discussion inaccurately 
`undefined` = initial state, `null` = value does not exist, `false` = unchecking the box correct? 
what does `auto` do? 

Proposing expected behavior
1. have both checkboxes checked and showing min & max on the x-axis by default when users entering Explore the first time
2. changes should be sticky- 1) unchecking the boxes -> hide values, change should persist/sticky for users 2) rechecking boxes -> show values, change persist

junlincc:
after discussion....given this is a known limitation which can't be easily solved in Echarts, let's kill the two checkboxes. have min max and x ticks display auto, to avoid overlapping. :)

zhaoyongjie (approved):
LGTM

PR #1176, opened 2021-06-16
fix(core): remove height and width requirement on SuperChart

villebro:
🐛 Bug Fix

When rendering a `SuperChart` with percentage width/height, the `SuperChartCore` component can sometimes get unmounted and remounted due to the initial width and height being zero before `ParentSize` provides non-zero values, causing major side-effects such as new queries being sent to the backend. This PR removes this check, making the logic for dynamic widths/heights more in-line with that of its non-dynamic counterpart.

Rejected alternative: In an earlier iteration of this PR, we introduced the state `hasBeenRendered`, which is `false` by default, but gets set to `true` at the first render. This way we could avoid rendering until width and height are greater than zero for the first time, while avoiding unmounting if they become zero at a later stage. Since this logic is not applied for non-dynamic widths/heights, and It was deemed unlikely that anyone is relying on this behavior, it was decided to prefer the simpler approach for now. Please check the initial commit for the alternative approach.

### DIFF
--- a/packages/superset-ui-core/src/chart/components/SuperChart.tsx
+++ b/packages/superset-ui-core/src/chart/components/SuperChart.tsx
@@ -55,7 +55,7 @@ export type Props = Omit<SuperChartCoreProps, 'chartProps'> &
 
 type PropsWithDefault = Props & Readonly<typeof defaultProps>;
 
-export default class SuperChart extends React.PureComponent<Props, { hasBeenRendered: boolean }> {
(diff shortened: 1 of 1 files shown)


michael-s-molina (approved):
Great description. If we find a scenario in the future where this is required we can easily rollback.

kgabryje:
Tested on a few different charts (nvd3, echarts), nothing breaks. And the performance boost for rendering native filters is very noticable 🙂 LGTM

kgabryje (approved):

PR #1180, opened 2021-06-19
fix(plugin-chart-echarts): radar legend control is duplicated

stephenLYZ:
🐛 Bug Fix
fix https://github.com/apache/superset/issues/15266

#### before
image

#### after
image


zhaoyongjie (approved):
LGTM

PR #1183, opened 2021-06-22
fix(legacy-preset-chart-nvd3): dist_bar use a guideline and floating tooltip 

xiezhongfu:
🐛 Bug Fix
https://github.com/apache/superset/issues/15097

### after
dist_bar-tooltip

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/src/NVD3Vis.js
+++ b/plugins/legacy-preset-chart-nvd3/src/NVD3Vis.js
@@ -424,6 +424,7 @@ function nvd3Vis(element, props) {
           width = computeBarChartWidth(data, isBarStacked, maxWidth);
         }
         chart.width(width);
+        chart.useInteractiveGuideline(true);
         break;
 
       case 'pie':


villebro:
I think the intent in https://github.com/apache/superset/issues/15097 is to add the possibility to add additional metrics that aren't displayed on the chart, but only in the tooltip. I think this should be fully doable, but we would need to do the following:
- Add a new optional control called "Tooltip metrics".
- Metrics that are added to that control get added to the chart data request (would need to be added to `viz.py` as this is a legacy chart).
- when drawing the chart, only the regular metrics would be drawn on the chart, while the tooltip metrics would only be shown in the tooltip panel.

I think this would be a cool feature and a nice POC for adding this type of functionality to other charts.

zhaoyongjie:
This is a good experiment. I totally agree with making a `tooltip` control, and this control can be placed in the metric showed inside the tooltip of viz.

Exemple of tableau: 

image

junlincc:
how does it look when bars are stacked?

villebro:

PR #1184, opened 2021-06-22
fix: NVD3 bar chart sort by metric

zhaoyongjie:
🐛 Bug Fix
currently, nvd3 bar chart(non-timeseries) can not be sorted by metric. this PR fixed it.

### before
issue link: https://github.com/apache/superset/issues/15260

### after
https://user-images.githubusercontent.com/2016594/122980678-83eb7480-d3cb-11eb-9898-3a2529f434b0.mp4


zhaoyongjie:
There is something wrong with this fix, I am trying other methods

PR #1187, opened 2021-06-23
fix(plugin-chart-echarts): [time-series][mixed timeseries] marker control does not work

stephenLYZ:
🐛 Bug Fix

Unchecking rich tooltip will force marker to be shown, and marker control is not working. This is to support item-based tooltips. But I'd almost prefer to use only marker checkbox to control. The only side effect is that we need to check the marker after unchecking rich tooltip. But it can guarantee a single responsibility.
Or we can check the marker control when unchecking rich tooltip. It may be better implemented in dynamic control.

related to https://github.com/apache/superset/issues/15189

### before
https://user-images.githubusercontent.com/81597121/122186275-bd0f6c00-ce42-11eb-8886-448806d93993.mov

### after

https://user-images.githubusercontent.com/11830681/123062153-d6b24400-d43e-11eb-98ee-3d51a2885699.mov

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -129,7 +129,6 @@ export default function transformProps(chartProps: ChartProps): EchartsProps {
       opacity,
       seriesType,
       stack,
-      richTooltip,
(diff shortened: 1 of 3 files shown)


zhaoyongjie (approved):
LGTM. Tested in local

PR #1188, opened 2021-06-23
fix: Pivot table respect D3 format from metric

xneg:
🐛 Bug Fix

Addresses this bug:
apache/superset#7650

This is same fix as from PR https://github.com/apache-superset/superset-ui-plugins/pull/280 for pivot table.


junlincc:
@xneg this is great! thanks for the improvement. @kgabryje please review the code and lmk if it's ready for qa, @jinghua-qa @rosemarie-chiu please add to test case.

junlincc:
@xneg thanks for the PR! i took a look at both pivot table and pivot table v2 on master, they both work as expected - aggregated columns respect d3 format. what version are you at?  and can you add more details to the PR?

kgabryje:
@xneg @junlincc Using d3 formatting from metrics seems to work correctly on current version of pivot table (on the latest master). Here you can see that the metric formatting defined in the dataset overrides formatting selected from Customize tab.
image
image

This feature doesn't work yet in Pivot Table v2 (see https://github.com/apache/superset/issues/14812). I'm currently working on implementation, it should be ready soon.

villebro:
I also checked that this works on current master 👍

xneg:
@junlincc @kgabryje Sorry, I did not provide screenshots. We found this problem when having multiple grouped metrics in one column. To fix this my coworker wrote another code, so I hope he will make PR soon enough.

kgabryje:
@xneg Ah, I see, thanks for providing details. @junlincc @villebro I can confirm that when "Combine metrics" option is checked, d3 formatting defined in dataset is not respected. Here, the `count` metric should have formatting `.1f` applied and it doesn't:
image

Moreover, I don't think we have that feature in Pivot Table v2. @junlincc Do you think we should allocate some time for this to achieve feature parity?

junlincc:
hi @xneg thanks for the additional info! looks like your team also has some interests in contributing to pivot table. let us know what other area you would like to work/collaborating on. 
@kgabryje yes, please do add the functionality to reach feature parity. do you mind also giving this PR a review?  many thanks!🙏

kgabryje:
@xneg I think your change works as expected when there's only 1 metric added. If there are 2 or more metrics and they have custom d3 formats defined, we take the format of the first metric and apply it to all columns in the pivot table. Is that how it's expected to work? cc @junlincc 

https://user-images.githubusercontent.com/15073128/125276716-d526be00-e310-11eb-8c5c-1d8b6dd7d8b9.mov

If that's good enough, we can merge (as we use the same piece of code in some of the other charts, e.g. nvd3 charts). Otherwise, I'd suggest waiting ~1 more week for the next release of Pivot Table v2 and migrating to it - the changes include applying d3 formats defined in dataset for each metric individually. I'll ping you here when it's ready 🙂

kgabryje:
@xneg Feel free to give Pivot Table v2 a shot - there are several improvements compared to the original Pivot Table chart, including d3 formatting from dataset

xneg:
@kgabryje Ok, thank you!
I close this PR cause I do nothing for this in the last few days and we should focus on Pivot Table v2.

PR #1191, opened 2021-06-25
feat(core): Registry listeners

suddjian:
💔 Breaking Changes

🏆 Enhancements

Adds a listener interface to registries. This will make it easier to use registries from React, specifically wrt dynamic plugins because we can do things like make hooks that re-render components when the registry changes.

Cleaned up some types while I was in there.

📜 Documentation

🐛 Bug Fix

🏠 Internal


zhaoyongjie (approved):
LGTM

PR #1196, opened 2021-07-01
fix(legacy-preset-chart-big-number): example images got mixed up

suddjian:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

This puts the trendline image on the big number with trendline, and the regular image on the regular big number

🏠 Internal


junlincc:
thanks @suddjian 🙏! mind adding corresponding screenshot to the PR?

suddjian:
There aren't really screenshots to give you, this doesn't change UI or anything about the plugin. These are just the images used in the examples list in the new modal. The Big Number with Trendline ended up getting the non-trendline images and the non-trendline Big Number got the example image with trendline, so I switched them to be correct.

rusackas (approved):

PR #1203, opened 2021-07-04
feat(plugin-chart-echarts): add  x-filtering in gauge chart

stephenLYZ:
🏆 Enhancements

https://user-images.githubusercontent.com/11830681/124379511-1eac5300-dcea-11eb-8c1e-970719cd38e9.mov

### DIFF
--- a/plugins/plugin-chart-echarts/src/Gauge/EchartsGauge.tsx
+++ b/plugins/plugin-chart-echarts/src/Gauge/EchartsGauge.tsx
@@ -29,11 +29,11 @@ export default function EchartsGauge({
   labelMap,
   groupby,
   selectedValues,
-  formData,
+  formData: { emitFilter },
 }: GaugeChartTransformedProps) {
   const handleChange = useCallback(
     (values: string[]) => {
-      if (!formData.emitFilter) {
+      if (!emitFilter) {
         return;
       }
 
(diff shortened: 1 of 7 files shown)


stephenLYZ:
@ktmud @zhaoyongjie @xiezhongfu  cc

plugins/plugin-chart-echarts/src/Gauge/EchartsGauge.tsx
```diff
@@ -16,10 +16,76 @@
...
-  return <Echart height={height} width={width} echartOptions={echartOptions} />;
+export default function EchartsGauge({
+  height,
+  width,
+  echartOptions,
+  setDataMask,
+  labelMap,
+  groupby,
+  selectedValues,
+  formData,
```
ktmud, line 32:
```suggestion
  formData: { emitFilter }
```

Can probably expand the input args directly if `formData` is not going to be used elsewhere.

ktmud (approved):
LGTM. Although I can't imagine why would someone place multiple ticks on a gauge.... 😱

plugins/plugin-chart-echarts/src/Gauge/EchartsGauge.tsx
stephenLYZ: > LGTM. Although I can't imagine why would someone place multiple ticks on a gauge.... 😱

I do think this kind of case is relatively rare, but it can provide this ability to users.😂
Also, we plan to add x-filter functionality to all Echarts.

junlincc:
thanks for the review!! @ktmud

PR #1204, opened 2021-07-04
feat(plugin-chart-echarts): single select by default for pie chart

stephenLYZ:
🏆 Enhancements
single select by default for pie chart and set highlighting effect when applying x-filtering. 
### after
https://user-images.githubusercontent.com/11830681/124380570-ec055900-dcef-11eb-83cb-1d36f62dd810.mov
### before

https://user-images.githubusercontent.com/11830681/124380602-1d7e2480-dcf0-11eb-8579-8367169d2a21.mov

### DIFF
--- a/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Pie/transformProps.ts
@@ -43,6 +43,7 @@ import {
   sanitizeHtml,
 } from '../utils/series';
 import { defaultGrid, defaultTooltip } from '../defaults';
+import { OpacityEnum } from '../constants';
 
 const percentFormatter = getNumberFormatter(NumberFormats.PERCENT_2_POINT);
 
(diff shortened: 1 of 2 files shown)


plugins/plugin-chart-echarts/src/Pie/transformProps.ts
```diff
@@ -140,11 +140,14 @@ export default function transformProps(chartProps: EchartsPieChartProps): PieCha
...
     });
 
+    const isFiltered = filterState.selectedValues && !filterState.selectedValues.includes(name);
+
     return {
       value: datum[metricLabel],
       name,
       itemStyle: {
         color: colorFn(name),
+        opacity: isFiltered ? 0.3 : 1,
```
zhaoyongjie, line 150: the `opacity` should be added to echart constants.

stephenLYZ: In https://github.com/apache-superset/superset-ui/pull/1205, a constant about opacity was extracted.  I think we can wait for this PR merge to modify this PR.

junlincc (approved):
product sign-off. confirmed that this is expected behavior for X-filtering

zhaoyongjie (approved):
LGTM

PR #1205, opened 2021-07-04
feat(plugin-chart-echarts): x-filtering improvement in radar chart

stephenLYZ:
🏆 Enhancements
single select by default and improve UI when applying x-filtering.

### after

https://user-images.githubusercontent.com/11830681/124386101-4233c580-dd0b-11eb-8916-7c69316bf83e.mov

### before

https://user-images.githubusercontent.com/11830681/124386045-e9642d00-dd0a-11eb-9783-6eaed55088f1.mov

### DIFF
--- a/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
@@ -114,12 +114,19 @@ export default function transformProps(
       groupby.map(col => datum[col]),
     );
 
+    const isFiltered =
+      filterState.selectedValues && !filterState.selectedValues.includes(joinedName);
+
     // generate transformedData
     transformedData.push({
       value: metricsLabel.map(metricLabel => datum[metricLabel]),
       name: joinedName,
       itemStyle: {
         color: colorFn(joinedName),
+        opacity: isFiltered ? 0 : 1,
+      },
+      lineStyle: {
+        opacity: isFiltered ? 0.3 : 1,
       },
       label: {
         show: showLabels,
--- a/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Radar/transformProps.ts
@@ -37,7 +37,7 @@ import {
 import { DEFAULT_LEGEND_FORM_DATA } from '../types';
 import { extractGroupbyLabel, getColtypesMapping, getLegendProps } from '../utils/series';
 import { defaultGrid, defaultTooltip } from '../defaults';
-import { OPACITY } from '../constants';
+import { OpacityEnum } from '../constants';
 
 export function formatLabel({
   params,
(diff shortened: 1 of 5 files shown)


plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -114,12 +114,19 @@ export default function transformProps(
...
     // generate transformedData
     transformedData.push({
       value: metricsLabel.map(metricLabel => datum[metricLabel]),
       name: joinedName,
       itemStyle: {
         color: colorFn(joinedName),
+        opacity: isFiltered ? 0 : 1,
+      },
+      lineStyle: {
+        opacity: isFiltered ? 0.3 : 1,
```
zhaoyongjie, line 129: the opacity should be added to echart constants.

plugins/plugin-chart-echarts/src/Radar/transformProps.ts
```diff
@@ -114,12 +114,19 @@ export default function transformProps(
...
+    const isFiltered =
+      filterState.selectedValues && !filterState.selectedValues.includes(joinedName);
+
     // generate transformedData
     transformedData.push({
       value: metricsLabel.map(metricLabel => datum[metricLabel]),
       name: joinedName,
       itemStyle: {
         color: colorFn(joinedName),
+        opacity: isFiltered ? 0 : 1,
```
zhaoyongjie, line 126: same after

plugins/plugin-chart-echarts/src/constants.ts
```diff
@@ -53,3 +53,9 @@ export const LABEL_POSITION: [LabelPositionEnum, string][] = [
   [LabelPositionEnum.InsideTopRight, 'Inside top right'],
   [LabelPositionEnum.InsideBottomRight, 'Inside bottom right'],
 ];
+
+export const OPACITY = {
```
zhaoyongjie, line 57: The enumerated type follows the PascalCase style.

```suggestion
export enum OpacityEnum {
  Transparent: 0,
  SemiTransparent: 0.3,
  NonTransparent: 1,
}
```

zhaoyongjie: If we only have `Transparent`, `SemiTransparent` and `NonTransparent`, then the constants make sense. If there are many enumerations, it makes no sense. 

Do we have many transparency rate use cases now?

junlincc: @xiezhongfu please help review 🙏

stephenLYZ: In the gauge chart, I use `0.1` as the transparency value, which looks more suitable.

zhaoyongjie: The purpose of the design here is to standardize the constant values of the `opacity rate`, rather than reset them for each viz and different name them. 
- the `COLOR_ALPHA` used in Treemap for the `semi-transparent`
- All except gauge chart are used 0.3 for the `semi-transparent` 

I suggest that this semi-transparent should be consistent in all viz.

image

stephenLYZ: You are right！Thanks for the explanation.I will use this constant in all charts.

zhaoyongjie: Thanks a lot!

zhaoyongjie (approved):
LGTM

PR #1206, opened 2021-07-06
feat: Normalize orderby clause

zhaoyongjie:
🏆 Enhancements
This PR makes a normalized orderby object in QueryObject instead of `timeseries_limit_metric` and `order_desc`,

also fixed `mixed timeseries chart` unable to sort by. related issue: https://github.com/apache/superset/issues/15345

image

### DIFF
--- a/packages/superset-ui-core/src/query/normalizeOrderBy.ts
+++ b/packages/superset-ui-core/src/query/normalizeOrderBy.ts
@@ -1,0 +1,64 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import isEmpty from 'lodash/isEmpty';
+import isBoolean from 'lodash/isBoolean';
+
+import { QueryObject } from './types';
+
+export default function normalizeOrderBy(queryObject: QueryObject): QueryObject {
+  if (Array.isArray(queryObject.orderby) && queryObject.orderby.length > 0) {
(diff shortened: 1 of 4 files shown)


junlincc:
@zhaoyongjie can you add more description on how this PR is gonna impact Explore/chart?

zhaoyongjie:
> @zhaoyongjie can you add more description on how this PR is gonna impact Explore/chart?

this PR fixed the problem that the mixed time-series chart unable to be sorted. related issue: apache/superset#15345

packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // ensure that remove invalid orderby clause
+  const cloneQueryObject = { ...queryObject };
+  delete cloneQueryObject.timeseries_limit_metric;
+  delete cloneQueryObject.order_desc;
+  delete cloneQueryObject.orderby;
+
+  const isAsc = !queryObject.order_desc;
+  if (
+    queryObject.timeseries_limit_metric !== undefined &&
```
stephenLYZ, line 46: we can use `!isNil(queryObject.timeseries_limit_metric)`

packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -0,0 +1,64 @@
...
+      orderbyClause.length === 2 &&
+      !isEmpty(orderbyClause[0]) &&
+      isBoolean(orderbyClause[1])
+    ) {
+      return queryObject;
+    }
+  }
+
+  // ensure that remove invalid orderby clause
+  const cloneQueryObject = { ...queryObject };
```
stephenLYZ, line 39: I think `const cloneQueryObject = omit(queryObject, ['timeseries_limit_metric', 'order_desc', 'orderby'])` can be more concise. It returns a new object.

stephenLYZ (approved):
looks good to me! some suggestions.

packages/superset-ui-core/src/query/normalizeOrderBy.ts
zhaoyongjie: the typescript `omit` seem that return a type.

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

stephenLYZ: here we can use lodash omit

zhaoyongjie: Oh, thanks! Follow package core's pattern, and try to use a more "native" approach at here.

packages/superset-ui-core/src/query/normalizeOrderBy.ts
zhaoyongjie: same after

PR #1211, opened 2021-07-09
feat(plugin-chart-echarts): crossfilter with single selection in the funnel chart

xiezhongfu:
add cross-filter for funnel

### after
x-filter-funnel
in `Video Game Sales` dashbord.


zhaoyongjie (approved):
LGTM

PR #1215, opened 2021-07-13
feat(chart-controls): implement color formatters utils

kgabryje:
This PR implements utils functions and types necessary for adding conditional formatting to Pivot Table v2 and Table charts.
See https://github.com/apache/superset/pull/15651 for final result


michael-s-molina (approved):
LGTM

PR #1218, opened 2021-07-14
feat(plugin-chart-table): implement conditional formatting

kgabryje:
Adds conditional formatting to Table chart.
Conditional formatting feature is described in https://github.com/apache/superset/pull/15651
In the case of Table chart, if conditional formatting is defined (at least 1 formatter has been added), cell bars are disabled and replaced with formatter's colors.

https://user-images.githubusercontent.com/15073128/125609934-3a04ff76-c8a2-48f3-916e-23c5ca567fd0.mov

CC @junlincc @rusackas @michael-s-molina

### DIFF
--- a/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/plugins/plugin-chart-table/src/TableChart.tsx
@@ -163,6 +163,7 @@ export default function TableChart<D extends DataRecord = DataRecord>(
     sortDesc = false,
     filters: initialFilters = {},
     sticky = true, // whether to use sticky header
+    columnColorFormatters,
   } = props;
 
   const [filters, setFilters] = useState(initialFilters);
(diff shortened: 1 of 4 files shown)


junlincc:
@ktmud would love your review and approval. 🙏

ktmud (approved):
Code LGTM. But I'm hoping we could update the control in `superset-frontend` to implement the color schemes using the existing colorScheme registry in superset-ui.

The Heatmap chart has example of sequential color schemes:



This should provide more consistency across Superset and add the ability to apply diverging colorschemes (i.e. green for positives, red for negatives).

It would also be nice to let users choose whether to apply the colors on the background or text.

kgabryje:
Thanks for the suggestions Jesse! I agree that there's still a lot to improve, but that was meant to be an MVP version - we'll build up on it in the future.

PR #1220, opened 2021-07-14
feat(plugin-chart-table): table x-filtering supports single select

stephenLYZ:
🏆 Enhancements

### before

https://user-images.githubusercontent.com/11830681/125647071-0b284574-5ae9-4821-a586-561001d2d8e3.mov

### after

https://user-images.githubusercontent.com/11830681/125647045-c8c3a122-de96-43d2-8336-2b1dbc3fdb07.mov

### DIFF
--- a/plugins/plugin-chart-table/src/TableChart.tsx
+++ b/plugins/plugin-chart-table/src/TableChart.tsx
@@ -232,11 +232,20 @@ export default function TableChart<D extends DataRecord = DataRecord>(
 
   const toggleFilter = useCallback(
     function toggleFilter(key: string, val: DataRecordValue) {
-      const updatedFilters = { ...(filters || {}) };
+      let updatedFilters = { ...(filters || {}) };
+      // mutil select
+      // if (filters && isActiveFilterValue(key, val)) {
+      //   updatedFilters[key] = filters[key].filter((x: DataRecordValue) => x !== val);
+      // } else {
+      //   updatedFilters[key] = [...(filters?.[key] || []), val];
+      // }
+      // single select
       if (filters && isActiveFilterValue(key, val)) {
-        updatedFilters[key] = filters[key].filter((x: DataRecordValue) => x !== val);
+        updatedFilters = {};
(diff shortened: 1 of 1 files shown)


ktmud:
Are we should this is what we want?

A lot of datasources would probably appreciate multi-select more than filtering only by one dimension & one value.

It some users have asked for single select, we should make the filters configurable instead of changing the behavior for all.

ktmud (changes requested):
Let's make this configurable or support Cmd/Ctr & Shift + Click.

rusackas:
At this moment, we're aiming for consistency, so a bunch of plugins are being moved to single-select. We can then do a small round of UX/UI design work to figure out the right interaction mechanism, affordance/indication, and means to clear one or all selects. Let's get simple cross-filtering working across he board, then step it up a notch across the board, rather than leaving it mix-and-match

plugins/plugin-chart-table/src/TableChart.tsx
```diff
@@ -232,11 +232,20 @@ export default function TableChart<D extends DataRecord = DataRecord>(
 
   const toggleFilter = useCallback(
     function toggleFilter(key: string, val: DataRecordValue) {
-      const updatedFilters = { ...(filters || {}) };
+      let updatedFilters = { ...(filters || {}) };
+      // mutil select
```
rusackas, line 236: Maybe worth a bit more of a note as to why these are here. Something like:

> Leaving this code here and commented out, for a future iteration where all (or at least most) plugins will support multi-select cross-filter emission.

rusackas (approved):

ktmud (approved):
I'd imagine the code be very different when we do add multi-select support so maybe the comments should be cleaned up instead.

stephenLYZ:
> I'd imagine the code be very different when we do add multi-select support so maybe the comments should be cleaned up instead.

Agreed, but I think it might be useful to implement multiple-select later. Here I'll remove the comments first.

PR #1221, opened 2021-07-14
feat(plugin-chart-pivot-table): pivot table x-filtering supports single select

stephenLYZ:
🏆 Enhancements

### before

https://user-images.githubusercontent.com/11830681/125649358-bb3b95e6-f9c0-4a0d-8f84-0fdcadf4eade.mov

### after

https://user-images.githubusercontent.com/11830681/125649377-0c44cfbe-6212-487a-90ff-03d2b16e4d36.mov

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -172,11 +172,20 @@ export default function PivotTableChart(props: PivotTableProps) {
 
       const [key, val] = filtersEntries[filtersEntries.length - 1];
 
-      const updatedFilters = { ...(selectedFilters || {}) };
+      let updatedFilters = { ...(selectedFilters || {}) };
+      // multi select
+      // if (selectedFilters && isActiveFilterValue(key, val)) {
+      //   updatedFilters[key] = selectedFilters[key].filter((x: DataRecordValue) => x !== val);
+      // } else {
+      //   updatedFilters[key] = [...(selectedFilters?.[key] || []), val];
+      // }
+      // single select
       if (selectedFilters && isActiveFilterValue(key, val)) {
-        updatedFilters[key] = selectedFilters[key].filter((x: DataRecordValue) => x !== val);
+        updatedFilters = {};
       } else {
-        updatedFilters[key] = [...(selectedFilters?.[key] || []), val];
+        updatedFilters = {
+          [key]: [val],
+        };
       }
       if (Array.isArray(updatedFilters[key]) && updatedFilters[key].length === 0) {
         delete updatedFilters[key];


ktmud (changes requested):
Similar to https://github.com/apache-superset/superset-ui/pull/1220, let's change the behavior to:

- Click to select a single filter
- Cmd/Ctrl + Click to add multiple filters (either filter values or dimensions)

junlincc:
@ktmud  we also considered 'Cmd/Ctrl + Click to add multiple filters' this behavior before...and decided not to implement it at this phase. We assume X-filtering is more cater for business users who need to quickly browse dashboard metric. Given the constraints that 1. we don't have a good UI design ready for 1) enabling emitting xfiltering behavior on dashboard 2) clearing x filtering selection 3) setting up emit and receive relationship between charts on dashboard, we chose to keep X-filtering behavior as simple as possible, and only allow single select.  
if advanced users want to add multiple filter values to further explore the data, they can always fall back on native filters. 
We sure will get to supporting multiple select soon, with a more user friendly UI approach. 🙏

ktmud:
Could you elaborate what are the additional UI changes needed for multi-select? I understand the technical challenge but I'm not sure which are the additional UI changes you need to make it work. Current design (click to toggle filter, cmd + click to toggle additional filters, show emitted values in filter indicators) seems work fine for me.

rusackas:
@ktmud I think you're quite right that the technical challenges are surmountable. I think the UI challenges are surmountable as well. That said, the main issue we're looking to bring to the experience is consistency. All crossfilters in charts should behave the same way, with a consistent UX/UI. For this iteration, we're attempting to take single-select as far and wide as we can, thus this PR. This work will bring the confidence we need to turn on the feature flag for our releases.

In a future iteration, after we have that consistency around a single-select experience, we want to address the design/UI of multi-select as a separate enhancement. For this to work, we need to make sure that:
* Multi select is enabled in all charts for a consistent experience (too ambitious for this iteration)
* Users have a clear/consistent means to clear one or all filter emitters (a clear button on the chart? Clear buttons in the filter indicators? Some other mechanism?) added consistently in all charts.
* Improving discoverability around how these cross filter relationships map from chart to chart, which might be more complex with multiple filters

Hope that makes sense, but happy to have a call and dive deeper.

ktmud:
@rusackas Thanks for the explanations. This makes sense.

FWIW, I don't think a clear button is necessary if users can just re-click to unselect.

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -172,11 +172,20 @@ export default function PivotTableChart(props: PivotTableProps) {
...
 
-      const updatedFilters = { ...(selectedFilters || {}) };
+      let updatedFilters = { ...(selectedFilters || {}) };
+      // multi select
+      // if (selectedFilters && isActiveFilterValue(key, val)) {
+      //   updatedFilters[key] = selectedFilters[key].filter((x: DataRecordValue) => x !== val);
+      // } else {
+      //   updatedFilters[key] = [...(selectedFilters?.[key] || []), val];
+      // }
+      // single select
```
ktmud, line 182: Can we clean up the comments?

ktmud (approved):

PR #1223, opened 2021-07-14
fix(legacy-plugin-chart-sankey-loop): clear element before redrawing

jenseng:
🐛 Bug Fix

Otherwise every (re)render adds a new svg to the element. This is essentially the same as the fix that was applied to legacy-plugin-chart-sankey back in apache/superset#254


villebro (approved):
LGTM, thanks for the fix!

PR #1225, opened 2021-07-15
fix: Reduce bundle sizes from react-icons

etr2460:
🏆 Enhancements

This should allow us to tree-shake and only pull in the icons that we actually use from react-icons. see here: https://github.com/react-icons/react-icons#installation-for-meteorjs-gatsbyjs-etc

This should help us reduce the bundle size in superset proper (see the massive react-icons/fa chunk here): 
image

Test plan:
Load up the storybook, and see icons in the table still render
Screen Shot 2021-07-15 at 8 25 12 PM

to: @ktmud @kgabryje

### DIFF
--- a/packages/superset-ui-chart-controls/package.json
+++ b/packages/superset-ui-chart-controls/package.json
@@ -26,6 +26,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@react-icons/all-files": "^4.1.0",
     "@superset-ui/core": "0.17.64",
     "lodash": "^4.17.15",
     "prop-types": "^15.7.2"
@@ -33,7 +34,6 @@
   "peerDependencies": {
     "@types/react": "*",
     "antd": "^4.9.4",
-    "react": "^16.13.1",
-    "react-icons": "^4.2.0"
+    "react": "^16.13.1"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/superset-ui-chart-controls/package.json
```diff
@@ -26,14 +26,14 @@
...
+    "@react-icons/all-files": "^4.1.0",
     "@superset-ui/core": "0.17.64",
     "lodash": "^4.17.15",
     "prop-types": "^15.7.2"
   },
   "peerDependencies": {
     "@types/react": "*",
     "antd": "^4.9.4",
-    "react": "^16.13.1",
-    "react-icons": "^4.2.0"
```
etr2460, line 37: note: moved from being a peer dep because i'm going to remove it from superset proper (it's not being used anywhere there)

ktmud:
Ah, I thought webpack treeshaking already only exports icons imported.

ktmud (approved):

kgabryje:
> Ah, I thought webpack treeshaking already only exports icons imported.

I suppose it's a similar case as with lodash - if we import from `lodash` instead of some subpackage like `lodash/debounce`, the whole thing is getting included in the bundle

rusackas:
Thanks for this @etr2460 and @ktmud  - happy to see the machete come out for these things. Looking forward to the "after" picture once this is merged into Superset.

PR #1226, opened 2021-07-16
feat(chart-controls): dnd version of time column

kgabryje:
Prerequisite: https://github.com/apache/superset/pull/15740

When `ENABLE_EXPLORE_DRAG_AND_DROP` is enabled, changes Time Column control to drag and drop. The behaviour is similar to the old Select component: 
- only temporal columns can be dropped
- the default value is either `main_dttm_col` defined in dataset or the first temporal column
- if there are no temporal columns in dataset, the control is left empty
- selection is not removable

In order to test:
- make sure that you are using the changes from https://github.com/apache/superset/pull/15740
- open a dataset with `main_dttm_col` set (e.g. channels from test data). `created` should be selected by default
- you can change the selection by drag and dropping other temporal columns. Non-temporal columns shouldn't be accepted by the control
- open a dataset with `main_dttm_col` unset, e.g. FCC 2018 Survey. `time_start` should be selected by default as it's a first temporal column
- open a dataset with no temporal columns, e.g. covid_vaccine. Time Column should be empty

https://user-images.githubusercontent.com/15073128/125951145-2b841317-647c-4eb8-b277-097861f24b72.mov

CC: @junlincc


zhaoyongjie (approved):
LGTM

PR #1231, opened 2021-07-22
fix: missing nan value in timeseries chart

zhaoyongjie:
currently, the time-series line chart render `null` value as 0. this PR fix that.

A quick fix for the time-series chart. 

I will clean up all the operators when the advanced analytics merge into master.

🐛 Bug Fix
### Before
image

### After
image


kgabryje (approved):
LGTM

jinghua-qa:
Verify fixed in master, null value is shown in time series chart. Thank you for fixing this issue!
Screen Shot 2021-07-22 at 3 19 06 PM

PR #1232, opened 2021-07-22
feat(plugin-chart-pivot-table): multiple improvements

kgabryje:
List of changes:
1. Header and totals row are sticky
2. Added aggregation function name to "Total" label
3. Changed labels from "Totals" to "Total"
4. Moved "Show total" checkboxes to Data tab
5. "Show total" checkboxes are unchecked by default

https://user-images.githubusercontent.com/15073128/126642403-c2f4bc96-77e9-49e4-8a2c-fa033df06ff6.mov

CC: @junlincc @rusackas

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -24,6 +24,7 @@ import {
   getNumberFormatter,
   DataRecordValue,
   NumberFormatter,
+  useTheme,
 } from '@superset-ui/core';
 // @ts-ignore
 import PivotTable from '@superset-ui/react-pivottable/PivotTable';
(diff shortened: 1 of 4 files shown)


plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -39,9 +40,9 @@ import {
...
 
 const Styles = styled.div<PivotTableStylesProps>`
-  padding: ${({ theme }) => theme.gridUnit * 4}px;
-  height: ${({ height }) => height}px;
-  width: ${({ width }) => width}px;
+  margin: ${({ margin }) => margin}px;
+  height: ${({ height, margin }) => height - margin * 2}px;
+  width: ${({ width, margin }) => width - margin * 2}px;
   overflow-y: scroll;
   }
```
michael-s-molina, line 47:
```suggestion
  ${({ height, width, margin }) => `
      margin: ${margin}px;
      height: ${height - margin * 2}px;
      width: ${width - margin * 2}px;
      overflow-y: scroll;
 `}
```

zhaoyongjie (approved):
the codes LGTM! thanks, Kamil.

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
kgabryje: Done, thanks!

jaimefr:
@kgabryje thanks for the great contribution. One question, why does sticky was only applied to the "top row header" and not to the "left column header"?

PR #1239, opened 2021-07-26
fix(plugin-chart-echarts): make filtered pie slices semi-transparent

villebro:
🐛 Bug Fix

Currently unselected slices are fully transparent. This changes the opacity of filtered slices to `SemiTransparent` (=0.3) instead of `Transparent` (0).

### AFTER
image

### BEFORE
image

Closes https://github.com/apache/superset/issues/15832


zhaoyongjie (approved):
LGTM. Thanks for the fix.

PR #1240, opened 2021-07-26
feat(plugin-chart-pivot-table): sort by metric

kgabryje:
Implement sorting query results by metric, descending (default) or ascending. If no sort by metric is specified, query results will be sorted by the first adhoc metric, descending (the same behaviour as in Table chart for consistency).

Testing instructions:
1. Create a pivot table chart
2. Use a saved metric or adhoc metric in "Sort by" control. I also suggest setting a low row limit (e.g. 10) to see the effect of sort by clearly.

https://user-images.githubusercontent.com/15073128/126995425-95ccfeb6-05be-4139-93e7-883eab3f59ff.mov

CC: @junlincc 

closes https://github.com/apache/superset/issues/15883#issuecomment-885968829

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
+++ b/plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
@@ -16,19 +16,30 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { buildQueryContext, ensureIsArray } from '@superset-ui/core';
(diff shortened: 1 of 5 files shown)


zhaoyongjie:
There is a helper function to nomalize orderby. you can check this https://github.com/apache-superset/superset-ui/blob/master/packages/superset-ui-core/src/query/normalizeOrderBy.ts

zhaoyongjie (approved):
LGTM! Thanks for this change.

PR #1242, opened 2021-07-27
chore(core): move emotion deps to peer deps

kgabryje:
Move `@emotion/cache` and `@emotion/react` from dependencies to peer dependencies in order to fix build errors when linking superset-ui-core to superset locally


villebro (approved):
This will make life a 🏖️ !

zhaoyongjie (approved):
Awesome!

PR #1243, opened 2021-07-27
feat(plugin-chart-pivot-table): add sort by result to data pane

kgabryje:
Adds "Sort by" metric's values to south data table (below the chart).

image

CC @junlincc

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
+++ b/plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
@@ -16,21 +16,38 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { buildQueryContext, ensureIsArray, normalizeOrderBy } from '@superset-ui/core';
+import { AdhocMetric, buildQueryContext, ensureIsArray, normalizeOrderBy } from '@superset-ui/core';
 import { PivotTableQueryFormData } from '../types';
 
 export default function buildQuery(formData: PivotTableQueryFormData) {
-  const { groupbyColumns = [], groupbyRows = [], order_desc = true } = formData;
+  const {
+    groupbyColumns = [],
+    groupbyRows = [],
+    order_desc = true,
+    timeseries_limit_metric,
+  } = formData;
   const groupbySet = new Set([
     ...ensureIsArray<string>(groupbyColumns),
     ...ensureIsArray<string>(groupbyRows),
   ]);
   return buildQueryContext(formData, baseQueryObject => {
     const queryObject = normalizeOrderBy({ ...baseQueryObject, order_desc });
+    const { metrics } = queryObject;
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
```diff
@@ -16,21 +16,38 @@
...
+    const orderBy = ensureIsArray(timeseries_limit_metric);
+    if (
+      orderBy.length > 0 &&
+      ((typeof orderBy[0] === 'string' && !metrics?.includes(orderBy[0])) ||
+        !metrics?.find(
+          metric => (metric as AdhocMetric).label === (orderBy[0] as AdhocMetric).label,
+        ))
+    ) {
+      metrics?.push(orderBy[0]);
+    }
```
zhaoyongjie, line 45: Why do we use orderby clause append to metrics?

plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
```diff
@@ -16,21 +16,38 @@
...
   const groupbySet = new Set([
     ...ensureIsArray<string>(groupbyColumns),
     ...ensureIsArray<string>(groupbyRows),
   ]);
   return buildQueryContext(formData, baseQueryObject => {
     const queryObject = normalizeOrderBy({ ...baseQueryObject, order_desc });
+    const { metrics } = queryObject;
+    const orderBy = ensureIsArray(timeseries_limit_metric);
+    if (
+      orderBy.length > 0 &&
```
villebro, line 38: nit:
```suggestion
      orderBy.length &&
```

plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
kgabryje: We want to append the values of metric that we sort by to the data table (below the chart)

villebro, line 45: Can we use `getMetricLabel` here? https://github.com/apache-superset/superset-ui/blob/9b92a69d8eeb1c31548ce74a0538e70901c5bf26/packages/superset-ui-core/src/query/getMetricLabel.ts#L3-L14

villebro (commented):
I recall there being discussion about this a while back (always adding orderbys to the select), but I'm not sure what the outcome was. But I'm ok with this until we decide one way or another.

plugins/plugin-chart-pivot-table/src/plugin/buildQuery.ts
kgabryje: Thank you, there are so many nice helper functions that I'm not aware of 😄

zhaoyongjie (approved):
The code, LGTM

PR #1244, opened 2021-07-28
feat(plugin-chart-echarts): implement x-filter opacity in ts chart

villebro:
🏆 Enhancements

Add proper cross filter highlighting to TimeSeries chart and make behavior more uniform with other charts. Other minor improvements:
- Weight the area chart opacity with the line opacity (see screenshot in after/before)
- Make event code more in-line with other charts (these will be generalized further in a follow-up)
- ABC order props
- remove unnecessary type cast

### AFTER
https://user-images.githubusercontent.com/33317356/127293407-8ed7bce5-7a1b-4ccb-8dfa-c3dd73764dd1.mp4

Area chart opacity is also weighed based on the line opacity. If area opacity is 0.5 and line opacity is 0.3 (`SemiTransparent`), then area opacity becomes 0.5 x 0.3 = 0.5:
image

### BEFORE
https://user-images.githubusercontent.com/33317356/127293456-d9504fb5-f10c-45cc-bf89-7a044bcb5cf3.mp4


plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
```diff
@@ -113,14 +119,14 @@ export function transformSeries(
   } else {
     plotType = seriesType === 'bar' ? 'bar' : 'line';
   }
-  const lineStyle = isConfidenceBand ? { opacity: 0 } : {};
-
+  const lineStyle = isConfidenceBand ? { opacity: 0 } : { opacity };
```
stephenLYZ, line 122: I think we can use `OpacityEnum.Transparent ` to represent `0`.

plugins/plugin-chart-echarts/src/Timeseries/types.ts
```diff
@@ -116,12 +116,13 @@ export interface EchartsTimeseriesChartProps extends ChartProps {
...
 
 export interface TimeseriesChartTransformedProps {
-  formData: EchartsTimeseriesFormData;
-  height: number;
-  width: number;
   echartOptions: EChartsOption;
   emitFilter: boolean;
-  setDataMask: SetDataMaskHook;
-  labelMap: Record<string, DataRecordValue[]>;
+  formData: EchartsTimeseriesFormData;
```
stephenLYZ, line 121: This part of the type seems to be the same in every echarts except the formdata declaration, so I abstracted an interface in `plugin-chart-echarts/src/types.ts`,  maybe we can reuse it?

stephenLYZ (approved):
LGTM! Thanks for the jobs. Just leave some suggestions on types.

plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
villebro: Thanks, good catch!

plugins/plugin-chart-echarts/src/Timeseries/types.ts
villebro: Great improvement! 👍

zhaoyongjie (approved):
LGTM!

PR #1245, opened 2021-07-28
chore: only keey Distribution category

stephenLYZ:
🏆 Enhancements
Distribution is a category and a tag,  we just keep the category.


suddjian (approved):
thanks!

PR #1248, opened 2021-07-29
feat(chart-controls): move cross-filtering control to data panel

stephenLYZ:
🏆 Enhancements
- Move the cross filter control from the "Customize" tab to the "Query" tab under the "(adhoc) filter" control.
- Rename from "Enable emitting filters" to "Emit dashboard cross filters"

releated: https://github.com/apache/superset/issues/15910


plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
```diff
@@ -37,6 +35,7 @@ export default {
       controlSetRows: [
         ['metrics'],
         ['adhoc_filters'],
+        emitFilterControl,
```
zhaoyongjie, line 38: We should use `sections.emitFilterControl` for consistency namespace.

plugins/plugin-chart-echarts/src/Funnel/controlPanel.tsx
```diff
@@ -43,6 +44,7 @@ const config: ControlPanelConfig = {
         ['groupby'],
         ['metric'],
         ['adhoc_filters'],
+        emitFilterControl,
```
zhaoyongjie, line 47: same before.

plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
villebro: I think this is ok - IMO `sections`  should be reserved for `ControlPanelSectionConfig` objects, not `ControlSetRow` objects.

villebro (approved):
LGTM, but it seems you need to add a unit test for when the feature flag is set and unset (codecov needs to be 100 % for the modules under `packages/`).

PR #1250, opened 2021-07-29
chore: remove V2 and add Deprecated

stephenLYZ:
🏆 Enhancements
image

image


plugins/legacy-plugin-chart-pivot-table/src/index.js
```diff
@@ -26,7 +26,7 @@ const metadata = new ChartMetadata({
   description: t(`Used to summarize a set of data by grouping together multiple statistics along two axes. Examples: Sales numbers by region and month, tasks by status and assignee, active users by age and location.
 
   This chart is being deprecated and we recommend checking out Pivot Table V2 instead!`),
```
suddjian, line 28:
```suggestion
  This chart is being deprecated and we recommend checking out the new Pivot Table chart instead!`),
```

suddjian (approved):

PR #1251, opened 2021-07-29
fix: add feature flags for pagination control in table chart

pkdotson:
💔 Breaking Changes

🏆 Enhancements
 
📜 Documentation

🐛 Bug Fix
This Pr fixes an issue where server pagination is enable in table chart but without feature flags native filter or dashboard cross filters not enabled. Feature flags cross filters or native filters must be enabled to use table pagination.

🏠 Internal


villebro (approved):
LGTM!

PR #1256, opened 2021-07-31
feat(plugin-chart-echarts): implement cross filter in mixd-timeseries chart

stephenLYZ:
🏆 Enhancements
Add cross filter highlighting to MixdTimeSeries chart.

Both series support setting cross filter separately.
image

https://user-images.githubusercontent.com/11830681/127747100-889620ec-7ece-43f8-b9c1-6577a6e1ccf1.mov

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
@@ -16,10 +16,98 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import React from 'react';
-import { EchartsProps } from '../types';
+import React, { useCallback } from 'react';
+import { EchartsMixedTimeseriesChartTransformedProps } from './types';
 import Echart from '../components/Echart';
+import { EventHandlers } from '../types';
 
-export default function EchartsMixedTimeseries({ height, width, echartOptions }: EchartsProps) {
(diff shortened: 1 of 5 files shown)


plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
```diff
@@ -16,10 +16,92 @@
...
+  setDataMask,
+  labelMap,
+  labelMapB,
+  groupby,
+  groupbyB,
+  selectedValues,
+  formData,
+  seriesBreakdown,
+}: EchartsMixedTimeseriesChartTransformedProps) {
+  const isFirstQuery = useCallback((seriesIndex: number) => seriesIndex < seriesBreakdown, [
```
stephenLYZ, line 37: Here we need to determine which rawSeriesData was clicked (rawSeriesA or rawSeriesB)

villebro (approved):
LGTM and tested to work well! The only thing that's slightly confusing is that we have two different places where we need to enable x-filtering, but that's more due to the plugin implementation, not this feature. Some linting issues still need to be fixed to be able to remove the `@ts-ignore`s, but we can do that in a follow-up PR.

zhaoyongjie:
Tested in my local. Looks good!

https://user-images.githubusercontent.com/2016594/127811048-a65195ca-b281-466c-b271-6c266fc98e5f.mov

zhaoyongjie (approved):
LGTM

PR #1259, opened 2021-08-02
fix(legacy-preset-chart-nvd3): dual line dnd control missing

kgabryje:
1 of the controls in Dual Line chart was still using the old UI, even with DND feature flag on. This PR fixes it.
I also moved some controls to separate rows to avoid ugly overflow.

Before: https://github.com/apache/superset/issues/16018
After:
image

CC: @jinghua-qa @junlincc


villebro (approved):
LGTM 👍

PR #1264, opened 2021-08-03
feat(plugin-chart-pivot-table): add automatic conditional formatter

villebro:
🏆 Enhancements

Add a new conditional formatting operator to enable automatic formatting. This is a non-breaking change, as it only adds support for a new option.

### SCREENSHOT

image


kgabryje (approved):
Lgtm!

PR #1265, opened 2021-08-03
fix(plugin-chart-pivot-table): dont display nulls as 0

kgabryje:
Before, we were displaying nulls as zeros when using "Sum" aggregate function. Now we filter out the nulls before passing them to the pivot table, so they are not included in the calculations.

Before: see https://github.com/apache/superset/issues/15938
After:
image

CC @junlincc @jinghua-qa


villebro (approved):
LGTM!

PR #1267, opened 2021-08-04
feat(legacy-plugin-chart-country-map): add austria map to country map charts

Anushka-Hapukotuwa:
🏆 Enhancements
Austria map added to country map chart


villebro (approved):
Vielen dank @Anushka-Hapukotuwa 🇦🇹 !

PR #1270, opened 2021-08-04
fix(plugin-chart-echarts): fix null labels on pie and funnel charts

villebro:
🐛 Bug Fix

Remove label escaping in Pie and Funnel charts that were causing hml escaping where it wasn't needed. Closes https://github.com/apache/superset/issues/16057

### AFTER
image

### BEFORE
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
@@ -50,13 +50,15 @@ export function formatFunnelLabel({
   params,
   labelType,
   numberFormatter,
+  sanitizeName = false,
 }: {
-  params: CallbackDataParams;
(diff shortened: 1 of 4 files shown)


plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
```diff
@@ -50,13 +50,15 @@ export function formatFunnelLabel({
   params,
   labelType,
   numberFormatter,
+  sanitizeName = false,
 }: {
-  params: CallbackDataParams;
+  params: Pick<CallbackDataParams, 'name' | 'value' | 'percent'>;
```
villebro, line 55: Made more specific to fix linting error in UTs

stephenLYZ:
LGTM! Thanks for the fix !

kgabryje (approved):
LGTM

rosemarie-chiu:
🏷 2021.31

PR #1272, opened 2021-08-04
chore: stacks and sorts Heatmap controls

michael-s-molina:
🏆 Enhancements
Stacks and sorts Heatmap controls to a better UI/UX. The proposed change targets a more uniform display of the information for the user and gives the controls more space to present the UI elements. The idea is also to avoid, as much as we can, the expansion of the controls section, preserving the available space for displaying the charts.

Before:

https://user-images.githubusercontent.com/70410625/128224568-28aee072-f59a-45d1-857c-95def7ef4b76.mov

After:

https://user-images.githubusercontent.com/70410625/128225226-a005d21b-c359-48ce-a4d3-a976526c8e8e.mov

### DIFF
--- a/plugins/legacy-plugin-chart-heatmap/src/controlPanel.ts
+++ b/plugins/legacy-plugin-chart-heatmap/src/controlPanel.ts
@@ -112,6 +112,8 @@ const config: ControlPanelConfig = {
               description: t('Number of steps to take between ticks when displaying the X scale'),
             },
           },
+        ],
+        [
           {
             name: 'yscale_interval',
             config: {
@@ -143,6 +145,8 @@ const config: ControlPanelConfig = {
               ),
             },
           },
+        ],
+        [
           {
             name: 'normalize_across',
             config: {
@@ -176,6 +180,8 @@ const config: ControlPanelConfig = {
               description: t('Left margin, in pixels, allowing for more room for axis labels'),
             },
           },
+        ],
+        [
           {
             name: 'bottom_margin',
             config: {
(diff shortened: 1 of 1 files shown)


plugins/legacy-plugin-chart-heatmap/src/controlPanel.ts
```diff
@@ -204,7 +210,31 @@ const config: ControlPanelConfig = {
...
+        ['y_axis_format'],
+        [
+          {
+            name: 'sort_x_axis',
+            config: {
+              type: 'SelectControl',
+              label: t('Sort X Axis'),
+              choices: sortAxisChoices,
+              clearable: false,
+              default: 'alpha_asc',
```
villebro, line 223: I think we can add `renderTrigger: true` here (it shouldn't affect the query based on my tests)

plugins/legacy-plugin-chart-heatmap/src/controlPanel.ts
```diff
@@ -204,7 +210,31 @@ const config: ControlPanelConfig = {
...
+        ],
+        [
+          {
+            name: 'sort_y_axis',
+            config: {
+              type: 'SelectControl',
+              label: t('Sort Y Axis'),
+              choices: sortAxisChoices,
+              clearable: false,
+              default: 'alpha_asc',
```
villebro, line 235: same here

villebro (approved):
LGTM with one small improvement proposal (non-blocking as it's outside the scope of this PR)

PR #1276, opened 2021-08-05
fix(legacy-preset-chart-nvd3): bar chart unexpected error

kgabryje:
Bar chart doesn't allow "Series" and "Breakdown" controls to overlap. If user adds the same column to those controls, we get "Unexpected error". This PR disallows adding a column to "Breakdown" if it's already in "Series" (and the other way around too) when drag and drop is enabled.

Requires https://github.com/apache/superset/pull/16090 to work

Fixes https://github.com/apache/superset/issues/16071

https://user-images.githubusercontent.com/15073128/128378703-bbebd2ae-a948-4fce-a0bf-bd61cf7c8b1a.mov

CC @junlincc @jinghua-qa

### DIFF
--- a/plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
+++ b/plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
@@ -16,8 +16,13 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import { t, validateNonEmpty } from '@superset-ui/core';
-import { ControlPanelConfig, sections } from '@superset-ui/chart-controls';
+import { ensureIsArray, t, validateNonEmpty } from '@superset-ui/core';
+import {
+  ColumnMeta,
+  ControlPanelConfig,
+  sections,
+  sharedControls,
+} from '@superset-ui/chart-controls';
 import {
   showLegend,
   showControls,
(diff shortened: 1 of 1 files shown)


plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
```diff
@@ -107,10 +112,24 @@ const config: ControlPanelConfig = {
     groupby: {
       label: t('Series'),
       validators: [validateNonEmpty],
+      mapStateToProps: (state, controlState) => {
+        const groupbyProps = sharedControls.groupby.mapStateToProps?.(state, controlState) || {};
+        groupbyProps.canDropValue = (column: ColumnMeta) =>
+          !ensureIsArray(state.controls.columns.value).includes(column.column_name);
```
villebro, line 118: Question: Is it safe to assume the whole chain `state.controls.columns.value` is defined? Feels slightly safer to put `?` in between those nested properties..

plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
```diff
@@ -107,10 +112,24 @@ const config: ControlPanelConfig = {
...
+      },
+      rerender: ['columns'],
     },
     columns: {
       label: t('Breakdowns'),
       description: t('Defines how each series is broken down'),
+      mapStateToProps: (state, controlState) => {
+        const columnsProps = sharedControls.columns.mapStateToProps?.(state, controlState) || {};
+        columnsProps.canDropValue = (column: ColumnMeta) =>
+          !ensureIsArray(state.controls.groupby.value).includes(column.column_name);
```
villebro, line 129: Same here

villebro (commented):
LGTM - One simple question before hitting approve

plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
kgabryje: Done

plugins/legacy-preset-chart-nvd3/src/DistBar/controlPanel.ts
kgabryje: Done

villebro (approved):

PR #1277, opened 2021-08-05
fix: Match Pivot Table V2 style with Pivot Table

geido:
### SUMMARY

This PR changes the border color, font family and font size of the Pivot Table V2 to match those of the original Pivot Table

Fixes: https://github.com/apache/superset/issues/15826

### BEFORE
Screen Shot 2021-07-21 at 2 16 20 PM

### AFTER

Screen Shot 2021-08-09 at 17 36 32

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -40,11 +40,32 @@ import {
 } from './types';
 
 const Styles = styled.div<PivotTableStylesProps>`
-  ${({ height, width, margin }) => `
+  ${({ height, width, margin, borderColor }) => `
       margin: ${margin}px;
       height: ${height - margin * 2}px;
       width: ${width - margin * 2}px;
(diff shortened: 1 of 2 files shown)


ktmud:
The font and cell paddings are also a little inconsistent with a regular table vis. Could you optimize that as well?

junlincc:
> The font and cell paddings are also a little inconsistent with a regular table vis. Could you optimize those as well?

@ktmud thanks for catching! you have 🦅 👀....

geido:
This should be ready for another round of reviews. Thank you

kgabryje:
Thanks for the change. You can see on the screenshot that some borders are grey, but some still have the old blueish colour. It will be easier to change the css in react-pivottable than to overwrite the styles in superset-ui. Would you mind if I made that change and pushed bumped react-pivottable version to this PR?

geido:
I wasn't aware that we have forked react-pivottable. Of course, making the changes on the fork would make much more sense. We can then discard this PR.

PR #1278, opened 2021-08-06
feat(legacy-preset-chart-big-number): add timestamp above number

krsnik93:
💔 Breaking Changes: No

🏆 Enhancements: add timestamp above the number in BigNumber chart(s)

image

📜 Documentation: NA

🐛 Bug Fix: NA

🏠 Internal: NA

### DIFF
--- a/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
+++ b/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
@@ -17,7 +17,13 @@
  * under the License.
  */
 import { t } from '@superset-ui/core';
-import { ControlPanelConfig, formatSelectOptions, sections } from '@superset-ui/chart-controls';
+import {
+  ControlPanelConfig,
+  D3_FORMAT_DOCS,
+  D3_TIME_FORMAT_OPTIONS,
+  formatSelectOptions,
+  sections,
+} from '@superset-ui/chart-controls';
 import React from 'react';
 import { headerFontSize, subheaderFontSize } from '../sharedControls';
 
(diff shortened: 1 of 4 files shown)


plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
```diff
@@ -56,6 +56,18 @@ const config: ControlPanelConfig = {
...
+            name: 'show_timestamp',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Timestamp'),
+              renderTrigger: true,
+              default: false,
+              description: t('Whether to display the timestamp'),
+            },
+          },
+        ],
```
villebro, line 70: Could we also add a date formatter control that is made visible if `show_timestamp` is `true`? Here's a time formatter control from the ECharts timeseries charts: 
image

villebro (commented):
LGTM - let's just add a time formatter to make it possible to customize the time format

plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
krsnik93: added!

villebro:
I'm getting some weird behavior when applying smart date formatting.With adaptive formatter it looks like this:
image

And with some of the other formats it looks like this:
image

To make this behavior more intuitive maybe we could borrow some of the logic from table chart:
https://github.com/apache-superset/superset-ui/blob/1c9baeeb6f03e05ba797452dc463eb8def7996bc/plugins/plugin-chart-table/src/transformProps.ts#L124-L140

villebro (approved):
Awesome, thanks for quickly iterating on the review comments!

junlincc:
thanks for the improvement! a new control is added?

villebro:
> thanks for the improvement! a new control is added?

Yes - two new ones in fact. One for the checkbox "Show timestamp" (defaults to false), and if that is checked a date format select is displayed to customize the format.

PR #1279, opened 2021-08-08
feat(plugin-chart-echarts): Add support for series values to ECharts timeseries charts

stephenLYZ:
🏆 Enhancements
related to https://github.com/apache/superset/issues/16063

- Add a new control call `Show Value`. 
- Support show series values to all echarts timeseries chart.

Unstacked bars:
image

Stacked bars:
image

Unstacked area:
image

Stacked area:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
@@ -32,7 +32,7 @@ import {
   EchartsTimeseriesContributionType,
   EchartsTimeseriesSeriesType,
 } from '../types';
-import { legendSection } from '../../controls';
+import { legendSection, showValueControl } from '../../controls';
 
 const {
   annotationLayers,
@@ -56,7 +56,6 @@ const {
   yAxisBounds,
   zoomable,
   xAxisLabelRotation,
-  showValue,
 } = DEFAULT_FORM_DATA;
 const config: ControlPanelConfig = {
   controlPanelSections: [
@@ -252,18 +251,7 @@ const config: ControlPanelConfig = {
             },
           },
         ],
-        [
-          {
-            name: 'show_value',
-            config: {
-              type: 'CheckboxControl',
-              label: t('Show Value'),
-              default: showValue,
-              renderTrigger: true,
-              description: t('Show the value on top of the chart'),
-            },
-          },
-        ],
+        [showValueControl],
         [
           {
             name: 'stack',
--- a/plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
@@ -56,6 +56,7 @@ const {
   yAxisBounds,
   zoomable,
   xAxisLabelRotation,
+  showValue,
 } = DEFAULT_FORM_DATA;
 const config: ControlPanelConfig = {
   controlPanelSections: [
@@ -248,6 +249,18 @@ const config: ControlPanelConfig = {
               step: 0.1,
               default: opacity,
               description: t('Opacity of Area Chart. Also applies to confidence band.'),
+            },
+          },
+        ],
+        [
+          {
+            name: 'show_value',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Value'),
+              default: showValue,
+              renderTrigger: true,
+              description: t('Show the value on top of the chart'),
             },
           },
         ],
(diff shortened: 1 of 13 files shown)


plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -105,6 +106,31 @@ export default function transformProps(
...
+        return prev + (value as number);
+      }, 0);
+      totalStackedValues.push(values);
+    });
+
+    rawSeries.forEach((entry, seriesIndex) => {
+      const { data = [] } = entry;
+      (data as [Date, number][]).forEach((datum, dataIndex) => {
+        if (datum[1] !== null) {
+          showValueIndexes[dataIndex] = seriesIndex;
```
stephenLYZ, line 128: `showValueIndexes` records the highest series index (is not a null value) of each data, so that we can show the total value at the top.

plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
```diff
@@ -141,10 +162,29 @@ export function transformSeries(
...
+        const {
+          value: [_, numericValue],
+          dataIndex,
+          seriesIndex,
+        } = params;
+        if (formatter) {
+          if (!stack) {
+            return formatter(numericValue);
+          }
+          if (seriesIndex === showValueIndexes[dataIndex]) {
```
stephenLYZ, line 180: Here is the key to show the stack value, only if the series index is equal to our `showValueIndexes` will we show the value, otherwise we return the empty string.

plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
```diff
@@ -251,6 +252,18 @@ const config: ControlPanelConfig = {
...
+            name: 'show_value',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Value'),
+              default: showValue,
+              renderTrigger: true,
+              description: t('Show the value on top of the chart'),
+            },
+          },
+        ],
```
villebro, line 266: To avoid duplication, we should probably place this in `controls.ts` and then reuse it in all charts where it's needed

plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
```diff
@@ -251,6 +252,18 @@ const config: ControlPanelConfig = {
...
         ],
+        [
+          {
+            name: 'show_value',
+            config: {
+              type: 'CheckboxControl',
+              label: t('Show Value'),
+              default: showValue,
+              renderTrigger: true,
+              description: t('Show the value on top of the chart'),
```
villebro, line 263: Could we rephrase this to something like "Show series values on the chart" (on top of the chart might make it sound like it's on top of the whole chart rectangle).

plugins/plugin-chart-echarts/src/Timeseries/Area/index.ts
```diff
@@ -30,7 +30,7 @@ const areaTransformProps = (chartProps: EchartsTimeseriesChartProps) =>
     formData: { ...chartProps.formData, area: true },
   });
 
-export default class EchartsTimeseriesScatterChartPlugin extends ChartPlugin<
+export default class EchartsAreaChartPlugin extends ChartPlugin<
```
villebro, line 33: Whoops! 😆  Thanks for fixing!

villebro (commented):
This is really great work @stephenLYZ! A few minor comments. Also, it would be great if we could add a checkbox for "Show only total" that is visible if `show_value` is enabled that toggles between either only showing the stacked total value or the individual values. So if we have the following series:
- bicycles: 100
- cars: 200

In this case:
- If we `show_value` is unchecked, `only_total` control isn't displayed at all
- If we only check `show_value` but not `only_total`, we would see 100 next to bicycles and 200 next to cars.
- If we check both `sow_value` and `only_total`, we would see 300 next to whichever value is on top of the stack.

plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
stephenLYZ: You are right!

plugins/plugin-chart-echarts/src/Timeseries/Area/controlPanel.tsx
stephenLYZ: > This is really great work @stephenLYZ! A few minor comments. Also, it would be great if we could add a checkbox for "Show only total" that is visible if `show_value` is enabled that toggles between either only showing the stacked total value or the individual values. So if we have the following series:
> 
> * bicycles: 100
> * cars: 200
> 
> In this case:
> 
> * If we `show_value` is unchecked, `only_total` control isn't displayed at all
> * If we only check `show_value` but not `only_total`, we would see 100 next to bicycles and 200 next to cars.
> * If we check both `sow_value` and `only_total`, we would see 300 next to whichever value is on top of the stack.

Interesting~ I had previously thought that total value was only for stack scenarios. 😂. I think I can put it for the next task.

villebro (approved):
LGTM!

PR #1282, opened 2021-08-09
fix: 0 indicator on radar viz

zhaoyongjie:
🐛 Bug Fix
closes: https://github.com/apache/superset/issues/16120

### before
image

### after
image

### Test instructions
- Select `cleaned_sales_data` dataset
- Open explore page and switch to Radar Chart
- Add sum(sales), avg(sales) and sum(0) metrics into Chart
- Run query, the sum(0) metric at the center of polar coordinates.


villebro (approved):
LGTM, thanks for fixing!

PR #1283, opened 2021-08-09
fix: Hide Safari default tooltip

geido:
### SUMMARY
This PR implements a small hack to prevent Safari from showing its default tooltip when a text is truncated and a custom Antdesign tooltip is already present.

### RELATED PR ON SUPERSET
https://github.com/apache/superset/pull/16145

### BEFORE
Screen Shot 2021-08-09 at 16 50 59

### AFTER
Screen Shot 2021-08-09 at 16 48 47

### DIFF
--- a/packages/superset-ui-chart-controls/package.json
+++ b/packages/superset-ui-chart-controls/package.json
@@ -26,6 +26,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@emotion/react": "^11.4.1",
     "@react-icons/all-files": "^4.1.0",
     "@superset-ui/core": "0.17.75",
     "lodash": "^4.17.15",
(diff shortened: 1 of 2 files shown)


packages/superset-ui-chart-controls/package.json
```diff
@@ -26,6 +26,7 @@
     "access": "public"
   },
   "dependencies": {
+    "@emotion/react": "^11.4.1",
```
rusackas, line 29: We should import this as a peer dependency. We should also probably either use `^11.1.5` or consider making this version upgrade everywhere.

I think we also looked at just exporting Global from `@superset-ui/core` but I think there was something that got in the way of doing that last time we tried.

geido: Hey @rusackas. Thanks for your input. I do agree that we should export it from the core. It makes sense. For the sake of this PR, I have just made your suggested changes. Please have a look

kgabryje (approved):
Looks good 👍

PR #1285, opened 2021-08-09
chore: change Highly-used back to Popular

junlincc:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

Addressing requests to change "Highly Used" tag to "Popular" as was originally implemented. The version bump for these packages will need to coincide with a similar change on the Superset side.

🐛 Bug Fix

🏠 Internal

### DIFF
--- a/plugins/legacy-plugin-chart-world-map/src/index.js
+++ b/plugins/legacy-plugin-chart-world-map/src/index.js
@@ -39,7 +39,7 @@ const metadata = new ChartMetadata({
     t('Multi-Layers'),
     t('Multi-Variables'),
     t('Scatter'),
-    t('Highly-used'),
+    t('Popular'),
   ],
   thumbnail,
   useLegacyApi: true,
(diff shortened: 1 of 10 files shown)


plugins/legacy-plugin-chart-world-map/src/index.js
```diff
@@ -39,7 +39,7 @@ const metadata = new ChartMetadata({
     t('Multi-Layers'),
     t('Multi-Variables'),
     t('Scatter'),
-    t('Highly-used'),
+    t('Popular'),
```
villebro, line 42: We really need to change these to enums..

villebro (approved):
LGTM with yet another rant about refactoring the strings to enums 😄

ktmud:
While I fully support this change (because "Popular" sounds more colloquial and friendlier), it would be nice to note where does this "request" come from and why was "highly-used" introduced in the first place.

PR #1297, opened 2021-08-10
chore(core): add ENABLE_DND_WITH_CLICK_UX ff

kgabryje:
Add ENABLE_DND_WITH_CLICK_UX in order to hide features introduced in https://github.com/apache/superset/pull/16119 behind a feature flag


villebro (approved):
LGTM, although I personally prefer to make this part of the regular dnd feature

PR #1303, opened 2021-08-12
fix(plugin-chart-echarts): fix temporary filters

### LINKED ISSUE #15858: [chart][mixed time series chart] Mixed time series chart only show dashboard filter from query A
#### How to reproduce the bug

1. Create a mixed time-series chart, a filter box, and add them into a dashboard. see sample dashboard below.
2. apply filter in the dashboard, you will see filter is applied to both queries in the mixed time-series chart.
AYgZr9z3oV

3. open explore view from dashboard,
4. See filter carried from dashboard: only Query A has correct filter value, but not in Query B:

Screen_Shot_2021-07-22_at_11_42_23_PM

### Environment

latest master

### Checklist

Make sure to follow these steps before submitting your issue - thank you!

- [x] I have checked the superset logs for python stacktraces and included it here as text if there are any.
- [x] I have reproduced the issue with at least the latest released version of superset.
- [x] I have checked the issue tracker for the same issue and I haven't found one similar.

villebro:
🐛 Bug Fix

When going from the dashboard to Explore view with extra filters (native filters or filter box), the filters are only appended to the first query. In this PR we add the first query's adhoc filters that have `isExtra` set to true. In addition, the `showMinLabel` and `showMaxLabel` contrls are removed, as they're no longer used in the main Timeseries chart.

Tested to work both with native filters and filter box.

Closes https://github.com/apache/superset/issues/15858

### AFTER 
Notice how the same temporary filter is available in both Query A and Query B
image

### BEFORE
Notice how the temporary filter is missing from Query B:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -73,7 +73,7 @@ function createQuerySection(label: string, controlSuffix: string): ControlPanelS
           config: {
             ...sharedControls.adhoc_filters,
             mapStateToProps: (state, controlState) => ({
-              ...(sharedControls?.adhoc_filters?.mapStateToProps?.(state, controlState) || {}),
+              ...(sharedControls.adhoc_filters?.mapStateToProps?.(state, controlState) || {}),
               value: [
                 ...((state?.controls?.adhoc_filters?.value as any[]) || []).filter(
                   filter => filter.isExtra,
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
@@ -72,17 +72,15 @@ function createQuerySection(label: string, controlSuffix: string): ControlPanelS
           name: `adhoc_filters${controlSuffix}`,
           config: {
             ...sharedControls.adhoc_filters,
-            mapStateToProps: (state, controlState) => {
-              return {
-                ...(sharedControls?.adhoc_filters?.mapStateToProps?.(state, controlState) || {}),
-                value: [
-                  ...((state?.controls?.adhoc_filters?.value as any[]) || []).filter(
-                    filter => filter.isExtra,
-                  ),
-                  ...((state?.controls?.adhoc_filters_2?.value as any[]) || []),
-                ],
-              };
-            },
+            mapStateToProps: (state, controlState) => ({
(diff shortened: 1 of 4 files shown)


plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
```diff
@@ -70,7 +70,18 @@ function createQuerySection(label: string, controlSuffix: string): ControlPanelS
       [
         {
           name: `adhoc_filters${controlSuffix}`,
-          config: sharedControls.adhoc_filters,
+          config: {
+            ...sharedControls.adhoc_filters,
+            mapStateToProps: (state, controlState) => ({
+              ...(sharedControls?.adhoc_filters?.mapStateToProps?.(state, controlState) || {}),
```
kgabryje, line 76: I think that null check of `sharedControls` is not necessary here, as we don't check the null-ness 2 lines above 🙂

kgabryje (approved):
1 small nit, otherwise looks good!

plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
villebro: Got it 👍  I'm always unsure about these, hence I always end up padding everything with `?`

plugins/plugin-chart-echarts/src/MixedTimeseries/controlPanel.tsx
```diff
@@ -70,7 +70,18 @@ function createQuerySection(label: string, controlSuffix: string): ControlPanelS
...
-          config: sharedControls.adhoc_filters,
+          config: {
+            ...sharedControls.adhoc_filters,
+            mapStateToProps: (state, controlState) => ({
+              ...(sharedControls.adhoc_filters?.mapStateToProps?.(state, controlState) || {}),
+              value: [
+                ...((state?.controls?.adhoc_filters?.value as any[]) || []).filter(
+                  filter => filter.isExtra,
+                ),
+                ...((state?.controls?.adhoc_filters_2?.value as any[]) || []),
```
kgabryje, line 81:
```suggestion
                ...((state?.controls?.adhoc_filters_b?.value as any[]) || []),
```

kgabryje (changes requested):
1 bug noticed - `adhoc_filters_2` instead of `adhoc_filters_b`

stephenLYZ:
LGTM! Just fix the confilcting file.

villebro:

PR #1304, opened 2021-08-12
feat(plugin-chart-echarts): rich tooltip in ts chart supports scroll and highlighted

stephenLYZ:
🐛 Bug Fix
https://github.com/apache/superset/issues/16177

- Address all time series chart and mixd-timeseries chart
- Hovering on specific series will get highlighted.

### after

https://user-images.githubusercontent.com/11830681/129236128-b7c8966d-0645-45f1-8c65-fa24da9c502f.mov

#### bar chart
image

#### area chart
image

#### scatter chart
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
@@ -20,6 +20,7 @@ import React, { useCallback } from 'react';
 import { EchartsMixedTimeseriesChartTransformedProps } from './types';
 import Echart from '../components/Echart';
 import { EventHandlers } from '../types';
+import { currentSeries } from '../utils/series';
 
 export default function EchartsMixedTimeseries({
   height,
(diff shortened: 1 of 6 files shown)


villebro (approved):
This is super cool! 🤩 LGTM

villebro:
Removed 'hold' label - I was thinking about replacing `currentSeries` with a `Ref` but this requires changes that feel outside the scope of this PR. @stephenLYZ the unit tests need updating, after that this is good to go.

stephenLYZ:
@villebro yeah, I thought so too. I'm not sure if we can pass it to transformProps via setDataMask hook. Any good ideas on how to maintain this state?

villebro:
> @villebro yeah, I thought so too. I'm not sure if we can pass it to transformProps via setDataMask hook. Any good ideas on how to maintain this state?

Yeah, that would be one option, however, the problem is then we'd need to pass it via `ownState` which triggers a new query. Let's think about this some more. But in the meantime I think this is fine. If you have the time just add a `# TODO:` to that piece of code so we can refactor it later.

stephenLYZ:
> > @villebro yeah, I thought so too. I'm not sure if we can pass it to transformProps via setDataMask hook. Any good ideas on how to maintain this state?
> 
> 
> 
> Yeah, that would be one option, however, the problem is then we'd need to pass it via `ownState` which triggers a new query. Let's think about this some more. But in the meantime I think this is fine. If you have the time just add a `# TODO:` to that piece of code so we can refactor it later.

Okay~ I think the currentSerie maybe useful in some later scenarios, it points to the series of the current mouse hover.

PR #1307, opened 2021-08-13
fix: bignumber format by time formatter

zhaoyongjie:
🏆 Enhancements
closes: https://github.com/apache/superset/issues/16221

Support timestamp format in BigNumber viz

image

Test instruction:
1. Select birth_names dataset
2. Switch to Big Number viz
3. Select Max(ds) as Metric
4. Select TimeStamp format checkbox

### DIFF
--- a/plugins/legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
+++ b/plugins/legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
@@ -78,6 +78,7 @@ export default function transformProps(chartProps: BigNumberChartProps) {
   } = formData;
   const granularity = extractTimegrain(rawFormData as QueryFormData);
   let { yAxisFormat } = formData;
+  const { headerFormatSelector, headerTimestampFormat } = formData;
   const { data = [], from_dttm: fromDatetime, to_dttm: toDatetime } = queriesData[0];
   const metricName = typeof metric === 'string' ? metric : metric.label;
   const compareLag = Number(compareLag_) || 0;
@@ -146,7 +147,9 @@ export default function transformProps(chartProps: BigNumberChartProps) {
     });
   }
 
-  const formatNumber = getNumberFormatter(yAxisFormat);
+  const headerFormatter = headerFormatSelector
+    ? getTimeFormatter(headerTimestampFormat)
+    : getNumberFormatter(yAxisFormat);
   const formatTime =
     timeFormat === smartDateFormatter.id
       ? getTimeFormatterForGranularity(granularity)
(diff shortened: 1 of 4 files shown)


plugins/legacy-preset-chart-big-number/src/BigNumberTotal/controlPanel.ts
```diff
@@ -43,6 +48,36 @@ export default {
...
+            },
+          },
+        ],
+        [
+          {
+            name: 'header_timestamp_format',
+            config: {
+              type: 'SelectControl',
+              freeForm: true,
+              label: t('Timestamp format'),
```
kgabryje, line 69: Can we name it `Date format` for consistency with other charts?
Also, instead of adding a checkbox  to control visibility, why don't we just make this select button clearable?

zhaoyongjie: We need an explicit way to tell the user whether to use `number formatter` or `time formatter`

plugins/legacy-preset-chart-big-number/test/transformProps.test.ts
```diff
@@ -135,7 +135,7 @@ describe('BigNumber', () => {
         },
       };
       const transformed = transformProps(propsWithDatasource);
-      expect(transformed.formatNumber(transformed.bigNumber)).toStrictEqual('1.23');
+      expect(transformed.headerFormatter(transformed.bigNumber)).toStrictEqual('1.23');
```
kgabryje, line 138: Can you add a test for timestamp?

plugins/legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
```diff
@@ -146,7 +147,9 @@ export default function transformProps(chartProps: BigNumberChartProps) {
     });
   }
 
-  const formatNumber = getNumberFormatter(yAxisFormat);
+  const headerFormatter = headerFormatSelector
```
kgabryje, line 150: Can we rename `headerFormatSelector` to `isFormatTimestamp` or something like that? It's not obvious that it's a boolean value which is true if formatter is for timestamp

kgabryje (approved):
2 unblocking comments, other than that LGTM

plugins/legacy-preset-chart-big-number/test/transformProps.test.ts
zhaoyongjie: I will push a separate PR after this. Thanks!

plugins/legacy-preset-chart-big-number/src/BigNumber/transformProps.ts
zhaoyongjie: I will push a separate PR after this. Thanks!

PR #1308, opened 2021-08-13
fix: add advanced analytics to all of timeseries viz

zhaoyongjie:
🏆 Enhancements
closes: https://github.com/apache/superset/issues/16238

add & refactor advanced analytics pane to all of timeseries viz

image


packages/superset-ui-chart-controls/src/sections/forecastInterval.tsx
```diff
@@ -0,0 +1,123 @@
...
+        },
+      },
+    ],
+    [
+      {
+        name: 'forecastPeriods',
+        config: {
+          type: 'TextControl',
+          label: t('Forecast periods'),
+          validators: [legacyValidateInteger],
```
kgabryje, line 51: Why do we use `legacyValidateInteger` instead of `validateInteger`?

packages/superset-ui-chart-controls/src/sections/forecastInterval.tsx
```diff
@@ -0,0 +1,123 @@
...
+        },
+      },
+    ],
+    [
+      {
+        name: 'forecastInterval',
+        config: {
+          type: 'TextControl',
+          label: t('Confidence interval'),
+          validators: [legacyValidateNumber],
```
kgabryje, line 63: Same question as above

packages/superset-ui-chart-controls/src/sections/forecastInterval.tsx
zhaoyongjie: It is just a code move, no logical changes have been made. I can do refactor later.

packages/superset-ui-chart-controls/src/sections/forecastInterval.tsx
```diff
@@ -0,0 +1,123 @@
...
+ */
+import { legacyValidateInteger, legacyValidateNumber, t } from '@superset-ui/core';
+import { ControlPanelSectionConfig } from '../types';
+
+export const forecastEnabled = false;
+export const forecastInterval = 0.8;
+export const forecastPeriods = 10;
+export const forecastSeasonalityDaily = null;
+export const forecastSeasonalityWeekly = null;
+export const forecastSeasonalityYearly = null;
```
kgabryje, line 27: Can we export those params as a single object? Like `const FORECAST_DEFAULT_DATA`?

packages/superset-ui-chart-controls/src/sections/forecastInterval.tsx
kgabryje: 👍

villebro: That's a good question. I remember someone long ago said to just juse legacy for now, but I can't remember why. Maybe now is a good time to migrate over and fix whatever potential bugs there are in the new one.

rusackas: Now is not a great moment to introduce risk... I'd rather add that in a follow-up.

rusackas (approved):
LGTM! Thanks for all the cleanup!

PR #1309, opened 2021-08-13
fix(plugin-chart-echarts): default to use node name for color category

villebro:
🐛 Bug Fix

When a source or target color hasn't been chosen, the color scheme isn't used. This changes the behavior so that the node name will be used to determine the color when a category hasn't been chosen. This matches how colors are usually picked by other charts. When a the source or target color column has been chosen the behavior is unchanged. In addition, the control name and descriptions are updated to be more explanatory.

### AFTER
image

### BEFORE
image


ktmud:
what if users do want colorless graphs?

villebro:
That's a good point. At any rate we should make the single color respect the color scheme by picking e.g. the first color from the palette.

villebro:

PR #1310, opened 2021-08-13
fix: Make the scrollbar appear inside the table  

geido:
### SUMMARY

This PR only solves an issue with the scrollbar not appearing in the table container but at the far right. Read [this comment](https://github.com/apache/superset/issues/15935#issuecomment-897699534) for context 

### BEFORE

https://user-images.githubusercontent.com/60598000/129375411-4e61d037-2a27-4762-a6f7-13848813e763.mp4

### AFTER

https://user-images.githubusercontent.com/60598000/129585796-2fe468b7-1dfa-4367-9459-34a3f3fb5c5b.mp4

### TESTING INSTRUCTIONS

1. Open Explore with a Pivot Table V2 viz
2. Observe the table and scroll
3. Make sure the scrollbar appears within the table container

### ADDITIONAL INFORMATION

- [x] Has associated issue: https://github.com/apache/superset/issues/15935
- [ ] Changes UI
- [ ] Includes DB Migration (follow approval process in [SIP-59](https://github.com/apache/superset/issues/13351))
  - [ ] Migration is atomic, supports rollback & is backwards-compatible
  - [ ] Confirm DB migration upgrade and downgrade tested
  - [ ] Runtime estimates and downtime expectations provided
- [ ] Introduces new feature or API
- [ ] Removes existing feature or API

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -49,8 +49,9 @@ const Styles = styled.div<PivotTableStylesProps>`
 
 const PivotTableWrapper = styled.div`
   height: 100%;
+  max-width: 100%;
+  overflow: auto;
   width: min-content;
-  overflow-y: auto;
 `;
 
 const METRIC_KEY = 'metric';
--- a/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
+++ b/plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
@@ -49,9 +49,8 @@ const Styles = styled.div<PivotTableStylesProps>`
 
 const PivotTableWrapper = styled.div`
   height: 100%;
-  max-width: 100%;
+  max-width: fit-content;
   overflow: auto;
-  width: min-content;
 `;
 
 const METRIC_KEY = 'metric';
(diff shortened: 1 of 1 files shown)


rusackas:
Code LGTM... just going to do a little `npm link` action to test it with tables of various widths/heights in different browsers.

rusackas:
When the table goes super wide, it scrolls far too much content. Only the table itself should scroll, not the header/toolbar above or the data preview below. I think you might just need to add a max-width to the new wrapper you've created.


https://user-images.githubusercontent.com/812905/129410971-c6325586-5045-476b-8fb2-ef0eb4f3151d.mp4

geido:
Hey @rusackas I have included the fix for the X axis as well in this PR as you suggested. I have also updated the video in the PR description. Thanks!

plugins/plugin-chart-pivot-table/src/PivotTableChart.tsx
```diff
@@ -44,10 +44,16 @@ const Styles = styled.div<PivotTableStylesProps>`
...
       width: ${width - margin * 2}px;
-      overflow-y: scroll;
  `}
 `;
 
+const PivotTableWrapper = styled.div`
+  height: 100%;
+  max-width: 100%;
+  overflow: auto;
+  width: min-content;
```
kgabryje, line 54: Can we replace `max-width: 100%` and `width: min-content` with `max-width: fit-content`? 
Using `width: min-content` causes some unnecessary wrapping in the table, whereas `max-width: fit-content` doesn't affect the table, but sets the containers width to the child's width.
According to https://caniuse.com/?search=fit-content it's widely supported

https://user-images.githubusercontent.com/15073128/129593475-c849e84c-63a8-4e75-a953-377dbd55839e.mov

geido: Hey @kgabryje that was a very nice catch! Thanks for that. I changed it.

michael-s-molina (approved):

PR #1314, opened 2021-08-16
feat(core): hide chart when all metrics values are null

kgabryje:
For some charts, when all metrics have only null values it doesn't make much sense to display the chart. Instead, we want to show a message that says "All metrics values are null for this query". 
This PR introduces that behaviour in Pivot Table v2 and Table charts.

https://user-images.githubusercontent.com/15073128/129589847-a221bd46-6f44-40a2-994a-a3870671d7f1.mov

@junlincc @jinghua-qa @adam-stasiak

### DIFF
--- a/packages/superset-ui-core/src/chart/components/SuperChart.tsx
+++ b/packages/superset-ui-core/src/chart/components/SuperChart.tsx
@@ -135,16 +135,21 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
       enableNoResults &&
       (!queriesData ||
         queriesData.every(({ data }) => !data || (Array.isArray(data) && data.length === 0)));
-    const metricLabels = ensureIsArray(chartProps.formData?.metrics || chartProps.formData?.metric)
-      .map((metric: AdhocMetric | string) => (typeof metric === 'string' ? metric : metric.label))
-      .filter(Boolean);
-    const hasNullMetrics =
-      metricLabels.length > 0 &&
-      metricLabels.every(label =>
-        queriesData?.every(({ data }) =>
-          ensureIsArray(data).every(record => record[label!] === null),
-        ),
-      );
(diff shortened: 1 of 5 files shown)


packages/superset-ui-core/src/chart/components/SuperChart.tsx
```diff
@@ -132,8 +135,33 @@ export default class SuperChart extends React.PureComponent<Props, {}> {
       enableNoResults &&
       (!queriesData ||
         queriesData.every(({ data }) => !data || (Array.isArray(data) && data.length === 0)));
+    let hasNullMetrics = false;
+    if (chartProps.formData?.disallowNullMetrics) {
```
villebro, line 139: This name is a bit ambiguous in the sense that it's unclear what disallowing exactly means. This is a bit of a monster, but perhaps something along these lines could work? `showNoResultsOnAllNullMetrics` (ok that's really horrible, but you get the idea)

villebro (approved):
Minor naming comment (since I can't come up with a better proposal, so approving). One thing that I think is good to consider is that some users may not consider all null metrics to be a "No result". For instance, in the case of the pivot table, if we have dimension data for both rows and columns, a user might mistake the "All metrics values are null for this query" result for the regular "No results were returned for this query" (some people, myself included, don't always properly read error messages). In this case, showing the pivot table with populated row and column labels but with empty cells would communicate to the user "hey, there's obviously some data here, but it's just the metric data that's null", which would then prompt the user to perhaps check if the aggregate expression is incorrect.

Anyway, I think this is an edge case, and I expect users to prefer all null metrics values to display the "No results" view over the "All empty cells" result. But I just wanted to bring this case up, and personally I probably prefer seeing an empty pivot table with labels over the "No results" page.

villebro:

PR #1315, opened 2021-08-17
fix(plugin-chart-table): apply correct date formatting for time grain

kgabryje:
Before, proper date formatting based on selected timegrain was applied only when column's name was `__timestamp`. This PR changes that behaviour in Table and Pivot Table v2 charts so that if granularity and adaptive formatting are selected, the date formatter proper for the time grain is used.

Before: see https://github.com/apache/superset/issues/16283
After: 
image

Fixes https://github.com/apache/superset/issues/16283

A PR with the change on backend is in progress (made by @villebro)

CC: @junlincc

### DIFF
--- a/plugins/plugin-chart-pivot-table/src/plugin/transformProps.ts
+++ b/plugins/plugin-chart-pivot-table/src/plugin/transformProps.ts
@@ -31,7 +31,6 @@ import { getColorFormatters } from '@superset-ui/chart-controls';
 import { DateFormatter } from '../types';
 
 const { DATABASE_DATETIME } = TimeFormats;
-const TIME_COLUMN = '__timestamp';
 
 function isNumeric(key: string, data: DataRecord[] = []) {
   return data.every(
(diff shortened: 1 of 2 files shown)


plugins/plugin-chart-pivot-table/src/plugin/transformProps.ts
```diff
@@ -108,7 +107,7 @@ export default function transformProps(chartProps: ChartProps<QueryFormData>) {
     .reduce((acc: Record<string, DateFormatter | undefined>, temporalColname: string) => {
       let formatter: DateFormatter | undefined;
       if (dateFormat === smartDateFormatter.id) {
-        if (temporalColname === TIME_COLUMN) {
+        if (granularity) {
```
villebro, line 110: Should we also ensure that the column type is temporal? Otherwise we will end up applying the temporal formatter on all groupbys if a time grain and smart date formatter is selected.

kgabryje: We do that in line 106 :) 
`.filter((colname: string, index: number) => coltypes[index] === GenericDataType.TEMPORAL)`

villebro: Lol nvm!

villebro (approved):

PR #1319, opened 2021-08-19
feat: added italian regions country map

romanfed:


plugins/legacy-plugin-chart-country-map/src/countries.ts
```diff
@@ -79,6 +81,9 @@ export const countryOptions = Object.keys(countries).map(x => {
   if (x === 'uk' || x === 'usa') {
     return [x, x.toUpperCase()];
   }
+  if (x === 'italy_regions') {
+    return [x, 'Italy (regions)'];
```
villebro, line 85: I believe this should be translated:
```suggestion
    return [x, t('Italy (regions)')];
```

villebro (changes requested):

plugins/legacy-plugin-chart-country-map/src/countries.ts
romanfed: If I'm not wrong, there aren't translations for country map names. I think adding translations should be another task.

junlincc:
how can we get this PR through? @villebro

plugins/legacy-plugin-chart-country-map/src/countries.ts
villebro: Fair enough - let's do a translation sweep later when that becomes a priority.

villebro (approved):

ktmud:
What's the difference between "Italy" and "Italy (regions)"? Can we add more PR description just for context?

PR #1320, opened 2021-08-19
fix: Improve big number time format UX

etr2460:
🐛 Bug Fix
https://github.com/apache-superset/superset-ui/pull/1278 introduced the ability to show a timestamp for the value of big number, but as a result it overrode the default time format for the tooltip in big number trendline charts to a somewhat obscure (for americans) `%d-%m-%Y %H:%M:%S` format. This PR:
- makes the time formatter always visible (so that the big number trendline chart can always set it)
- Defaults to a format that's used globally, `'%Y-%m-%d %H:%M:%S'`

to: @krsnik93 @villebro @ktmud

### DIFF
--- a/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
+++ b/plugins/legacy-preset-chart-big-number/src/BigNumber/controlPanel.tsx
@@ -71,7 +71,6 @@ const config: ControlPanelConfig = {
               label: t('Timestamp format'),
               renderTrigger: true,
               choices: D3_TIME_FORMAT_OPTIONS,
-              default: '%Y-%m-%d %H:%M:%S',
               description: D3_FORMAT_DOCS,
             },
           },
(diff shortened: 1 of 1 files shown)


krsnik93:
Hello @etr2460 . Thank you for the improvements and sorry for defaulting to a non standard time format.

I am completely fine with changing the default time format. Regarding removing the `show_timestamp` boolean control and always showing the timestamp, I think:

- additional changes might be needed in https://github.com/apache-superset/superset-ui/blob/68abb0d7c13d4d6dd50a8c976c1b07d0cea4d0d1/plugins/legacy-preset-chart-big-number/src/BigNumber/BigNumber.tsx#L106 because otherwise the timestamp will never be shown
- if the latest returned value from the db is null, the next non-null value is displayed along with a warning (that the value is not fresh) in the position where the timestamp would be now. If the timestamp were to be always shown, then this fallback warning could be removed. I think @villebro would be able to say more in terms of whether this is a good idea.

krsnik93:
Hello @etr2460 thanks for the improvements and sorry for defaulting to a non-standard time format.

Changes are completely fine with me.

ktmud (approved):
YYYY-MM-DD for the win!

ktmud:
But shouldn't the default be Smart Date? I don't think we even need to provide a default if the control is going to be displayed always.

etr2460:
Hmm, if that's true then i agree removing it is correct. I'll make the change

villebro (commented):
LGTM

PR #1325, opened 2021-08-21
fix(plugin-chart-echarts): [scatter-plot] remove contribution mode

stephenLYZ:
🏆 Enhancements

refer to https://github.com/apache/superset/issues/16342#issuecomment-902418851


villebro (approved):
LGTM. If we end up having to do small changes to the default control panel on all charts, it might be a good idea to break out the control panel into a factory function that then returns a control panel based on requested removals/overrides. But for now this is a good solution until this development settles down.

PR #1327, opened 2021-08-23
feat(legacy-plugin-chart-country-map): add six new maps

villebro:
🏆 Enhancements

Add the following maps:
- Denmark
- Estonia
- Iceland
- Norway (move+scale Svalbard closer to mainland)
- Poland
- Sweden

Also update notebook to add logic for Finland and Norway map changes.


ktmud (approved):
Would be nice to include the notebook changes, too.

villebro:
I'll do that - I had to do some modifications to the Finland map earlier, and might also pick in Svalbard closer to mainland Norway to make that available (though I suspect it's of less use due to being so remote+sparsely populated). I'll open a PR later today.

PR #1337, opened 2021-08-29
fix(plugin-chart-echarts): [feature-parity] display custom metric label

stephenLYZ:
🐛 Bug Fix
This PR fixes the problem that custom metric label in the legend cannot show correctly.

refer to https://github.com/apache/superset/issues/16332

### before
image

### after

image

### DIFF
--- a/plugins/plugin-chart-echarts/src/utils/prophet.ts
+++ b/plugins/plugin-chart-echarts/src/utils/prophet.ts
@@ -90,7 +90,10 @@ export const formatProphetTooltipSeries = ({
   return `${row.trim()}`;
 };
 
-export function rebaseTimeseriesDatum(data: TimeseriesDataRecord[]) {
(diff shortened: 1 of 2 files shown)


zhaoyongjie (approved):
LGTM, leave some no-blocking comments.

plugins/plugin-chart-echarts/src/utils/prophet.ts
```diff
@@ -90,7 +90,10 @@ export const formatProphetTooltipSeries = ({
   return `${row.trim()}`;
 };
 
-export function rebaseTimeseriesDatum(data: TimeseriesDataRecord[]) {
+export function rebaseTimeseriesDatum(
+  data: TimeseriesDataRecord[],
+  verboseMap: Record<string, string> = {},
+) {
```
zhaoyongjie, line 96: Could you add a unit test for this function change?

PR #1338, opened 2021-08-30
refactor(core): normalize orderby in QueryObject

zhaoyongjie:
🏆 Enhancements
Currently, Superset hasn't got a consistent `order by` clause in the v1 query API.
- Some Charts can generate `order by` with `ASC`, some charts can't.
- The Superset query always appends the `limit` clause, but often forgets to append the `order by` clause. (This can lead to inconsistent query results once the data has changed.)
- There are 3 ways of orderby definitions in QueryObject. (a: order_desc and timeseries_limit_metric, b: first of metrics and form_data.sort_by_metric, c: orderby)

image

This PR implements a uniform approach to `orderby`, avoiding to processing of each viz.


packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -19,46 +19,50 @@
...
+    const orderbyClauses = formData.orderby.map(item => {
+      // value can be in the format of `['["col1", true]', '["col2", false]']`,
+      // where the option strings come directly from `order_by_choices`.
+      if (typeof item === 'string') {
+        try {
+          return JSON.parse(item);
+        } catch (error) {
+          throw new Error(t('Found invalid orderby options'));
+        }
+      }
```
villebro, line 40: Why are we `JSON.parse`ing a `string` object here?

villebro (commented):
A few quick comments - will review and test more properly soon

packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -19,46 +19,50 @@
...
 
-export default function normalizeOrderBy(queryObject: QueryObject): QueryObject {
-  if (Array.isArray(queryObject.orderby) && queryObject.orderby.length > 0) {
+const defaultOrderby = undefined;
+
+export default function normalizeOrderBy(
+  formData: FormDataResidual,
+): QueryFormOrderBy[] | typeof defaultOrderby {
+  if (Array.isArray(formData.orderby) && formData.orderby.length > 0) {
+    const orderbyClauses = formData.orderby.map(item => {
```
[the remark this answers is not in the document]
kgabryje: We can even skip `if (length > 0)` - map won't do anything on array with length 0

packages/superset-ui-core/src/query/normalizeOrderBy.ts
zhaoyongjie: The item my be a JSON payload for orderby
```
['["col1", true]', '["col2", false]']
```

villebro: Oh now I noticed you were just moving this code unchanged... I wonder if this is just for backwards compatibility with some old version of the control, or if it's still actively passing JSON formatted strings. If it's still doing that, we should probably update the type here: https://github.com/apache-superset/superset-ui/blob/8249e92f0d94fd6f7792121ed2354631d9532aee/packages/superset-ui-chart-controls/src/types.ts#L54 Or better yet, make sure it actually passes `[string, string][] | null`, not `str[] | null` 😆

packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -19,46 +19,55 @@
...
       Array.isArray(orderbyClause) &&
       orderbyClause.length === 2 &&
       !isEmpty(orderbyClause[0]) &&
-      isBoolean(orderbyClause[1])
-    ) {
-      return queryObject;
-    }
+      isBoolean(orderbyClause[1]),
+  );
+  if (validatedOrderbys.length > 0) {
```
villebro, line 51: nit:
```suggestion
  if (validatedOrderbys.length) {
```

packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -19,46 +19,55 @@
...
 
-  // ensure that remove invalid orderby clause
-  const cloneQueryObject = { ...queryObject };
-  delete cloneQueryObject.timeseries_limit_metric;
-  delete cloneQueryObject.order_desc;
-  delete cloneQueryObject.orderby;
-
-  const isAsc = !queryObject.order_desc;
+  // the `timeseries_limit_metric` and `order_desc` are used for orderby
+  const isAsc = !formData.order_desc;
```
villebro, line 56: I personally think we should avoid using `timeseries_limit_metric` to order the main query. I know many charts, especially legacy ones, seem to use it both for series limiting (subquery) and ordering the main query, but IMO timeseries charts `is_timeseries = true` should rather default to sorting primarily by the temporal column, and only then default to the `timeseries_limit_metric` if necessary.

A good example of this is the table chart, where setting the 'Include time' control to `true` orders the data in descending order according to the metric, NOT the temporal column. This has a very weird effect, where the data is displayed in non-chronological order.

packages/superset-ui-core/test/query/buildQueryObject.test.ts
```diff
@@ -149,7 +151,8 @@ describe('buildQueryObject', () => {
       viz_type: 'table',
       timeseries_limit_metric: metric,
     });
-    expect(query.timeseries_limit_metric).toEqual(metric);
+    expect(query.orderby).toEqual([[metric, true]]);
+    expect(query.timeseries_limit_metric).toEqual(undefined);
```
villebro, line 155: IMO this test should sort primarily by `granularity_sqla`, not `metric`. So something like 
```javascript
expect(query.orderby).toEqual([['ds, true], [metric, true]]);
```

villebro (approved):
This has a great "cleansing effect" on the `buildQuery` callbacks! :) LGTM as this is in-line with how this currently works, but I'd really like to see the discussion being opened up about sorting primarily time col for `is_timeseries === true`.

packages/superset-ui-core/src/query/normalizeOrderBy.ts
zhaoyongjie: Let's discuss offline how to deal with the `order by` logic of inner and outer queries. ^_^

zhaoyongjie: okay. I will remove the transformation.

villebro (approved):
LGTM 👍

villebro:

PR #1339, opened 2021-08-31
chore: isolate the one actual jquery dependency

suddjian:
💔 Breaking Changes

🏆 Enhancements

📜 Documentation

🐛 Bug Fix

🏠 Internal

It uses the Data Tables jquery plugin and is deprecated, so don't want to try to remove jquery from it. Instead will make it an explicit dependency so that jquery can be removed from the rest of superset.

### DIFF
--- a/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
+++ b/plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
@@ -19,6 +19,7 @@
 /* eslint-disable react/sort-prop-types */
 import dt from 'datatables.net-bs';
 import PropTypes from 'prop-types';
+import $ from 'jquery';
 import {
   getTimeFormatter,
   getTimeFormatterForGranularity,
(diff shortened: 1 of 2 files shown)


plugins/legacy-plugin-chart-pivot-table/src/PivotTable.js
```diff
@@ -28,10 +29,7 @@ import { formatCellValue, formatDateCellValue } from './utils/formatCells';
 import fixTableHeight from './utils/fixTableHeight';
 import 'datatables.net-bs/css/dataTables.bootstrap.css';
 
-if (window.$) {
-  dt(window, window.$);
-}
-const $ = window.$ || dt.$;
+dt(window, $);
```
nytai, line 32: Does this break since `$` will now be undefined.

suddjian: see line 22

villebro (approved):

PR #1341, opened 2021-09-01
Update to finland map

Anushka-Hapukotuwa:
🏆 Enhancements
🐛 Bug Fix

Fix to Finland map not loading correctly.
finland-country-map-not-loading


villebro:
This was already fixed by #1321 (this PR also added Åland Islands which was also missing)

PR #1344, opened 2021-09-01
fix(legacy-plugin-chart-country-map): Correct ISO 3166-2 county codes for Estonia

intgr:
🐛 Bug Fix

I have no idea how this happened, but most Estonian county codes were
off by one.

Unfortunately I couldn't compare it to latest data from
naturalearthdata.com, since download links are currently down.

Source: https://en.wikipedia.org/wiki/ISO_3166-2:EE#Counties

Primary source: https://www.evs.ee/en/iso-3166-2-2020 (but I'm not going
to pay 100 €)


villebro:
Thanks for the fix @intgr! I generated the original `geojson` file with this notebook https://github.com/apache-superset/superset-ui/blob/master/plugins/legacy-plugin-chart-country-map/scripts/Country%20Map%20GeoJSON%20Generator.ipynb . I quickly checked and I agree; based on Wikipedia Pärnu is "EE-68", not "EE-67",  So it appears Natural Earth Data have incorrect data. I wonder if they have a channel for submitting corrections? I also noticed they had a weird incorrect polygon in Norway, so perhaps I could get that fixed, too.

villebro (approved):

PR #1347, opened 2021-09-06
chore(core): move @emotion/styled to peer dependency

villebro:
🏠 Internal

Move `@emotion/styled` to peer dependency to fix version conflict on `apache/superset`. Also bump other `@emotion` deps to match those in https://github.com/apache/superset/pull/16559


zhaoyongjie (approved):
LGTM

PR #1350, opened 2021-09-06
feat(plugin-chart-echarts): add x and y label support for timeseries

MaxHuiYYDS:
fix #16512

🏆 Enhancements
This PR supports two features:
1 User can add X/Y axis label in timeseries line chart. 
2 Changed previews default top Y axis position to dynamic left/top choice.
related to:
https://github.com/apache/superset/issues/16512

After:

https://user-images.githubusercontent.com/51693396/132231973-6fcae913-95c9-496a-a102-e63290c9b815.mov


MaxHuiYYDS:
@junlincc Hi Junlin, please check the video above, seems I have done further more, not only the axis, but  also the left/top y axis margin and x axis bottom margin. 
I just added timeseries line chart currently, if you think that is fine, I prefer add this enhancement in another pr for other charts.

PR #1352, opened 2021-09-07
chore: clean up mixed-timeseries chart

zhaoyongjie:
🏆 Enhancements
clean up codebase, no functional changes

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
@@ -16,13 +16,8 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-import {
-  buildQueryContext,
-  getMetricLabel,
-  QueryFormData,
-  QueryObject,
-  normalizeOrderBy,
-} from '@superset-ui/core';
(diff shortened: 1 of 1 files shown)


plugins/plugin-chart-echarts/src/MixedTimeseries/buildQuery.ts
```diff
@@ -61,45 +56,19 @@ export default function buildQuery(formData: QueryFormData) {
...
       is_timeseries: true,
-      post_processing: [
-        {
-          operation: 'pivot',
-          options: {
-            index: ['__timestamp'],
-            columns: formData1.columns || [],
-            // Create 'dummy' sum aggregates to assign cell values in pivot table
-            aggregates: Object.fromEntries(
-              metricLabels.map(metric => [metric, { operator: 'sum' }]),
```
suddjian, line 76: `pivotOperator()`  uses `mean` rather than `sum`. Does these aggregates being "dummy" mean that it doesn't matter whether which operator we use?

zhaoyongjie: Yes, It does. The main reason for using `mean` is to avoid `drop NaN`. refer to: https://github.com/apache-superset/superset-ui/pull/1231

villebro (approved):

PR #1356, opened 2021-09-10
feat(core): add support for non-temporal series limit

villebro:
🏆 Enhancements

Currently it is impossible to restrict the number of series in non-temporal charts. This adds three new `QueryForm` and `QueryObject` properties, along with relevant shared controls for them:
- `series_columns`: the columns to group by when limiting series (previously always assumed to be all groupby columns except the temporal column). This can be left `undefined`/`[]`, in which case series limiting won't happen.
- `series_limit`: to replace `timeseries_limit`
- `series_limit_metric`: to replace `timeseries_limit_metric`

As a POC, this functionality is implemented on the BoxPlot chart where this functionality was mostly broken (only working if distribution was across temporal column).

Related PR on `apache/superset`: https://github.com/apache/superset/pull/16660

### AFTER
Here we've defined a non-temporal series limit column `countries` with a limit of 5 series, and the chart renders correctly with the top 5 regions:
image

### BEFORE
Before the x-axis would populate more than the specified 5 series:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/BoxPlot/buildQuery.ts
+++ b/plugins/plugin-chart-echarts/src/BoxPlot/buildQuery.ts
@@ -22,14 +22,14 @@ import { BoxPlotQueryFormData, BoxPlotQueryObjectWhiskerType } from './types';
 const PERCENTILE_REGEX = /(\d+)\/(\d+) percentiles/;
 
 export default function buildQuery(formData: BoxPlotQueryFormData) {
-  const { whiskerOptions, columns: distributionColumns = [] } = formData;
+  const { columns = [], granularity_sqla, groupby = [], whiskerOptions } = formData;
   return buildQueryContext(formData, baseQueryObject => {
     let whiskerType: BoxPlotQueryObjectWhiskerType;
     let percentiles: [number, number] | undefined;
-    const { columns = [], metrics = [] } = baseQueryObject;
+    const { metrics = [] } = baseQueryObject;
     const percentileMatch = PERCENTILE_REGEX.exec(whiskerOptions as string);
 
-    if (whiskerOptions === 'Tukey') {
+    if (whiskerOptions === 'Tukey' || whiskerOptions === undefined) {
       whiskerType = 'tukey';
     } else if (whiskerOptions === 'Min/max (no outliers)') {
       whiskerType = 'min/max';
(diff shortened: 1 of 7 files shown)


plugins/plugin-chart-echarts/src/BoxPlot/controlPanel.ts
```diff
@@ -37,8 +37,9 @@ export default {
         ['adhoc_filters'],
         emitFilterControl,
         ['groupby'],
-        ['columns'],
-        ['limit'],
+        ['columns'], // this should be migrated to `series_columns`
```
michael-s-molina, line 40:
```suggestion
        ['columns'], // TODO: this should be migrated to `series_columns`
```

plugins/plugin-chart-echarts/src/BoxPlot/buildQuery.ts
```diff
@@ -22,14 +22,14 @@ import { BoxPlotQueryFormData, BoxPlotQueryObjectWhiskerType } from './types';
...
+  const { columns = [], granularity_sqla, groupby = [], whiskerOptions } = formData;
   return buildQueryContext(formData, baseQueryObject => {
     let whiskerType: BoxPlotQueryObjectWhiskerType;
     let percentiles: [number, number] | undefined;
-    const { columns = [], metrics = [] } = baseQueryObject;
+    const { metrics = [] } = baseQueryObject;
     const percentileMatch = PERCENTILE_REGEX.exec(whiskerOptions as string);
 
-    if (whiskerOptions === 'Tukey') {
+    if (whiskerOptions === 'Tukey' || whiskerOptions === undefined) {
```
kgabryje, line 32: image

Should we check for null instead of undefined?

villebro: Oh, good catch! Although we should probably make the whisker type mandatory. Anyhoo, needs to be fixed 👍

kgabryje (approved):
LGTM!

plugins/plugin-chart-echarts/src/BoxPlot/buildQuery.ts
villebro: I made it unclearable and changed this to check for `!whiskerOptions` to ensure backwards compatibility in case someone has saved it with an empty whisker type.

zhaoyongjie (changes requested):
`sort by` control can't append `order by` clause in the main-query(outer-query).

Could we make some trade-off, how to do sort in both main-query(outer) and series-limit query(inner)

image

ktmud (approved):
LGTM

zhaoyongjie (approved):
LGTM

villebro:
Thanks all for the input and reviews! Looking forward to help make generic x-axis happen!

PR #1362, opened 2021-09-16
chore: replace mathjs with mexp

villebro:
🏠 Internal
Replace `mathjs` (11 Mb bundle) with `math-expression-evaluator` (71 kb) as done on https://github.com/apache/superset/pull/16701.

Checklist:
- added unit tests to `plugins-chart-echart` for formula evaluation
- ensure formula annotations work on both NVD3 and ECharts plugin:

NVD3 plugin after migration:
image

ECharts plugin after migration:
image

Also fix z-index of formula to make sure it is displayed above other charts:

### AFTER
image

### BEFORE
image


kgabryje (approved):
Lgtm, just remove the console log please 🙂

PR #1364, opened 2021-09-17
refactor: migrate timeseries_limit_metric to legacy_order_by

zhaoyongjie:
🏆 Enhancements
backend changes at: https://github.com/apache/superset/pull/16849

currently, Superset use `timeseries_limit_metric` control for main query orderby. While this is actually prepared for the inner query when viz is timeseries-like.

To fix the original incorrect use, so this PR introduce a `legacy_order_by` control.

This PR only migrate `pivot table v2`.

### DIFF
--- a/packages/superset-ui-core/src/query/normalizeOrderBy.ts
+++ b/packages/superset-ui-core/src/query/normalizeOrderBy.ts
@@ -38,6 +38,7 @@ export default function normalizeOrderBy(queryObject: QueryObject): QueryObject 
   // ensure that remove invalid orderby clause
   const cloneQueryObject = { ...queryObject };
   delete cloneQueryObject.timeseries_limit_metric;
+  delete cloneQueryObject.legacy_order_by;
   delete cloneQueryObject.order_desc;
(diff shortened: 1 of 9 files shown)


packages/superset-ui-core/src/query/normalizeOrderBy.ts
```diff
@@ -53,6 +54,19 @@ export default function normalizeOrderBy(queryObject: QueryObject): QueryObject
     };
   }
 
+  // todo: Removed `legacy_ordery_by` after refactoring
```
zhaoyongjie, line 57: ﻿This part of the logic will be refactored in other [PR](https://github.com/apache-superset/superset-ui/pull/1338) .

villebro (approved):
LGTM - only comment is should we perform a db migration to avoid breaking existing charts that are leveraging the old sorting control?

rusackas:
@zhaoyongjie is the migration that @villebro mentioned being tackled? I hope to avoid any reports of broken charts from users.

PR #1370, opened 2021-09-21
chore(plugin-chart-echarts): bump to ECharts 5.2.1

villebro:
🐛 Bug Fix

Bump to latest version of ECharts. Also set `axisLabel.hideOverlap` to `true` to get rid of overlapping x-axis labels. See https://github.com/apache/echarts/pull/15712 and https://github.com/apache/echarts/pull/15583 for details.

### AFTER
image

### BEFORE
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
@@ -181,6 +181,7 @@ export default function transformProps(
           fontWeight: 'bold',
         },
       },
+      // @ts-ignore
       data: transformedData,
     },
   ];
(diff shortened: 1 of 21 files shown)


plugins/plugin-chart-echarts/src/Funnel/transformProps.ts
```diff
@@ -181,11 +181,12 @@ export default function transformProps(
           fontWeight: 'bold',
         },
       },
+      // @ts-ignore
```
villebro, line 184: There seems to be a minor regression in the `FunnelSeriesOption` interface (will look into fixing this upstream later)

zhaoyongjie (approved):
Tested in local, LGTM(except for typing check)

PR #1374, opened 2021-09-23
chore: upgrade to Node 16

villebro:
🏠 Internal

Upgrade CI to Node 16 and make all Node versions at or above 14 to be supported.


zhaoyongjie (approved):
CI passed! so I think is successful. Thank for the upgrade!

PR #1375, opened 2021-09-23
chore: Remove immutable.js from map-box

etr2460:
We barely use `immutable` in superset so let's remove it.

I have no clue how to test this. I'm hoping it's in the storybook

to: @ktmud @graceguo-supercat @nytai

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12958,7 +12958,7 @@ immutability-helper@^3.1.1:
   resolved "https://registry.yarnpkg.com/immutability-helper/-/immutability-helper-3.1.1.tgz#2b86b2286ed3b1241c9e23b7b21e0444f52f77b7"
   integrity sha512-Q0QaXjPjwIju/28TsugCHNEASwoCcJSyJV3uO1sOIQGI0jKgm9f41Lvz0DZj3n46cNCyAZTsEYoY4C2bVRUzyQ==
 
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -12958,7 +12958,7 @@ immutability-helper@^3.1.1:
   resolved "https://registry.yarnpkg.com/immutability-helper/-/immutability-helper-3.1.1.tgz#2b86b2286ed3b1241c9e23b7b21e0444f52f77b7"
   integrity sha512-Q0QaXjPjwIju/28TsugCHNEASwoCcJSyJV3uO1sOIQGI0jKgm9f41Lvz0DZj3n46cNCyAZTsEYoY4C2bVRUzyQ==
 
-immutable@^3.8.1, immutable@^3.8.2:
+immutable@^3.8.1:
```
etr2460, line 12961: immutable is still here because `@data-ui/event-flow` has it as a dependency. However, it's not actually used in that package at all, so treeshaking should remove it from the bundle

etr2460:
this still looks right: https://superset-ui-git-erik-ritter-remove-immutable-superset.vercel.app/?path=/story/legacy-chart-plugins-legacy-plugin-chart-map-box--basic

ktmud (approved):
Thanks for the cleanup!

PR #1383, opened 2021-10-01
fix: replace absolute difference with difference in echart

zhaoyongjie:
🐛 Bug Fix

Currently, incorrect type of calculation in `time comparsion`, should replace absolute difference with difference.

backend codes at: https://github.com/apache/superset/pull/16930

related issue: https://github.com/apache/superset/issues/16766


villebro (approved):
LGTM, as long as we add a migration to migrate the affected chart metadata.

PR #1384, opened 2021-10-01
fix: replace absolute difference with difference in legacy charts

zhaoyongjie:
🐛 Bug Fix
replace `absolute difference` with `difference` in legacy charts

#### After:

image

related issue: https://github.com/apache/superset/issues/16766


villebro (approved):
LGTM with the caveat of adding a migration

PR #1386, opened 2021-10-02
fix: rolling and cum operator in multiple series chart

zhaoyongjie:
🐛 Bug Fix

Currently, Rolling Window calculation results are incorrect on multiple series. 

Backendcode at: https://github.com/apache/superset/pull/16945


villebro (approved):
LGTM

geido (approved):
LGTM!

villebro:
One minor improvement proposal if you have the time: it would be nice to set `isVisible` to `false` on the the `rolling_periods` control when `rolling_type` is `null`. But this is totally optional - maybe we can do a separate improvement PR to do a similar thing to all advanced analytics controls.

PR #1389, opened 2021-10-05
feat(plugin-chart-echarts): add support for custom forecasts

villebro:
🏆 Enhancements

The need to be able to create custom forecasts (without Prophet) has come up. In order to make this possible, we will assume that a forecast is present if there is a series with the forecast suffix (`__yhat`, `__yhat_lower` or `__yhat_upper`).

This is an intermediate step to decouple forecasts from Prophet - ultimately we should refactor the code in the following ways:
- move the relevant utilities/types from `plugin-chart-echarts` to `superset-ui/core`
- add support for other suffixes to make it possible to do similar inference from time offset series etc. Example: `my_metric` could be accompanied by `my_metric__offset(PT1D)` which could then be shown in the legend (with translations) as "my metric (1 day time offset)" or similar.
- Add documentation for how these can be used.

### SCREENSHOTS
Notice, that I haven't enabled forecasting, yet I'm able to create a chart with a forecast and confidence interval by applying the necessary suffixes to the series names:
image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
@@ -46,6 +46,7 @@ import {
 import { extractAnnotationLabels } from '../utils/annotation';
 import {
   extractForecastSeriesContext,
+  extractForecastSeriesContexts,
   extractProphetValuesFromTooltipParams,
   formatProphetTooltipSeries,
   rebaseTimeseriesDatum,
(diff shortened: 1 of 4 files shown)


BillGreerUX:
Thank you, this fits our use-case we requested to allow us to create custom forecasts without Prophet.

zhaoyongjie (commented):
I got a weird area when create custom forecasts

image

villebro:
> I got a weird area when create a custom forecasts
> 
> image

Oh, I think it's because your lower bound (`SUM(num) * 1.2`) is above the upper bound (`SUM(num) * 1.1`).

zhaoyongjie (approved):
tested in local. LGTM

image

zhaoyongjie:
> > I got a weird area when create a custom forecasts
> > image
> 
> Oh, I think it's because your lower bound (`SUM(num) * 1.2`) is above the upper bound (`SUM(num) * 1.1`).

Thanks! You are right.

PR #1423, opened 2021-10-22
feat: add drag and drop column rearrangement for table plugin

### LINKED ISSUE #1420: Drag and Drop user-reorganizable columns in the Table Viz
### Is your feature request related to a problem? Please describe.

Sometimes the dashboard designer doesn't know how the end user will want their columns organized in a table. Why not let the user rearrange the columns themselves if they want?

### Describe the solution you'd like

react-table has a built-in useColumnOrder hook, and there was a discussion online about how to implement drag-and-drop in a react-table.
https://github.com/tannerlinsley/react-table/issues/256

### Describe alternatives you've considered

None, this is the easiest approach.

### Additional context\*\* Add any other context or screenshots about the feature request here.

I've implemented it in a custom viz I built that is based on the table viz, and so I can redo it in the original plugin and issue a PR soon.

stevetracvc:
🏆 Enhancements
Add the ability for the end user to drag and drop columns for a table visualization

📜 Documentation
Table designer can enable column rearrangement in the config options when building the table. Then, user can drag columns around

These are volatile changes, so when the browser is refreshed, the column order goes back to default. Demo is in the chart edit page, which is just for simple demonstration purposes. This is intended for the **_end user_** to be allowed to rearrange the columns.

Not enabled
chrome-capture

Enabled
chrome-capture (1)

closes #1420


villebro:
I think this is a great improvement @stevetracvc! To make these changes stick, however, it would be nice if we could add drag'n'drop support to the "Customize columns" control, which would persist the order. If we added that support, do you think that would be sufficient, or would it make sense to also support the temporary column rearrangement introduced in this PR?

stevetracvc:
@villebro I don't think it's nearly as useful to the table _designer_, as there's already the drag-and-drop in the list of columns for reorganizing the columns on that page. This is intended to be used by the end users, so they can customize it to their liking. I only did the video on the explore page so you could see what happens after adding the check mark :) But if you think designers would want it too, that's another discussion.

And since it's the _end users_ doing the reordering, I didn't see any need to make the new ordering stick (at least globally, as someone could really screw with the rearrangement and ruin it for everyone).

One possible addition I can see would be to make an individual user's reorganization something that's stored in the local session, so that coming back in the future, the columns would be in the same order. But that seems like it could get messy and we'd need to figure out how to make sure that it's the same: user, dashboard, table, among other things, to make sure there's no adverse affects.

From my testing, applying new filters and even force refreshing the data does not reset the column order, it only resets when you browser refresh the dashboard.

stevetracvc:
Another thing I just noticed...column widths currently will only update after a data refresh

junlincc:
This is a great improvement! Thank you @stevetracvc It works similar to googlesheet and I don't think adding a control in the panel is necessary. but I do agree the changes should persist while the chart is added to dashboard. 

@villebro If the code looks ok and test is sufficient , I think we should merge it in and make incremental improvement later.

stevetracvc:
@junlincc are you saying you think the column order should persist on a given end user's computer, even after closing the browser and coming back?

I think that'd be nice, but I also think it'd be great for all the filters and such to be remembered too. From what I can tell, it appears that the general design of Superset is for it to start from scratch every time the dashboard is reloaded. Is there any infrastructure yet to make any user settings non-volatile? Stored for the user in the database?

rusackas:
I think @junlincc's comment is intending to state that _if_ a chart author rearranges columns, and then hits Save, then that column order should be persisted with the chart and reflected on dashboards.

I think the proper way to do this might be to have the drag and drop action also update the form data and update the order of the control values to match (dynamic controls, essentially). Then saving it would persist the form data that way, and all would be well.

Dashboard consumers could still re-order columns as much as they'd like, without persisting the state there.

stevetracvc:
@rusackas ok, but a few questions:
1. does the table designer really need to be able to rearrange columns in so many ways? dnd is already an option for the column layout in the COLUMNS area of the table design.
2. Can I accomplish this by just overwriting the all_columns value in the formData object with the array of columns in the new order?
3. how can I tell whether the table is in edit mode, vs it's just some random user viewing the table? Is there harm in updating the formData object if the table is actually in a dashboard and being manipulated by the user? I get that the metadata database wouldn't be updated since the dashboard wouldn't have a "save" button, but is it bad practice to modify this when it's not necessary?

villebro:

PR #1447, opened 2021-11-01
fix(plugin-chart-echarts): fix tick labels and tooltip

villebro:
🐛 Bug Fix

Fixes multiple issues on the Mixed Timeseries chart:
- Tooltip x-axis value shows `0NaN` when rich tooltip is disabled and the value is a decimal.
- On the case of a mixed line+bar chart, the bars below the line won't show in the tooltip.
- Remove bogus min and max x-axis values (these have been deprecated from the main Timeseries chart some time ago).

### AFTER
Tooltip correctly shows the x-axis value when the y-value is a decimal:
image

The erroneous min and max x-axis values are removed, and hovering on a bar underneath a line works as intended:
https://user-images.githubusercontent.com/33317356/139655494-8ab0257a-2ece-4821-a41e-727437cd8490.mp4

### BEFORE

Tooltip shows `0NaN` when the metric value is a decimal value:
image

The x-axis shows incorrect min and max values, and hovering on a bar underneath a line doesn't display the bar value correctly:
https://user-images.githubusercontent.com/33317356/139655424-20628415-b6a2-46a7-81e7-fb310e894e19.mp4

### DIFF
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
@@ -95,11 +95,11 @@ export default function EchartsMixedTimeseries({
         handleChange([seriesName], seriesIndex);
       }
     },
-    mousemove: params => {
-      currentSeries.name = params.seriesName;
-    },
     mouseout: () => {
       currentSeries.name = '';
+    },
+    mouseover: params => {
+      currentSeries.name = params.seriesName;
     },
   };
 
--- a/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
+++ b/plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
@@ -99,8 +99,6 @@ export default function transformProps(
     tooltipTimeFormat,
     yAxisFormat,
     yAxisFormatSecondary,
-    xAxisShowMinLabel,
-    xAxisShowMaxLabel,
     xAxisTimeFormat,
     yAxisBounds,
     yAxisIndex,
@@ -244,8 +242,6 @@ export default function transformProps(
       nameGap: xAxisTitleMargin,
       nameLocation: 'middle',
       axisLabel: {
-        showMinLabel: xAxisShowMinLabel,
-        showMaxLabel: xAxisShowMaxLabel,
         formatter: xAxisFormatter,
         rotate: xAxisLabelRotation,
       },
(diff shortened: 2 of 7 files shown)


plugins/plugin-chart-echarts/src/MixedTimeseries/transformProps.ts
```diff
@@ -282,14 +278,14 @@ export default function transformProps(
       appendToBody: true,
       trigger: richTooltip ? 'axis' : 'item',
       formatter: (params: any) => {
-        const value: number = !richTooltip ? params.value : params[0].value[0];
-        const prophetValue: any[] = !richTooltip ? [params] : params;
+        const xValue: number = richTooltip ? params[0].value[0] : params.value[0];
+        const prophetValue: any[] = richTooltip ? params : [params];
```
villebro, line 282: When `richTooltip` is disabled, we were referencing the full `value`, which is always a tuple with x and y value. Here I'm renaming it to `xValue` to better reflect that it is in fact the x-value (the first element of the tuple) and to disambiguate it from the `value` parameter in the `forEach` loop a few lines below.

plugins/plugin-chart-echarts/src/Timeseries/transformProps.ts
```diff
@@ -268,14 +268,14 @@ export default function transformProps(
       appendToBody: true,
       trigger: richTooltip ? 'axis' : 'item',
       formatter: (params: any) => {
-        const value: number = !richTooltip ? params.value : params[0].value[0];
-        const prophetValue: any[] = !richTooltip ? [params] : params;
+        const xValue: number = richTooltip ? params[0].value[0] : params.value[0];
+        const prophetValue: any[] = richTooltip ? params : [params];
```
villebro, line 272: The same bug appears on the main timeseries chart type.

plugins/plugin-chart-echarts/src/MixedTimeseries/EchartsMixedTimeseries.tsx
```diff
@@ -95,12 +95,12 @@ export default function EchartsMixedTimeseries({
...
     },
-    mousemove: params => {
-      currentSeries.name = params.seriesName;
-    },
     mouseout: () => {
       currentSeries.name = '';
     },
+    mouseover: params => {
+      currentSeries.name = params.seriesName;
+    },
```
villebro, line 103: Move to use the new `mouseover` event introduced for line charts in ECharts 5.2.2

plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
```diff
@@ -178,12 +178,14 @@ export function transformSeries(
...
-    areaStyle: {
-      opacity:
-        forecastSeries.type === ForecastSeriesEnum.ForecastUpper || area
-          ? opacity * areaOpacity
-          : 0,
-    },
+    areaStyle: area
+      ? {
+          opacity:
+            forecastSeries.type === ForecastSeriesEnum.ForecastUpper || area
```
stephenLYZ, line 184: The conditional expression is always true. Maybe we can remove it here.

stephenLYZ:
Thanks for the improvement! LGTM and just left a comment.

stephenLYZ (approved):

plugins/plugin-chart-echarts/src/Timeseries/transformers.ts
villebro: Whoops, good catch!

zhaoyongjie (approved):
LGTM

PR #1448, opened 2021-11-01
fix(plugin-chart-echarts): boxplot groupby incorrect

villebro:
🐛 Bug Fix
The post processing operation didn't include the groupby columns, resulting in NULL values on the x-axis. This PR fixes the bug, adds tests and adds type guards for all post processing rules.

Closes https://github.com/apache/superset/issues/16951 , closes https://github.com/apache/superset/issues/16985

### AFTER
image

### BEFORE
image

### DIFF
--- a/packages/superset-ui-chart-controls/src/operators/boxplotOperator.ts
+++ b/packages/superset-ui-chart-controls/src/operators/boxplotOperator.ts
@@ -16,12 +16,7 @@
  * specific language governing permissions and limitationsxw
  * under the License.
  */
-import {
-  PostProcessingBoxplot,
-  getMetricLabel,
-  ensureIsArray,
-  QueryFormColumn,
-} from '@superset-ui/core';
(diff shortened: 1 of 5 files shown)


packages/superset-ui-chart-controls/src/operators/boxplotOperator.ts
```diff
@@ -54,7 +49,7 @@ export const boxplotOperator: PostProcessingFactory<PostProcessingBoxplot | unde
       options: {
         whisker_type: whiskerType,
         percentiles,
-        groupby: ensureIsArray(queryObject.groupby as QueryFormColumn[]),
+        groupby: ensureIsArray(groupby),
```
villebro, line 52: `groupby` is deprecated on `QueryObject` and is ambigous with `columns`, hence we should use `groupby` from `formData` here, as it carries a specific meaning for the Boxplot chart.

zhaoyongjie (approved):
LGTM

PR #1450, opened 2021-11-02
feat:  remove the default Y-axis truncate in bar chart

stephenLYZ:
remove the default Y-axis truncate check box to “unchecked” and add a tooltip next to the checkbox saying “It’s not recommended to truncate y-axis in Bar chart”

image

### DIFF
--- a/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
+++ b/plugins/plugin-chart-echarts/src/Timeseries/Regular/Bar/controlPanel.tsx
@@ -93,7 +93,7 @@ const config: ControlPanelConfig = {
       label: t('Chart Options'),
       expanded: true,
       controlSetRows: [
-        ['color_scheme', 'label_colors'],
+        ['color_scheme'],
         ...showValueSection,
         [
           {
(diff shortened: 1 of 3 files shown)


zhaoyongjie (approved):
@junlin code LGTM. one nite: should we keep all echart timeseries viz consistent for this control?

villebro:
For context, y-axis truncation was enabled by default only because it was default behavior in the NVD3 charts. I'm personally against truncation of y-axes by default (I think it's ok in certain cases, but should be a conscious decision), so I'd be happy to see truncation being disabled by default in all charts. But I assume there may be community resistance to this, as changing these types of things have historically proven to be a fairly sensitive topic.

villebro (commented):
Code changes LGTM, but let's think this through once more before merging to make sure we have broad alignment for changing the default behavior.

junlincc:
Is the tooltip only showing in bar chart? if both behaviors 1)not set by default 2) showing tooltip are only showing in Bar chart, i think we are good.

zhaoyongjie:
> Is the tooltip only showing in bar chart? if both behaviors 1)not set by default 2) showing tooltip are only showing in Bar chart, i think we are good.

So, Do we change the `truncate y-axis` control to uncheck in all echart viz?

PR #1452, opened 2021-11-02
feat(legacy-plugin-chart-calendar): increase the contrast of calendar heatmap color and label

stephenLYZ:
🏆 Enhancements
The algorithm is to decide label color in white or black depending on background color (refer to https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color)

image

### DIFF
--- a/plugins/legacy-plugin-chart-calendar/src/vendor/cal-heatmap.js
+++ b/plugins/legacy-plugin-chart-calendar/src/vendor/cal-heatmap.js
@@ -9,6 +9,7 @@
 /* eslint-disable */
 
 import d3tip from 'd3-tip';
+import { getContrastingColor } from '@superset-ui/core';
 import './d3tip.css';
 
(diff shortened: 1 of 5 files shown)


plugins/legacy-plugin-chart-calendar/src/vendor/cal-heatmap.js
```diff
@@ -1652,6 +1652,20 @@ CalHeatMap.prototype = {
       }
     }
 
+    function formatTextFill(value) {
```
rusackas, line 1655: This is super nifty! I wonder if we should label it as something more descriptive, like "getContrastingColor" so that we can eventually (after monorepo migration, perhaps) export it from a shared module, like Superset's `colorUtils.ts` file.

zhaoyongjie (approved):
LGTM, Thanks for adding the complete testing.

PR #1462, opened 2021-11-09
fix: update chart sorting control labels/descriptions

john-bodley:
🏠 Internal

See https://github.com/apache/superset/pull/17383.


etr2460 (approved):
stamping, but please rebase to make sure ci passes

PR #1463, opened 2021-11-10
chore: coordinate superset-ui unittest with main repository

zhaoyongjie:
🏠 Internal
coordinate superset-ui unittest with main repository
1. Jest use source code for module import
2. remove dependence: `@airbnb/config-jest`
3. remove `@airbnb/config-jest` related "setup" before unit test
4. refactor some unittest
5. clean up

### DIFF
--- a/packages/superset-ui-core/test/translation/Translator.test.ts
+++ b/packages/superset-ui-core/test/translation/Translator.test.ts
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+import { logging } from '@superset-ui/core';
 import Translator from '@superset-ui/core/src/translation/Translator';
 import {
   configure,
@@ -34,6 +35,18 @@ configure({
 });
 
 describe('Translator', () => {
+  const spy = jest.spyOn(logging, 'warn');
+
+  beforeAll(() => {
+    spy.mockImplementation((info: any) => {
+      throw new Error(info);
+    });
+  });
+
+  afterAll(() => {
+    spy.mockRestore();
+  });
+
   describe('new Translator(config)', () => {
     it('initializes when config is not specified', () => {
       expect(new Translator()).toBeInstanceOf(Translator);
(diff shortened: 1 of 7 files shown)


villebro (commented):
LGTM, but I'd prefer to get @kgabryje @geido or @michael-s-molina to also take a look, as this is unfamiliar territory to me.

packages/superset-ui-core/test/translation/Translator.test.ts
```diff
@@ -34,6 +35,18 @@ configure({
 });
 
 describe('Translator', () => {
+  const spy = jest.spyOn(logging, 'warn');
+
+  beforeAll(() => {
+    spy.mockImplementation((info: any) => {
```
kgabryje, line 41: Can you explain why do we need to mock logging methods? And why do they throw errors?

packages/superset-ui-core/test/utils/logging.test.ts
```diff
@@ -36,9 +40,16 @@ describe('logging', () => {
     expect(() => {
       logging.error('error');
     }).toThrow('error');
+
+    // to support: npx jest --silent
+    const spy = jest.spyOn(logging, 'trace');
+    spy.mockImplementation(() => {
```
kgabryje, line 46: Should we include `logging.trace`'s argument here? Like:
`throw new Error('Trace: ' + traceMessage)`

zhaoyongjie: This mock just for `logging.trace()`, Ensure that when call the `logging.trace()`, there is `Trace:` at the beginning of the return.

packages/superset-ui-core/test/translation/Translator.test.ts
kgabryje: Oh, I see, so we just bring back relevant changes from removed library 👍

zhaoyongjie: There are some test cases that need to get warning message.

kgabryje (approved):
Looks good, but just like Ville I'm not 100% familiar with those configs. Let's have third pair of eyes take a look just to be sure

ktmud (approved):
lgtm

PR #1464, opened 2021-11-10
fix(plugin-chart-echarts): fix opacity on area chart

villebro:
🐛 Bug Fix

#1447 introduced a regression to area charts which broke both area and confidence bands. The reason for the original change was to remove area if it's not needed (a 0 opacity made it impossible to hover series underneath the series).

### AFTER
image

### BEFORE
image


kgabryje (approved):
LGTM

PR #1465, opened 2021-11-10
fix(plugin-chart-pivot-table): cross filtering by adhoc column

kgabryje:
Cross filtering in PivotTable v2 didn't work properly when applied to an adhoc column (new feature). This PR fixes the issue.

image


villebro (approved):
LGTM

PR #1472, opened 2021-11-15
fix: missing $ in histogram popover for percentile value

stevetracvc:
Fix: missing $ in popover for percentile value

🐛 Bug Fix

Sorry, I missed this in my earlier PR

### DIFF
--- a/plugins/legacy-plugin-chart-histogram/src/Histogram.jsx
+++ b/plugins/legacy-plugin-chart-histogram/src/Histogram.jsx
@@ -122,7 +122,7 @@ class CustomHistogram extends React.PureComponent {
                 </div>
                 <div>
                   <strong>{t('percentile (exclusive)')} </strong>
-                  {`{((datum.cumulativeDensity - datum.density) * 100).toPrecision(4)}th`}
+                  {`${((datum.cumulativeDensity - datum.density) * 100).toPrecision(4)}th`}
                 </div>
               </div>
             )}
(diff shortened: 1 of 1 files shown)


stevetracvc:
Sorry @zhaoyongjie I missed this in one of my tweaks before!

Also, the "th" at the end of a percentile might need to be changed, what do you think? My intent was to make sure people understood that it's out of 100 (though it isn't really the same as %). But for numbers like 95.62, is '95.62th' really a good way to write it?

junlincc:
Hi @stevetracvc thanks again for your contribution! Could you add video and screen that describe your changes. Thanks!!

stevetracvc:
@junlincc the fix is literally just adding a $ so the values get expanded in the string. Currently it shows the code in the string rather than the expansion. And then prettier wanted a few newlines.

villebro:
@stevetracvc I agree with @junlincc, it would be nice if you could add some more context in the PR description. I know it may seem like this is a trivial change, but as a general rule it really helps speed up the review process if the description is as detailed as possible. For instance, in addition to having a video/screenshot of before and after, linking to the PR that caused the regression also helps reviewers quickly get up to speed on why the change is necessary, reproduce the problem, identify any potential tests that may need to be added or identify any related code paths that may be affected.

stevetracvc:
@villebro I know, but for some reason my node dev-server won't run anymore (it dies from out of memory, no matter what settings I use) so it's like a 20 min round trip to go back and forth. And I just made the patch on github's website directly since my PR had already been incorporated and I figured hey, one character should be no biggie :)

And now I can't get it to load from my local superset-ui without errors. I'll try to get you a screenshot tonight.

villebro:

